PR #1, opened 2022-04-09
Setup base software stack via virtualenv

riga:
Currently, the initial software stack (mainly `luigi, six, scinum`) is installed using the system pip3 into a custom directory in `$AP_SOFTWARE`. This PR changes that slightly to use virtualenv, or rather the builtin `venv` module to setup this up in a controlled way.

Moreover, there are two virtualenv's setup simultaneously:
- `ap_prod`: The "production" environment with the minimal set of packages to run the tasks. For remote jobs, this env is bundled and send with jobs (not as input sandboxes though, but it's first uploaded to a remote location and then retrieved by the worker node).
- `ap_dev`: The "development" environment is the same as `ap_prod` **plus** additional tools that might be necessary or helpful when developing code, such as ipython or flake8.

The stacks are controlled with two requirement files `requirements_{prod,dev}.txt`. For user-defined packages that should be added to `ap_dev`, a custom `requirements_user.txt` file can be used (but not pushed to the repository). In case of a package that is also useful for others, consider adding it to `requirements_dev.txt` directly.

One should continue to use sandboxes to provide software for task-related code.

@uhh-cms/ag-schleper


riga:
I'm going ahead now without review, since we checked the setup in the last days (and we didn't discuss review rules yet :) )

PR #31, opened 2022-06-07
changed allowed values in category axis to only leaf cat ids

mafrahm:
To access the correct bin in an intCategory, one needs to provide the position of the bin instead of the value of the bin. This can be done by `hist.loc(i)`

### DIFF
--- a/ap/tasks/plotting.py
+++ b/ap/tasks/plotting.py
@@ -127,8 +127,7 @@ class Plotting(CalibratorsSelectorMixin, PlotMixin, law.LocalWorkflow, HTCondorW
 
                         # to access the correct bins in the IntCat axis, we need
                         # the position of the bins, not the id itself
-                        id_to_pos = {cat_id: i for i, cat_id in enumerate(h_in.axes["category"])}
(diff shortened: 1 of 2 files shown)


ap/tasks/plotting.py
```diff
@@ -279,18 +285,19 @@ def run(self):
...
                 elif c.get_category(category).is_leaf_category:
                     leaf_cats = [c.get_category(category).id]
                 else:
                     leaf_cats = [cat.id for cat in c.get_category(category).get_leaf_categories()]
 
-                # Note: this only works because the category axis is sorted and does not skip an integer
-                # h_in[{"category": [0,5]}] gives the categories that are at the position 0 and 5
-                h_in = h_in[{"category": leaf_cats}]
+                # to access the correct bins in the IntCat axis, we need the position of the bins, not the id itself
+                leaf_to_pos = [hist.loc(i) for i in leaf_cats]
```
riga, line 298: The linter complains about `hist` not being known at this point.

riga (changes requested):

riga (approved):

PR #35, opened 2022-06-07
Draft: Real jet energy corrections / jet energy resolution smearing

dsavoiu:
(This is a draft PR, so not everything is implemented yet. I'm sharing it here to get early feedback in case some aspects of the implementation could be refined.)

This PR implements actual JECs using the vectorized (?) tools implemented in coffea. A new `Calibrator` is implemented under `ap/calibration/jec.py` to contain all JEC-related code in one place. A number of utility functions are implemented to construct the Coffea helper objects used to parse the JEC text files and apply them to the event content. Memoization is used to cache the resulting objects. All JEC uncertainties found in an "UncertaintySources" text file are also applied as shifts.

Some things could still be improved:
- [ ] use new `external_files` implementation to read JEC files
  - where to get them from? [`JECDatabase`](https://github.com/cms-jet/JECDatabase) would be one possible choice
  - how to handle `tar.gz` files? have to extract them or does coffea code support `tar.gz` files?
- [ ] the JER uncertainty names are era-dependent. Since one wants to use these in the shift column names they have to passed to the `Calibrator` in a config-dependent way, but the current implementation only allows hard-coded values.

[JER smearing is not yet implemented.]


riga:
Nice!

I think #36 contains a few additions that solve your second todo (check [this gist](https://gist.github.com/riga/97e4a7b4b765b9b58b6e10ff7de395d9) for an overcomplete, yet not very sensible example).

For the first one:
- we could also add the particular txt files directly to the bundle
- if you want to go for tgz, then

```python
tmp = law.LocalDirectoryTarget(is_tmp=True)
target_wrapping_the_tgz.load(tmp, formatter="tar")
```

would do the trick.

ap/calibration/jets.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# first, some utility functions
+#
+
+@memoize
+def get_lookup_provider(files, conversion_func, provider_cls, names=None):
+    """
+    Create a coffea helper object for looking up information in files of various formats.
+
+    This function reads in the `files` containing lookup tables (e.g. JEC text files), extracts
```
riga, line 32: For the doc strings it might be better to use e.g. `*files*` to highlight arguments. RsT unfortunately doesn't play nicely with the default markdown style.

ap/calibration/jets.py
```diff
@@ -0,0 +1,432 @@
...
+        coffea_txt_converters.convert_jr_txt_file,
+        coffea_jetmet_tools.JetResolution,
+        names=jer_names,
+    )
+    jersf_provider = get_lookup_provider(
+        jersf_files,
+        coffea_txt_converters.convert_jersf_txt_file,
+        coffea_jetmet_tools.JetResolutionScaleFactor,
+        names=jersf_names,
+    )
```
riga, line 289: Should both `jer_provider` and `jersf_provider` be created once in the setup function? Or maybe I don't see the dependence in event content.

riga (changes requested):
Wow, this looks like a lot of work. I think many calibration methods can use this as a template in the future :) I added a few inline comments below, but nothing critical.

ap/calibration/jets.py
dsavoiu: Good point! I was thinking the (single) backticks will render as monospace in RsT as well, but I guess that is not as reliable as it depends on the style. I'll change them to asterisks for the args, and qualify the backticks with `:py:class:` etc. to conform to RsT syntax.

ap/config/analysis_st.py
```diff
@@ -45,6 +45,99 @@
 # create a config by passing the campaign, so id and name will be identical
 config_2018 = analysis_st.add_config(campaign_2018)
 
+# location of JEC txt files
+config_2018.set_aux("jec", DotDict.wrap({
+    "source": "https://raw.githubusercontent.com/cms-jet/JECDatabase/master/textFiles",
+    "campaign": "Summer19UL18",
+    "version": "V5",
+    "jet_type": "AK4PFchs",
+    "levels": ("L1FastJet", "L2Relative", "L2L3Residual", "L3Absolute"),
```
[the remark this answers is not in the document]
dsavoiu: Not sure what the 2016 situation was that you are referring to, but yes, the levels are technically different, in that `L2L3Residual` should only be applied in data. On the other hand, JME provides these files also for MC with a trivial factor of `1.0` so one can just take the same list as for data.

But it wouldn't hurt to make the levels dataset-dependent as a safeguard against nontrivial residuals commited by mistake for MC. I'll have a look.

ap/config/analysis_st.py
```diff
@@ -45,6 +45,99 @@
...
+    "data_eras": ("RunA", "RunB", "RunC", "RunD"),
```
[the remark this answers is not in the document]
dsavoiu: Sure, if you like :) I'm a bit paranoid about having mutable things in configs just in case some piece of user code ends up changing the value without realizing, but that's perhaps unjustified here.

ap/selection/test.py
```diff
@@ -20,9 +20,28 @@ def extract(array, idx):
...
+        for junc_dir in ('up', 'down')
+    }
+    if dataset_inst.is_mc:
+        self.shifts |= {"jer_up", "jer_down"}
+
+
 # object definitions
 # TODO: Producer instead of Selector
-@selector(uses={"Jet.pt", "Jet.eta"})
+@selector(uses={"Jet.pt", "Jet.eta"}, shifts={jet_energy_shifts})
```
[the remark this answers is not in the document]
dsavoiu: Yeah, I see how this could become problematic with more calibrators. But I'm not sure the alternative to implement the shifts only for the main selector is better. In principle one should be able to use each selector independently, so each one should properly declare its shifts.

Maybe we could just leave it as it is for the moment and see if we run into problems once more selectors/calibrators are added?

ap/calibration/jets.py
```diff
@@ -0,0 +1,432 @@
...
+        set_ak_column(events, f"Jet.mass_{name}_up", events.Jet.mass * jec_unc_factors[:, :, 0])
+        set_ak_column(events, f"Jet.mass_{name}_down", events.Jet.mass * jec_unc_factors[:, :, 1])
+
+    return events
+
+
+@jec.update
+def jec_update(self, config_inst, **kwargs):
+    """Add JEC uncertainty shifts to the list of produced columns."""
+    for shifted_var in ("pt", "mass"):
```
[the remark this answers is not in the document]
dsavoiu: Yes, that makes sense. I was doing something else with the shifted_var here originally, hence the outer loop, but that didn't make it to the final version. I'll change this.

ap/calibration/jets.py
dsavoiu: You're right, there is no dependence on the event content here and this can definitely be moved to `setup`. (One might have to revert this if we ever get JECs again that don't align nicely with the eras and we need to check the event content to get the run, but then some other things would also need to change too.)

ap/util.py
```diff
@@ -362,3 +363,37 @@ def dev_sandbox(sandbox: str) -> str:
...
+    """Constructs an immutable version of a native Python container.
+
+    Recursively replaces all mutable containers (`dict`, `list`, `set`) encountered within
+    `cont` by an immutable equivalent: Lists are converted to tuples, sets to `frozenset`
+    objects, and dictionaries to tuples of (*key*, *value*) pairs.
+    """
+
+    if isinstance(cont, dict):
+        return tuple((k, freeze(v)) for k, v in cont.items())
+    elif isinstance(cont, list) or isinstance(cont, tuple):
```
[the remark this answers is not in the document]
dsavoiu: Can't go wrong with sugar ;)

ap/config/analysis_st.py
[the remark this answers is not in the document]
riga: Thanks !

ap/calibration/jets.py
riga: Ah right.... run-dependence of data comes to mind. Since the calls are memoized are thus hopefully fast enough (esp. compared to the actual JEC runtime), maybe it's even better to leave them here :)

ap/config/analysis_st.py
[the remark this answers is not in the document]
dsavoiu: I think for the time being we should leave this as-is and revisit if necessary once we figure out how to organize the config for real data (see #50)

ap/calibration/jets.py
dsavoiu: Done

ap/calibration/jets.py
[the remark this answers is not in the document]
dsavoiu: Done

ap/calibration/jets.py
dsavoiu: OK, no harm leaving them here for the time being. I've added an explanatory comment in case we do decide to refactor this in the future.

ap/production/features.py
```diff
@@ -8,6 +8,8 @@
 from ap.util import maybe_import
 from ap.columnar_util import EMPTY, set_ak_column
 
+from ap.selection.test import jet_energy_shifts
```
dsavoiu, line 11: Actually, now I'm wondering if there is a better way to do this that avoids creating a pseudo-selector (particularly now that the  `variables` is a producer and lives in a completely different module). @riga Ideas?

riga: I think the pseudo selector is fine, but after thinking about it some more, I think I wouldn't add `jet_energy_shifts` to the "shifts" sets of all jet-related selectors. At the point where selectors are called, the jer and jes columns (nominal + shifts) are already created, so it's just about declaring the shifts once, which probably should happen in the "main" selector. This will cause the full set of shifts being propagated to the SelectEvents task, which sets the aliases accordingly, and all selectors will receive the varied columns. Or maybe I'm missing something? What do you think?

dsavoiu: I see what you mean. My intention was to declare the shifts as close as possible to the place they are actually needed. The advantage of that is that the individual selectors then are in principle independent and can be reused in case other "main" selectors are implemented without having to specify the shifts again. But maybe this is not taking into account all the use cases.

Since in the mean time we also discussed introducing a distinction between "exportable" selectors that can be used by tasks and "non-exportable" ones that are called as dependencies, maybe this will have to be refactored annyway. So at least for now I've moved the shifts to the main selector like you suggested.

PR #36, opened 2022-06-08
Refactor mixins to inject {calibrator,selector,producer} info to tasks

riga:
This PR contains some structural refactoring. Through the `{Calibrator,Selector,Producer}Mixin` classes, tasks obtain the feature to call arbitrary array functions, configured via `--{calibrator(s),selector,producer(s)}` arguments. However, to properly integrate these externalized functions in a plug-and-play fashion, information such as required / produced systematic shifts need to be correctly propagated. This happens mostly in class methods of the base tasks, since these decisions need to be taken _before_ a task class is instantiated.

The PR refactors

- how task classes can define their shifts (they can now depend on the above mentioned array functions, _plus_ they allow resolving this same information for dependent tasks), and
- how array functions can customize used / produced columns, as well as systematic shifts (they can wrap a function accepting analysis, config, dataset inst's, and performing updates to (e.g.) their shifts).

### DIFF
--- a/ap/columnar_util.py
+++ b/ap/columnar_util.py
@@ -718,21 +718,27 @@ class ArrayFunction(object):
     @property
     def used_columns(self) -> Set[str]:
         columns = set()
+
+        # add those of all other known intances
         for obj in self.uses:
             if isinstance(obj, self.__class__):
                 columns |= obj.used_columns
             else:
                 columns.add(obj)
+
         return columns
 
     @property
     def produced_columns(self) -> Set[str]:
         columns = set()
+
+        # add those of all other known intances
         for obj in self.produces:
             if isinstance(obj, self.__class__):
                 columns |= obj.produced_columns
             else:
                 columns.add(obj)
+
         return columns
 
     def __repr__(self) -> str:
@@ -759,7 +765,7 @@ class TaskArrayFunction(ArrayFunction):
     :py:class:`ArrayFunction` base class. It can be a sequence or set of shift names, or other
     instances that are called by this one. The :py:attr:`all_shifts` property returns a flat set of
     all shifts, potentially resolving information from other :py:class:`TaskArrayFunction`
-    instances.
+    instances registered in `py:attr:`uses`, `py:attr:`produces` and `py:attr:`shifts` itself.
 
     Custom task requirements are be defined in a programmatic way by wrapping a function through a
     decorator. A custom setup function, using results of the custom task requirements to (e.g.)
@@ -795,7 +801,6 @@ class TaskArrayFunction(ArrayFunction):
             weights = inputs["weights_task"].load(formatter="json")
             # add it to the call_kwargs which will be forwarded to every call to the main function
             call_kwargs["weights"] = weights
-
 
     For a possible implementation, see :py:mod:`ap.production.pileup`.
 
@@ -856,7 +861,7 @@ class TaskArrayFunction(ArrayFunction):
         shifts = set()
 
         # add those of all other known intances
-        for obj in self.shifts:
+        for obj in self.uses | self.produces | self.shifts:
             if isinstance(obj, self.__class__):
                 shifts |= obj.all_shifts
             else:


ap/columnar_util.py
```diff
@@ -745,8 +751,9 @@ def __call__(self, *args, **kwargs):
 class TaskArrayFunction(ArrayFunction):
     """
     Subclass of :py:class:`ArrayFunction` providing an interface to certain task features such as
-    declaring task requirements and dependent or produced shifts, and defining a custom setup
-    function.
+    declaring dependent or produced shifts, task requirements, and defining a custom setup
```
dsavoiu, line 754: AFAIU the `shifts` keyword currently only declares dependent shifts (as in the shifts that affect the result of the array function and for which the column aliases should be applied before the array function is executed). Wouldn't it make sense to also declare produced shifts via a separate keyword?

ap/columnar_util.py
```diff
@@ -842,14 +862,53 @@ def all_shifts(self):
...
+        self.update_func = func
+
+    def run_update(self, **kwargs) -> None:
+        """
+        Recursively runs the update function of this and all other known instances (via
+        :py:attr:`uses` and :py:attr:`produces`) forwarding *this* instance and all additional
+        *kwargs*.
+        """
+        # run the update of all other known instances
+        for obj in self.uses | self.produces:
```
dsavoiu, line 895: I think one should also run over `self.shifts` here. Right now if other array functions are specified via `shifts` the `update` function is not run on them, which may lead to wrong `shift` resolution. One workaround would be to add these to `uses`, but if one is just interested in the shifts and not the results, passing them via `shifts` would be semantically cleaner. 
See also PR #35 and in particular this commit https://github.com/uhh-cms/analysis_playground/pull/35/commits/607e534fec9c6bcc407afcafc984f25560f349c7 for an example where this becomes relevant.

ap/columnar_util.py
```diff
@@ -879,56 +938,57 @@ def run_requires(self, task: law.Task, reqs: Optional[dict] = None):
...
         dictionary which should be filled to store arguments which are later on passed to the
         wrapped function.
         """
-        if producer_kwargs is None:
-            producer_kwargs = {}
+        if call_kwargs is None:
+            call_kwargs = {}
 
         # run the setup of all other known instances
         for obj in self.uses | self.produces:
```
dsavoiu, line 974: Also run over `self.shifts` (see comment for `run_update`)

ap/columnar_util.py
```diff
@@ -863,7 +922,7 @@ def requires(self, func: Callable[[law.Task, dict], dict]) -> None:
         """
         self.requires_func = func
 
-    def run_requires(self, task: law.Task, reqs: Optional[dict] = None):
+    def run_requires(self, task: law.Task, reqs: Optional[dict] = None) -> dict:
```
dsavoiu, line 925: Also propagate to `self.shifts` (see comment for `run_update`)

ap/columnar_util.py
```diff
@@ -842,14 +862,53 @@ def all_shifts(self):
```
dsavoiu, line 862: Shoudn't the line above be `shifts |= obj.all_shifts` to properly resolve all dependent shifts?

ap/columnar_util.py
```diff
@@ -827,9 +846,10 @@ def __init__(
...
         # store attributes
         self.shifts = set(law.util.make_list(shifts)) if shifts else set()
+        self.update_func = None
         self.requires_func = None
         self.setup_func = None
-        self.producer_kwargs = None
+        self.call_kwargs = None
 
     @property
     def all_shifts(self):
```
dsavoiu, line 855: Also propagate to `self.shifts` (see comment for `run_update`)

ap/tasks/selection.py
```diff
@@ -20,23 +20,31 @@ class SelectEvents(DatasetTask, CalibratorsSelectorMixin, law.LocalWorkflow, HTC
 
     sandbox = dev_sandbox("bash::$AP_BASE/sandboxes/venv_columnar.sh")
 
-    shifts = CalibrateEvents.shifts | {"jec_up", "jec_down"}
+    shifts = {CalibrateEvents, "jec_up", "jec_down"}
```
dsavoiu, line 23: This should probably be removed and injected dynamically by the JEC calibrator (already done in PR #35)

dsavoiu (approved):
This is very nice! I played around a lot with this while implementing JEC/JER and it looks like everything is mostly working as expected, with the exception of some use cases where shifts don't seem to propagate properly to dependent array functions, or at least not as I would expect them to :) I've added some comments with more details at the relevant lines in the code.

I'm marking this as approved, but maybe you could also have a look at PR #35 before merging, since I have included these changes there as well and also added some commits that I believe solve the shift propagation issues I was seeing. If you agree with them there should be no need to further change this PR and it can be merged together with PR #35. In case there are any issues feel free to follow up here or there, otherwise everything looks good on my end.

ap/columnar_util.py
riga: I also thought about that :)
But as far as the ShiftTask mechanism is concerned, dependent and produced shifts don't have a special distinction. For this matter, running with a certain shifts only means that "the output will be different than for 'nominal'".

ap/columnar_util.py
riga: True! I'll address this in the next commit.

ap/columnar_util.py
riga: True! I'll address this in the next commit.

ap/columnar_util.py
riga: True! I'll address this in the next commit.

ap/columnar_util.py
riga: Yes it should. And thinking about it, one shouldn't loop here over the `uses` and `produces` sets, but just look for other `ArrayFunction`s in `self.shifts`. This requires adding a dependent function in (e.g.) both `produces` _and_ `shifts`, but at the same time provides a better level of configurability.

ap/columnar_util.py
riga: ✔︎

ap/tasks/selection.py
riga: Yep. I would leave it to #35 to remove that :)

ap/columnar_util.py
riga: Thinking about this further, we should leave these special cases for the future and only take action if required.

ap/columnar_util.py
dsavoiu: I see, thanks for clarifying. Guess I was just a bit confused by the formulation in the docstring, but when you put it that way it makes sense.

ap/columnar_util.py
dsavoiu: 👍

ap/columnar_util.py
dsavoiu: Thanks! Looks good now.

ap/columnar_util.py
dsavoiu: Hm, I think it actually makes sense to also loop over `produces` and `uses` here. To my understanding, one would always want the `shifts` from these dependencies to be propagated to the current function's `all_shifts`, since they indirectly affect the result of the current function as well.

I would add them for now, but we should keep this in mind in case we need to make this more flexible for some future use case.

dsavoiu:
Looks good now, going ahead with the merge.

Issue #11, opened 2022-06-01
Vectorized ∆R matching


riga:
We need to build up expertise on complex vectorized algorithms such as a generic, matrix based matching algorithm between two sets of momentum vectors (in cartesian or lorentz-invariant coordinates).

bonanomi:
Quick out-of-the-box Google search: https://github.com/CoffeaTeam/coffea/issues/395 might be useful

artlbv:
Hi @kramerto @pkausw , I once asked Jim Pivarski how to do that with awk arrays – would this solution help you?
https://github.com/scikit-hep/awkward/discussions/996

I also included an example solution code in the thread. Let me know if I can help!

pkausw:
> Hi @kramerto @pkausw , I once asked Jim Pivarski how to do that with awk arrays – would this solution help you? [scikit-hep/awkward#996](https://github.com/scikit-hep/awkward/discussions/996)
> 
> I also included an example solution code in the thread. Let me know if I can help!

Thanks @artlbv for pointing us to this! Coincidentally, this is exactly what we used because we found it in the source code during the hackathon :D

pkausw:
Just to keep this issue up to date: a first version of the vectorized DeltaR matching is used in the cleaning of fields introduced in PR #37 . I would suggest to keep this issue open though since there is also the matter of doing an actual matching (e.g. of Reco Jets w.r.t. Gen Jets) which requires a different kind of logic

PR #37, opened 2022-06-13
Perform cleaning of objects on arbitrary metric, e.g. delta R

pkausw:
This is a draft pull request for the above use case so we can easily work on this issue together.
It relates to issue #11 

@kramerto and I implemented the first version of a cleaning of objects, e.g. `Jets`, against other objects, e.g. `Leptons`.
The cleaning can be based on an arbitrary metric, such as a DeltaR criterion.

The current implementation is based on the `nearest` function of the `coffea.LorentzVector`s, see [the source code](https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/methods/vector.py#L680).
This function creates a table for an arbitrary metric for all permutations of the objects in two fields (e.g. `Jets` and `Leptons`) and saves the minima of these permutations.
We extract this table and require that all values for the `Jet`s are above a threshold, thus clearing away `Jet`s that are too close to other objects of interest.

In the current version, the cleaning itself works perfectly fine. 
However, there is an error in the `SelectionResult` class that is yet to be fixed.
Nevertheless, I thought it might be a good idea to create this draft pull request to get early feedback and to facilitate common efforts in the bug hunt.


pkausw:
I have identified the bug and created the new issue #42 for it. This PR now contains a workaround. In the `sort_ak_fields` function creates a copy of the input array with the sorted fields. This is not nice and will hopefully be improved in the future. However, it works for now (there is just one minor issue when computing the total event statistics, which do not exist in the deltaR selector)

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+    stats["sum_mc_weight"] += ak.sum(events.LHEWeight.originalXWGTUP)
+    stats["sum_mc_weight_selected"] += ak.sum(events_sel.LHEWeight.originalXWGTUP)
+
+    return results
+
+# deltaR cleaning
+
+
+def cleaning_factory(
+    selector_name: AnyStr,
```
riga, line 313: Actually, this is a rather global decision, but I think we should tend to only accept `str` to avoid having to en/decode bytes (same applies to the other places in this file).

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+    First, the necessary column names to construct four-momenta for the
+    different object fields are constructed, i.e. pt, eta, phi and e for the
+    different objects.
+    Finally, the actual selector function is generated, which uses these
+    columns.
+    """
+
+    # compile the list of variables that are necessary for the four momenta
+    # this list is always the same
+    variables_for_lorentzvec = "pt eta phi e".split()
```
riga, line 329: Might be picky :)

```suggestion
    variables_for_lorentzvec = ["pt", "eta", "phi", "e"]

```

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+
+    # sum up all fields aht are to be considered, i.e. the field *to_clean*
+    # and all fields in *clean_against*
+    all_fields = list(clean_against[:])
+    all_fields.append(to_clean)
```
riga, line 334:
```suggestion
    all_fields = clean_against + [to_clean]
```

A matter of taste of course.

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+    # additionally, also load the lengths of the different fields
+    uses |= {f"n{x}" for x in all_fields}
+
+    # finally, construct selector function itself
+    @selector(uses=uses, name=selector_name)
+    def func(
+        events: ak.Array,
+        to_clean: AnyStr,
+        clean_against: List[AnyStr],
+        metric: Callable = lambda a, b: a.delta_r(b),
```
riga, line 353: There's a pattern stating that one should try to avoid using mutables as default arguments since they are stored as singletons and can unintentionally carry their mutable state to the next call frame. I would perhaps let it default to `None`, and in the function body set it to `lambda a, b: a.delta_r(b)` when `None`.

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+        # WARNING: this still contains the bug with the application of the mask
+        #           which will be adressed in a PR in the very near future
+        # TODO: return the mask itself instead of the list of indices
+        sorted_list = ak.argsort(to_clean_field.pt, axis=-1, ascending=False)[mask]
+        return sorted_list
+
+    return func
+
+
+deltaR_jet_lepton = cleaning_factory("deltaR_jet_lepton", "Jet", "Muon", "Electron")
```
riga, line 408: `delta_r_jet_lepton`? Just for case-consistency. And also in the places below.

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+    """
+    do the cleaning of jets with respect to leptons.
+    returns indices of good (clean) jets, sorted by pt
+    *TODO*: this should probably changed to return the boolean masks
+    """
+    return deltaR_jet_lepton(events, to_clean, clean_against, threshold=threshold)
+
+
+@selector(uses={deltaR_jet_lepton})
+def jet_lepton_deltaR_selection(events, stats, threshold=0.4):
```
riga, line 426: type hints

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+                                        threshold=threshold,
+    )
+
+    return SelectionResult(
+        objects={"Jet": clean_jet_indices},
+    )
+
+
+@selector(uses={jet_lepton_deltaR_selection})
+def jet_lepton_deltaR_cleaning(events, stats, threshold=0.4, **kwargs):
```
riga, line 446: type hints

ap/tasks/selection.py
```diff
@@ -119,18 +119,19 @@ def run(self):
         sorted_chunks = [output_chunks[key] for key in sorted(output_chunks)]
         law.pyarrow.merge_parquet_task(self, sorted_chunks, outputs["res"], local=True)
 
-        # save stats
```
riga, line 122: Can be uncommented again, the zero-division error is fixed by #45.

ap/selection/test.py
```diff
@@ -1,306 +1,452 @@
...
+    """
+    Selector function that performs cleaning of jets against leptons
+    based on a deltaR requirement
+    """
+    results = jet_lepton_deltaR_selection(events, stats, threshold)
```
riga, line 451:
```suggestion
    results = jet_lepton_deltaR_selection(events, stats, threshold=threshold)

```

riga (changes requested):
Yay, looks good to me! 🎉

I only have a couple picky style comments below :)

ap/selection/test.py
pkausw: OK, will do :+1:

ap/selection/test.py
```diff
@@ -1,306 +1,467 @@
...
+
+    return results
+
+# deltaR cleaning
+
+
+def cleaning_factory(
+    selector_name: str,
+    to_clean: str,
+    metric: Callable,
```
riga, line 315:
```suggestion
    metric: Optional[Callable] = None,
```

ap/selection/test.py
```diff
@@ -1,306 +1,467 @@
...
+    returns indices of good (clean) jets, sorted by pt
+    *TODO*: this should probably changed to return the boolean masks
+    """
+    return delta_r_jet_lepton(events, to_clean, clean_against, threshold=threshold)
+
+
+@selector(uses={delta_r_jet_lepton})
+def jet_lepton_delta_r_cleaning_selection(
+    events: ak.Array,
+    stats: Dict,
```
riga, line 437:
```suggestion
    stats: dict,
```

ap/selection/test.py
```diff
@@ -1,306 +1,467 @@
...
+        objects={"Jet": clean_jet_indices},
+    )
+
+
+@selector(uses={jet_lepton_delta_r_cleaning_selection})
+def jet_lepton_delta_r_cleaning(
+    events: ak.Array,
+    stats: Dict,
+    threshold: float = 0.4,
+    **kwargs: Dict,
```
riga, line 459:
```suggestion
    stats: dict,
    threshold: float = 0.4,
    **kwargs,
```

I think `kwargs` don't need hints.

riga (changes requested):

riga (changes requested):

ap/selection/test.py
pkausw:
```suggestion
    stats: Dict[str, Union[int, float]],
    threshold: float = 0.4,
    **kwargs,
```

ap/selection/test.py
```diff
@@ -1,306 +1,467 @@
...
-    return results
+# coding: utf-8
+
+"""
+Selection methods for testing purposes.
+"""
+
+from ap.selection import selector, SelectionResult
+from ap.util import maybe_import
+from typing import Callable, Dict, List, Optional
```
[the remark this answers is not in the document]
pkausw:
```suggestion
from typing import Callable, Dict, List, Optional, Union
```

riga (approved):
LGTM!

PR #40, opened 2022-06-15
Add calibrators for deterministic seeds

riga:
This PR adds calibrator functions that create deterministic event and object seeds (currently only for jets), based on combining integer features with primes, hashing and hex-digesting. These seeds can e.g. be used in smearing methods already applied during calibration (e.g. JER) or later on. The algorithm is can be written in one-line, making it easy to adapt for potential other groups for sync purposes.

When merged, this closes #36.

### DIFF
--- a/ap/calibration/seeds.py
+++ b/ap/calibration/seeds.py
@@ -1,0 +1,111 @@
+# coding: utf-8
+
+"""
+Methods related to creating event and object seeds during calibration.
+"""
+
+import hashlib
+
+from ap.calibration import calibrator
+from ap.util import maybe_import, primes
+from ap.columnar_util import Route, get_ak_indices, set_ak_column, has_ak_column
+
+np = maybe_import("numpy")
+ak = maybe_import("awkward")
+
+
+@calibrator(
+    uses={
+        # global columns for event seed
+        "event", "nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon", "nElectron", "nTau", "nSV",
+        # first-object columns for event seed
+        "Tau.jetIdx", "Tau.decayMode",
+        "Muon.jetIdx", "Muon.nStations",
+        "Jet.nConstituents", "Jet.nElectrons", "Jet.nMuons",
+    },
+    produces={"deterministic_seed"},
+)
+def deterministic_event_seeds(events, create_seed, **kwargs):
+    # start with a seed of the event number itself as the most sensitive integer
+    seed = create_seed(events.event)
+
+    # get global integers
+    global_fields = ["nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon", "nElectron", "nTau", "nSV"]
+    for i, f in enumerate(global_fields, 3):
+        seed = seed + primes[i] * (events[f] if f in events.fields else ak.num(events[f[1:]]))
+
+    # get first-object integers
+    object_fields = [
+        "Tau.jetIdx", "Tau.decayMode", "Muon.jetIdx", "Muon.nStations", "Jet.nConstituents",
+        "Jet.nElectrons", "Jet.nMuons",
+    ]
+    for i, f in enumerate(object_fields, i + 1):
+        values = events[Route(f).fields]
+        seed = seed + primes[i] * (ak.fill_none(ak.firsts(values, axis=1), -1) + 1)
+
+    # create and store them
+    seed = ak.Array(create_seed(seed))
+    set_ak_column(events, "deterministic_seed", seed)
+
+    # uniqueness test across the chunk for debugging
+    # n_events = len(seed)
+    # n_seeds = len(set(seed))
+    # match_text = "yes" if n_events == n_seeds else "NO !!!"
+    # print(f"events: {n_events}, unique seeds: {n_seeds}, match: {match_text}")
+
+    return events
+
+
+@deterministic_event_seeds.setup
+def deterministic_event_seeds_setup(self, task, inputs, call_kwargs, **kwargs):
+    # define the vectorized seed creation function once
+    # strategy:
+    #   1. Gather a selection of unambiguous integer features.
+    #   2. Multiply them with a vector of primes.
+    #   3. Use the resulting integer as an input to sha256 and hex-digest the result.
+    #   4. Reverse it and int-cast the leading 16 characters, leading to a 64 bit int.
+    def create_seed(n: int) -> int:
+        return int(hashlib.sha256(bytes(str(n), "utf-8")).hexdigest()[:-16:-1], base=16)
+
+    # store a vectorized version in the call_kwargs
+    call_kwargs["create_seed"] = np.vectorize(create_seed, otypes=[np.uint64])
+
+
+@calibrator(
+    uses={deterministic_event_seeds, "nJet"},
+    produces={"Jet.deterministic_seed"},
+)
+def deterministic_jet_seeds(events, create_seed, **kwargs):
+    # create the event seeds if not already present
+    if not has_ak_column(events, "deterministic_seed"):
+        events = deterministic_event_seeds(events, create_seed=create_seed, **kwargs)
+
+    # create the per jet seeds
+    jet_seed = events.deterministic_seed + primes[18] * get_ak_indices(events.Jet)
+    np_jet_seed = np.asarray(ak.flatten(jet_seed))
+    np_jet_seed[:] = create_seed(np_jet_seed)
+
+    # store them
+    set_ak_column(events, "Jet.deterministic_seed", jet_seed)
+
+    # uniqueness test across all jets in the chunk for debugging
+    # n_jets = ak.sum(ak.num(events.Jet, axis=1))
+    # n_seeds = len(set(np_jet_seed))
+    # match_text = "yes" if n_jets == n_seeds else "NO !!!"
+    # print(f"jets: {n_jets}, unique seeds: {n_seeds}, match: {match_text}")
+
+    return events
+
+
+@calibrator(
+    uses={deterministic_event_seeds, deterministic_jet_seeds},
+    produces={deterministic_event_seeds, deterministic_jet_seeds},
+)
+def deterministic_seeds(events, **kwargs):
+    # create the event seeds
+    events = deterministic_event_seeds(events, **kwargs)
+
+    # create the jet seeds
+    events = deterministic_jet_seeds(events, **kwargs)
+
+    return events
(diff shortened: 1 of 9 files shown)


ap/tasks/calibration.py
```diff
@@ -82,22 +82,22 @@ def run(self):
             msg = f"iterate through {reader.n_entries} events ..."
             for events, pos in self.iter_progress(reader, reader.n_chunks, msg=msg):
                 # just invoke the calibration function
-                arr = self.calibrator_func(events, **self.get_calibrator_kwargs(self))
+                events = self.calibrator_func(events, **self.get_calibrator_kwargs(self))
```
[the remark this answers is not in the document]
riga: Yeah, I think this was rather a typo before, but calling it the same should immediately invoke the garbage collector to free the memory consumed by the previous `events`.

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+        # global columns for event seed
+        "event", "nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon", "nElectron", "nTau", "nSV",
+        # first-object columns for event seed
+        "Tau.jetIdx", "Tau.decayMode",
+        "Muon.jetIdx", "Muon.nStations",
+        "Jet.nConstituents", "Jet.nElectrons", "Jet.nMuons",
+    },
+    produces={"deterministic_seed"},
+)
+def deterministic_event_seeds(events, create_seed, **kwargs):
```
pkausw, line 28: Might be nice to add a doc string and the typing here for better usability

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+    global_fields = ["nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon", "nElectron", "nTau", "nSV"]
+    for i, f in enumerate(global_fields, 3):
+        seed = seed + primes[i] * (events[f] if f in events.fields else ak.num(events[f[1:]]))
+
+    # get first-object integers
+    object_fields = [
+        "Tau.jetIdx", "Tau.decayMode", "Muon.jetIdx", "Muon.nStations", "Jet.nConstituents",
+        "Jet.nElectrons", "Jet.nMuons",
+    ]
+    for i, f in enumerate(object_fields, i + 1):
```
pkausw, line 42: It seems odd (and a little prone to errors) to me that the starting point of `enumerate` depends on the local variable of the loop in [line 33](https://github.com/uhh-cms/analysis_playground/pull/40/files#diff-34ad48c89852a085bc359ba5fdd0a38777a80f202215a459f1b0549b58d89f63R33). Is this on purpose? If yes, why not set the numbers separately, e.g.

```suggestion
# set the length separately
n_global_fields = len(global_fields)

# choose the same offset as in the event seed loop
for i, f in enumerate(object_fields, n_global_fields+3):
```

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+    },
+    produces={"deterministic_seed"},
+)
+def deterministic_event_seeds(events, create_seed, **kwargs):
+    # start with a seed of the event number itself as the most sensitive integer
+    seed = create_seed(events.event)
+
+    # get global integers
+    global_fields = ["nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon", "nElectron", "nTau", "nSV"]
+    for i, f in enumerate(global_fields, 3):
```
pkausw, line 34: Where is this number `3` coming from? I assume it's an arbitrary offset of some sort? Perhaps it would be nice to make this a variable that is centrally initialized somewhere (or even an argurment to this function with some default value). If it's not arbitrary, could you perhaps provide a comment on where this is coming from?

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+
+@deterministic_event_seeds.setup
+def deterministic_event_seeds_setup(self, task, inputs, call_kwargs, **kwargs):
+    # define the vectorized seed creation function once
+    # strategy:
+    #   1. gather a selection of unambiguous integer features
+    #   2. multiply them with a vector of primes
+    #   3. use the resulting integer as an input to sha256 and hex-digest the result
+    #   4. reverse it and int-cast the leading 16 characters, leading to a 64 bit int
+    # the method below performs steps 3 and 4 while 1 and 2 should be done outside
```
pkausw, line 67: Similar to other comment, might be nice to enable typing and put the comments here in a doc string

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+@calibrator(
+    uses={
+        # global columns for event seed
+        "event", "nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon", "nElectron", "nTau", "nSV",
+        # first-object columns for event seed
+        "Tau.jetIdx", "Tau.decayMode",
+        "Muon.jetIdx", "Muon.nStations",
+        "Jet.nConstituents", "Jet.nElectrons", "Jet.nMuons",
```
pkausw, line 24: Wouldn't it be better/more robust/more consistent to completely rely on Routes here?

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # get global integers
+    global_fields = ["nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon", "nElectron", "nTau", "nSV"]
+    for i, f in enumerate(global_fields, 3):
+        seed = seed + primes[i] * (events[f] if f in events.fields else ak.num(events[f[1:]]))
+
+    # get first-object integers
+    object_fields = [
+        "Tau.jetIdx", "Tau.decayMode", "Muon.jetIdx", "Muon.nStations", "Jet.nConstituents",
+        "Jet.nElectrons", "Jet.nMuons",
```
pkausw, line 40: Similar to comment above, wouldn't be Routes better/more consistent to create these strings?

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+    def create_seed(n: int) -> int:
+        return int(hashlib.sha256(bytes(str(n), "utf-8")).hexdigest()[:-16:-1], base=16)
+
+    # store a vectorized version in the call_kwargs
+    call_kwargs["create_seed"] = np.vectorize(create_seed, otypes=[np.uint64])
+
+
+@calibrator(
+    uses={deterministic_event_seeds, "nJet"},
+    produces={"Jet.deterministic_seed"},
```
pkausw, line 77: See other comments regarding the usage of Route class

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+)
+def deterministic_jet_seeds(events, create_seed, **kwargs):
+    # create the event seeds if not already present
+    if not has_ak_column(events, "deterministic_seed"):
+        events = deterministic_event_seeds(events, create_seed=create_seed, **kwargs)
+
+    # create the per jet seeds
+    jet_seed = events.deterministic_seed + primes[18] * get_ak_indices(events.Jet)
```
pkausw, line 85: Might be nice to make the index of the list of primes part of the argument list or externalize this

ap/calibration/seeds.py
```diff
@@ -0,0 +1,112 @@
...
+    np_jet_seed = np.asarray(ak.flatten(jet_seed))
+    np_jet_seed[:] = create_seed(np_jet_seed)
+
+    # store them
+    set_ak_column(events, "Jet.deterministic_seed", jet_seed)
```
pkausw, line 90: Also use a Route object in this case?

pkausw (changes requested):
Hi, very nice PR! I have a couple of questions, but nothing critical I think :+1:

ap/columnar_util.py
```diff
@@ -721,7 +757,7 @@ def used_columns(self) -> Set[str]:
 
         # add those of all other known intances
         for obj in self.uses:
-            if isinstance(obj, self.__class__):
```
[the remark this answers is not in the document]
riga: Yup, I changed that 👍

ap/util.py
```diff
@@ -32,6 +32,15 @@
 #: Boolean denoting whether the environment is used for development (based on ``AP_DEV``).
 env_is_dev = not env_is_remote and law.util.flag_to_bool(os.getenv("AP_DEV", "0"))
 
+#: List of the first 100 primes.
+primes = [
```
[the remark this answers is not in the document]
riga: True. Would you be ok with keeping only 100 for now, and adapt this once we need more? :)

ap/calibration/seeds.py
riga: ✓

ap/calibration/seeds.py
riga: Yeah, this was kinda of on purpose but I agree that it's not very clean to use the "leaking" index from the previous loop. I'll add the change you suggested (just with an inline call to `len`).

ap/calibration/seeds.py
riga: ✓

ap/columnar_util.py
```diff
@@ -596,6 +597,41 @@ def sorted_ak_to_parquet(
...
+        offsets = np.array(ak_array.layout.offsets, dtype=np.uint64)[:-1]
+
+        # obtain indices by subtracting per-range offsets
+        indices = single_range_with_offsets - offsets
+
+        return indices
+
+    raise ValueError(f"only arrays with 1 or 2 dimensions are supported, found {ak_array.ndim}")
+
+
```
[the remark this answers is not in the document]
riga: Oh, I was looking for that one, but couldn't find it -.-
Thanks, the helper is removed!

ap/calibration/seeds.py
riga: ✓

ap/calibration/seeds.py
riga: In this case, it was just a choice that we would not change. Therefore I didn't want to make it a parameter / variable in order to not make the impression that it should be changed.

ap/tasks/calibration.py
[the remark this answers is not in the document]
pkausw: OK, thanks for the explanation :+1:

ap/columnar_util.py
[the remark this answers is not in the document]
pkausw: :+1:

ap/util.py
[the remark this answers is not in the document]
pkausw: Absolutely :+1:

pkausw (approved):

Issue #25, opened 2022-06-02
[permanent] Reiterate task structure


riga:
This issue is intended to be open for a longer period in order to interactively plan and discuss the task structure.

(at least as long as GitHub has no support yet for mermaid graphs in wikis)

### Legend

```mermaid
graph TD
    classDef WF stroke: #83b, stroke-width: 3px
    classDef TODO fill: #f17925
    classDef DEC fill: #0aacfc
    
    subgraph Legend
        Task(Task) --> Workflow(Workflow)
        Workflow --> TODO(TODO)
        class TODO TODO
        class Workflow WF
        
        X0(X) --> D0{Decision}
        D0 -- yes --> Y0s(Y)
        D0 -- no --> Z0(Z)
        class D0 DEC
    
        X1(X) -- single<br />dep. ---> Y1(Y)
        X2(X) -. 0 to n<br />deps. ..-> Y2(Y)
        X3(X) == merges<br />many ===> Y3(Y)
    end
```

### Graph

```mermaid
graph TD
    classDef WF stroke: #83b, stroke-width: 3px
    classDef TODO fill: #f17925
    classDef DEC fill: #0aacfc

    GetDatasetLFNs(GetDatasetLFNs)
    CalibrateEvents(CalibrateEvents)
    SelectEvents(SelectEvents)
    MergeSelectionMasks(MergeSelectionMasks)
    MergeSelectionStats(MergeSelectionStats)
    CreateCutflowHistograms(CreateCutflowHistograms)
    ReduceEvents(ReduceEvents)
    MergeReductionStats(MergeReductionStats)
    MergeReducedEvents(MergedReducedEvents)
    ProduceColumns(ProduceColumns)
    UniteColumns(UniteColumns)
    CreateHistograms(CreateHistograms)
    TrainDataSwitch{Used in<br />training?}
    PrepareMLEvents(PrepareMLEvents)
    MergeMLEvents(MergeMLEvents)
    MLTraining(MLTraining)
    MLEvaluation(MLEvaluation)
    MergeHistograms(MergeHistograms)
    MergeShiftedHistograms(MergeShiftedHistograms)
    MergeMLEvaluation(MergeMLEvaluation)
    WriteDatacards(WriteDatacards)
    WritePyhfWorkspace(WritePyhfWorkspace)
    CreateYieldTable(CreateYieldTable)
    PlotVariables1D(PlotVariables1D)
    PlotVariables2D(PlotVariables2D)
[...]

PR #45, opened 2022-06-16
Allow selection to produce columns, closes #44.

### LINKED ISSUE #44: Allow production of columns with selection through actual producers
Right now, the way that new columns can be created is somewhat two-fold. On the one hand, `SelectionResult`'s can contain new columns that will be injected into saved selection masks. On the other hand, we have the generic `producer` interface that does the same but a whole lot nicer. At the same time, `selector`'s have a `produces` member which is not utilized whatsover.

We should allow producers being called from within selectors and allow selection functions to return either

- a single value, being the `SelectionResult` (but a variant **without** new columns), or
- a 2-tuple, being the `SelectionResult` and additional columns in a separate array.

This would make much more sense.

@pkausw @dsavoiu

riga:
This PR tackles the ambiguity described in #44.

>  Right now, the way that new columns can be created is somewhat two-fold. On the one hand, SelectionResult's can contain new columns that will be injected into saved selection masks. On the other hand, we have the generic producer interface that does the same but a whole lot nicer. At the same time, selector's have a produces member which is not utilized whatsover.

Changes in this PR:

- `SelectionResult` does not longer have a field for new columns.
- Selectors now can return a single `SelectionResult` or a 2-tuple (`SelectionResult`, events)
- When the selector configured for the `SelectEvents` task produces columns, `SelectEvents` will have a third output representing the produced columns (besides the selection masks and the stats file). **However**, we should only use that feature if it's really necessary (e.g. for saving columns that are required for the selection step, but are otherwise expensive to reproduce) and adhere to the `ProduceColumns` task if possible (see the task structure in #25).
- Current selectors are adapted to the above structure.

At some point, we should make some of the selectors which are actually producing sth, actual producers :)

@mafrahm

### DIFF
--- a/ap/tasks/reduction.py
+++ b/ap/tasks/reduction.py
@@ -26,15 +26,15 @@ class ReduceEvents(DatasetTask, CalibratorsSelectorMixin, law.LocalWorkflow, HTC
         reqs = super().workflow_requires()
         reqs["lfns"] = GetDatasetLFNs.req(self)
         if not self.pilot:
-            reqs["calib"] = [CalibrateEvents.req(self, calibrator=c) for c in self.calibrators]
-            reqs["sel"] = SelectEvents.req(self)
+            reqs["calibrations"] = [CalibrateEvents.req(self, calibrator=c) for c in self.calibrators]
+            reqs["selection"] = SelectEvents.req(self)
         return reqs
 
     def requires(self):
         return {
             "lfns": GetDatasetLFNs.req(self),
-            "calib": [CalibrateEvents.req(self, calibrator=c) for c in self.calibrators],
-            "sel": SelectEvents.req(self),
+            "calibrations": [CalibrateEvents.req(self, calibrator=c) for c in self.calibrators],
+            "selection": SelectEvents.req(self),
         }
 
     def output(self):
@@ -77,10 +77,10 @@ class ReduceEvents(DatasetTask, CalibratorsSelectorMixin, law.LocalWorkflow, HTC
 
         # iterate over chunks of events and diffs
         input_paths = [nano_file]
-        input_paths.append(inputs["sel"]["res"].path)
-        input_paths.extend([inp.path for inp in inputs["calib"]])
+        input_paths.append(inputs["selection"]["results"].path)
+        input_paths.extend([inp.path for inp in inputs["calibrations"]])
         if self.selector_func.produced_columns:
-            input_paths.append(inputs["sel"]["cols"].path)
+            input_paths.append(inputs["selection"]["columns"].path)
         with ChunkedReader(
             input_paths,
             source_type=["coffea_root"] + (len(input_paths) - 1) * ["awkward_parquet"],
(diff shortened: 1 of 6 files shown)


ap/tasks/reduction.py
```diff
@@ -76,30 +76,38 @@ def run(self):
             nano_file = input_file.load(formatter="uproot")
 
         # iterate over chunks of events and diffs
+        input_paths = [nano_file]
+        input_paths.append(inputs["sel"]["res"].path)
+        input_paths.extend([inp.path for inp in inputs["calib"]])
+        if self.selector_func.produced_columns:
+            input_paths.append(inputs["sel"]["cols"].path)
```
pkausw, line 83: See other comments regarding descriptive power of keyword arguments

```suggestion
            input_paths.append(inputs["selection"]["columns"].path)
```

ap/tasks/reduction.py
```diff
@@ -76,30 +76,38 @@ def run(self):
...
+                events = events[event_mask]
 
                 # apply all object masks whose names are present
+                # TODO: this should not be done automagically based on names as there might be
+                #       multiple collections of the same object type
                 for name in sel.objects.fields:
                     if name in events.fields:
                         # apply the event mask to the object mask first
-                        object_mask = sel.objects[name][sel.event]
+                        object_mask = sel.objects[name][event_mask]
```
pkausw, line 110: This depends on wether the selection masks in `selection` can be stacked, right? But I guess this shouldn't break even if the selections were stacked before 🤔

ap/tasks/selection.py
```diff
@@ -46,25 +46,32 @@ def requires(self):
         return reqs
 
     def output(self):
-        return {
+        outputs = {
             "res": self.local_target(f"results_{self.branch}.parquet"),
             "stat": self.local_target(f"stats_{self.branch}.json"),
         }
```
pkausw, line 52: see comment in file `ap/tasks/reduction.py` regarding descriptive power of keys of dictionary. Wouldn't it be better to make it clearer?
```suggestion
        outputs = {
            "result": self.local_target(f"results_{self.branch}.parquet"),
            "stats": self.local_target(f"stats_{self.branch}.json"),
        }

File: ap/tasks/selection.py [hunk 4]
```diff
@@ -46,25 +46,32 @@ def requires(self):
...
 
+        # add additional columns in case the selector produces some
+        if self.selector_func.produced_columns:
+            outputs["cols"] = self.local_target(f"columns_{self.branch}.parquet")
```
pkausw, line 56: See 'descriptive power of keys' above
```suggestion
            outputs["columns"] = self.local_target(f"columns_{self.branch}.parquet")
```

pkausw (changes requested):

ap/tasks/reduction.py
```diff
@@ -76,30 +76,38 @@ def run(self):
             nano_file = input_file.load(formatter="uproot")
 
         # iterate over chunks of events and diffs
+        input_paths = [nano_file]
+        input_paths.append(inputs["sel"]["res"].path)
```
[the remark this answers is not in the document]
riga: Yep, added that ✓

ap/tasks/selection.py
riga: Yep, added that ✓

ap/tasks/selection.py
riga: Yep, added that ✓

ap/tasks/reduction.py
pkausw: OK :+1:

PR #46, opened 2022-06-17
Automate event weights.

riga:
This PR adds two new producers for creating event weights and removes some hard-coded (yet recently needed) assumptions from the `CreateHistograms` task, which now allow to create and apply event weights automated on-the-fly.

Changes in detail:

- Added `normalization_weights` to `production/normalization.py`. This producer takes the lumi from the config, the xs from the dataset (at least for now) and the sum of gen weights from selection stats to assign normalization weights. 
- Added `event_weights` to `production/weights.py`. This producer is just a wrapper around all the weights we currently have. In the future, this one might be analysis-dependent and should not be located in the "core" (or however we want to call it). Also, it defines all possible shifts that any of the weights might depend on.
- Added a new config aux entry "event_weights" which is a list of event weight columns to apply multiplicatively.
- Adjusted the `CreateHistograms` task to
  - not call the `variables` producer in a hard-wired way (rather add `variables` to `--producers` now), and
  - read the "event_weights" config entry, load the corresponding columns, multiply them and use this weight for the histogramming.
- For the moment, I set the `n_files` fields of all datasets in the campaign to 2 in order to test faster.

Also, since I was streamlining the producers, I took the liberty to move the `variables` selector into `production/features.py` and made it an actual producer while I was on it.

I nice command to test this is

```shell
law run PlotShifts \
    --version v7 \
    --producers event_weights,variables \
    --shift-sources minbias_xs \
    --view-cmd imgcat \
    --workers 4
```

([imgcat](https://iterm2.com/documentation-images.html) ftw)

@dsavoiu @pkausw @mafrahm


ap/columnar_util.py
```diff
@@ -36,9 +36,12 @@
 pq = maybe_import("pyarrow.parquet")
 
 
-#: columns that are always required when opening a nano file with coffea
+#: Columns that are always required when opening a nano file with coffea.
 mandatory_coffea_columns = {"run", "luminosityBlock", "event"}
 
+#: Empty-value definition in places where a number is expected.
+EMPTY = -99999
```
pkausw, line 43: Just to make sure: This quantity is then also accessible to and can be loaded by other modules, right? It might be nice to have this globally accessible :+1:

ap/production/features.py
```diff
@@ -0,0 +1,47 @@
...
+        "Electron.pt", "Electron.eta", "Muon.pt", "Muon.eta", "Jet.pt", "Jet.eta",
+        "Jet.btagDeepFlavB",
+    },
+    produces={"HT", "nJet", "nElectron", "nMuon", "nDeepjet", "Electron1_pt", "Muon1_pt"} | {
+        f"Jet{i}_{attr}"
+        for i in range(1, 4 + 1)
+        for attr in ["pt", "eta"]
+    },
+)
+def variables(events: ak.Array, **kwargs) -> ak.Array:
```
pkausw, line 34: This is an optional comment: I realize that this producer will probably be something analysis-dependent and should therefore not be part of the analysis playground core, but it might be nice to add a doc string to producers that explains what columns the producer creates. I propose to already start this in this early stage as future reference for other people. 

Now that I think about it, it might be beneficial to have a producer/producers that compute common variables that can be used in many analyses, e.g. HT 🤔

ap/production/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+ak = maybe_import("awkward")
+
+
+@producer(
+    uses={"LHEWeight.originalXWGTUP"},
+    produces={"normalization_weight"},
+)
+def normalization_weights(
+    events: ak.Array,
+    selection_stats: dict,
```
pkausw, line 23: Might be nice to type hint the contents of the selection_stats dict

ap/production/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # compute the weight and store it
+    norm_weight = events.LHEWeight.originalXWGTUP * lumi * xs / selection_stats["sum_mc_weight"]
+    set_ak_column(events, "normalization_weight", norm_weight)
+
+    return events
+
+
+@normalization_weights.requires
+def normalization_weights_requires(self: Producer, task: law.Task, reqs: dict) -> dict:
```
pkausw, line 47: Some comment as above regarding the type hinting of the contents of reqs and the return dict

ap/production/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+    task: law.Task,
+    inputs: dict,
+    call_kwargs: dict,
+    **kwargs,
+):
+    """
+    Loads the selection stats added through the requirements and saves them in the *call_kwargs* for
+    simpler access in the actual callable.
+    """
+    call_kwargs["selection_stats"] = inputs["selection_stats"].load(formatter="json")
```
pkausw, line 72: I'm wondering about this. Propose you have a producer that also employs a selection requirement. This would also change the `selection_stats`, right? Can we account for something like this with the current structure?

ap/production/pileup.py
```diff
@@ -4,41 +4,58 @@
...
 from ap.util import maybe_import
-from ap.production import producer
 from ap.columnar_util import set_ak_column
 
 ak = maybe_import("awkward")
 
 
 @producer(
     uses={"PV.npvs"},
-    produces={"pu_weight.nominal", "pu_weight.minbias_xs_up", "pu_weight.minbias_xs_down"},
```
pkausw, line 16: This is a matter of taste of course, but I actually liked the dot naming scheme here :D Of course one could argue that the `pu_weight` isn't a field and that therefore the names should only contain underscore 🤔

pkausw (changes requested):

ap/production/features.py
mafrahm: I think a set of default or exemplary producers/selectors/variables would be nice to 'quickstart' an analysis / get to know the framework for newcomers

ap/production/normalization.py
riga: The requirements dict is fully arbitrary as is the return value, so I wouldn't assume and hint any structure here.

ap/production/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+    normalization weight.
+    """
+    # get lumi and cross section of the first process contained in the dataset
+    # TODO: the process should depend on the process id per event for which we need a LUT
+    #       that is cached through a simple mutable argument
+    lumi = config_inst.x.luminosity.nominal
+    xs = dataset_inst.processes.get_first().get_xsec(config_inst.campaign.ecm).nominal
+
+    # compute the weight and store it
+    norm_weight = events.LHEWeight.originalXWGTUP * lumi * xs / selection_stats["sum_mc_weight"]
```
mafrahm, line 40: I might need to check that again, but I think the process cross section is already taken into account by the LHEWeight.

ap/production/normalization.py
riga: Since the `MergeSelectionStats` is required as

```python
MergeSelectionStats.req(task)
``` 

the internal choice of the `--selector` is passed from `task` to the `MergeSelectionStats` that will be created which should always be consistent and lead to the correct stats. Also, in this place, one only cares about the sum of event weights *before* any selection, so it's correct in any case :)

ap/production/normalization.py
mafrahm: Oh nevermind, you are dividing by the sum of weights, so this is already taken into account

ap/production/pileup.py
riga: Yeah, that's true. I switched to underscore (i.e. flat and not nested under `pu_weight`) so that the alias when applying this shift is simpler. I changed it to

```python
add_aliases("minbias_xs", {"pu_weight": "pu_weight_{name}"})
```

in this PR. With dot syntax it would have been

```python
add_aliases("minbias_xs", {"pu_weight.nominal": "pu_weight.{name}"})
```

which would read as "explicitly alias to nominal". In this sense, the "new" alias is better to read and consistent with e.g. "Jet.pt" vs "Jet.pt_jec_up".

ap/production/normalization.py
riga: Yup, added in c84111a2.

ap/production/features.py
riga: Yup, second it. At some point we're going to strip away the analysis-dependent implementations but in the meantime, it's nice to be able to plot sth that is not there from there start :)

In the long term, we should of course enforce doc strings as @pkausw suggests, especially for "core" stuff.

ap/production/normalization.py
riga: Oh I hope not. The cross sections from the database are usually not the ones one wants to use.

I remember from (mini)AOD times that "originalXWGTUP" used to be the generator dependent one without xs. I hope this didn't change with nanoAOD ...

ap/selection/test.py
```diff
@@ -14,17 +14,8 @@
...
-    array = ak.fill_none(array[:, idx], -999)
-    return array
-
-
 # object definitions
-# TODO: Producer instead of Selector
+# TODO: return masks instead of indices and build indices as late as possible
 @selector(uses={"Jet.pt", "Jet.eta"})
 def req_jet(events):
     mask = (events.Jet.pt > 30) & (abs(events.Jet.eta) < 2.4)
```
mafrahm, line 21: Just as a side note: this is also the point where the bug in the selection (e.g. selecting jets with pt<30GeV) was from. The argsort also needs to be applied on the mask, e.g. like that:

https://github.com/mafrahm/analysis_playground/blob/fix_object_req_argsort/ap/selection/test.py#L28-L29

riga: Yep, good point!

@pkausw Was planning to change the selectors that currently return indices to return masks again, and then apply them in the correct way you suggest in a consistent manner :)

mafrahm (approved):

pkausw (approved):

Issue #21, opened 2022-06-01
Externalize plotting functions


riga:
Currently, plotting logic is mixed with task logic, e.g. in ap/tasks/plotting.py. We should have a bunch of plotting functions in ap/plots that can be imported from within tasks and called without much overhead, simply forwarding information to the plot functions.

PR #61, opened 2022-07-06
WIP: implementation of cutflow plots workflow

mafrahm:
This PR adds the first implementation of the workflow to create cutflow plots. The workflow contains three tasks:

- `MergeSelectionMasks` merges all selection steps and columns defined by the selector into a single file per dataset. It also calls the `normalization_weights` Producer and therefore requires the `process_id` and `LHEWeight.originalXWGTUP` columns to be present in the output of the `SelectEvents` task (hard-coded in the `test` Selector for now)
- `CreateCutflowHistograms` creates histograms analogous to the `CreateHistograms` task but adds one additional `StrCategory` axis "selection\_steps". Each event that survives the selection up to this step is filled into the affiliated bin. "noCuts" is hard-coded as the first element of the "selection_steps" axis.
- `PlotCutflow` is a first implementation to create cutflow plots. At the moment, there is a large overlap of code with the other plotting tasks. The code duplication will be removed and customization options will be added as part of a pull request for the #21 issue. In the future, this task (or an additional task) should also allow to plot variables after applying each individual cut.

Command for testing:
`law run PlotCutflow --version c11 --processes st,tt_sl,tt_dl,tt_fh --categories incl --workers 4`

Minor changes:
- the `SelectorStepsMixin` is updated to allow the definition of defaults in the config
- an "incl" category is added to allow selecting all events in a plotting task
- `process_id`, `LHEWeight.originalXWGTUP` and `ht` (as an exemplary variable) are stored as columns by the `test` Selector
- `normalization_weights` only calls `process_ids` if the process_id field is not present

The WIP label is added due to the fact that the requirement of producing `LHEWeight.originalXWGTUP` in `SelectEvents` leads to an error in `ReduceEvents` when calling `update_ak_array` [1]. This problem should be solved before merging.

[1] https://github.com/uhh-cms/analysis_playground/blob/feature/cutflow_plots/ap/tasks/reduction.py#L101

### DIFF
--- a/ap/config/analysis_st.py
+++ b/ap/config/analysis_st.py
@@ -71,6 +71,12 @@ for dataset_name in dataset_names:
     config_2018.add_dataset(campaign_2018.get_dataset(dataset_name))
 
 
+# default calibrator, selector, producer and ml_model
+config_2018.set_aux("default_calibrator", "test")
+config_2018.set_aux("default_selector", "test")
+config_2018.set_aux("default_producer", "variables")
+config_2018.set_aux("default_ml_model", None)
+
 # process groups for conveniently looping over certain processs
 # (used in wrapper_factory and during plotting)
 config_2018.set_aux("process_groups", {})
@@ -91,14 +97,11 @@ config_2018.set_aux("variable_groups", {})
 # (used during plotting)
 config_2018.set_aux("shift_groups", {})
 
-# default calibrator, selector, producer and ml_model
-config_2018.set_aux("default_calibrator", "test")
-config_2018.set_aux("default_selector", "test")
-config_2018.set_aux("default_selector_steps", {
-    "test": ("Lepton", "Jet", "Deepjet"),
+# selector step groups for conveniently looping over certain steps
+# (used in cutflow tasks)
+config_2018.set_aux("selector_step_groups", {
+    "test": ["Lepton", "Jet", "Deepjet"],
 })
-config_2018.set_aux("default_producer", "variables")
-config_2018.set_aux("default_ml_model", None)
 
 # 2018 luminosity with values in inverse pb and uncertainties taken from
 # https://twiki.cern.ch/twiki/bin/view/CMS/TWikiLUM?rev=171#LumiComb
@@ -309,8 +312,8 @@ config_2018.set_aux("keep_columns", DotDict.wrap({
         "PV.npvs",
         "category_ids", "deterministic_seed",
     },
-    "CreateHistograms": {
-        "LHEWeight.originalXWGTUP",
+    "MergeSelectionMasks": {
+        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids",
     },
 }))
 
--- a/ap/config/analysis_st.py
+++ b/ap/config/analysis_st.py
@@ -91,10 +91,11 @@ config_2018.set_aux("variable_groups", {})
 # (used during plotting)
 config_2018.set_aux("shift_groups", {})
 
-# default calibrator, selector and producer
+# default calibrator, selector, producer and ml_model
 config_2018.set_aux("default_calibrator", "test")
 config_2018.set_aux("default_selector", "test")
 config_2018.set_aux("default_producer", "variables")
+config_2018.set_aux("default_ml_model", None)
 
 # 2018 luminosity with values in inverse pb and uncertainties taken from
 # https://twiki.cern.ch/twiki/bin/view/CMS/TWikiLUM?rev=171#LumiComb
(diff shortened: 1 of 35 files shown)


ap/config/categories.py
```diff
@@ -11,6 +11,13 @@ def add_categories(config: od.Config) -> None:
     """
     Adds all categories to a *config*.
     """
+    config.add_category(
+        name="incl",
+        expression="sel_incl",
+        id=99,
+        label="inclusive",
+        selection="sel_incl",
+    )
```
riga, line 20: Btw: I made a mistake in order. Categories should not have expressions but only selections. I reverted this and adapted the code in a commit to this branch. Hope that is ok for you.

ap/production/normalization.py
```diff
@@ -39,7 +39,8 @@ def normalization_weights(
     normalization weight.
     """
     # add process ids
-    self.stack.process_ids(events, config_inst=config_inst, dataset_inst=dataset_inst, **kwargs)
+    if "process_id" not in events.fields:
+        self.stack.process_ids(events, config_inst=config_inst, dataset_inst=dataset_inst, **kwargs)
```
riga, line 43: I think it's okay to always process them. Internally there is a cache that checks of a producer already ran, and if yes, doesn't run it again. it would also catch this one here.

```suggestion
    self.stack.process_ids(events, config_inst=config_inst, dataset_inst=dataset_inst, **kwargs)
```

ap/tasks/framework/mixins.py
```diff
@@ -235,6 +235,14 @@ class SelectorStepsMixin(SelectorMixin):
     def modify_param_values(cls, params: Dict[str, Any]) -> Dict[str, Any]:
         params = super().modify_param_values(params)
 
+        # add the default selector_steps when empty
+        if "config_inst" in params and params.get("selector_steps") == ():
+            config_inst = params["config_inst"]
+            selector = params.get("selector", None)
+            if config_inst.x("default_selector_steps", None):
+                if selector in config_inst.x.default_selector_steps:
+                    params["selector_steps"] = config_inst.x.default_selector_steps[selector]
```
riga, line 244: See the first review comment. I think having default selection steps might not be convenient in the long run.

ap/selection/test.py
```diff
@@ -269,6 +276,12 @@ def test(
     # build categories
     self.stack.category_ids(events, config_inst=config_inst, dataset_inst=dataset_inst, **kwargs)
 
+    # for cutflow plots
+    self.stack.process_ids(events, dataset_inst=dataset_inst, **kwargs)
+    # calculating variables before and after applying object cleaning requires two definitions
+    # of the variable; might lead to ambiguities if not careful
+    set_ak_column(events, "ht", self.stack.var_HT(events))
```
riga, line 283: Mh, why again do you need the variable? Just to make some last cuts later on?

ap/selection/test.py
```diff
@@ -269,6 +276,12 @@ def test(
     # build categories
     self.stack.category_ids(events, config_inst=config_inst, dataset_inst=dataset_inst, **kwargs)
 
+    # for cutflow plots
+    self.stack.process_ids(events, dataset_inst=dataset_inst, **kwargs)
```
riga, line 280: Just quick suggestion as we will always need process IDs during the selection in the future (for saving event weight sums per process id for sample stitching).

```suggestion
    # create process ids
    self.stack.process_ids(events, dataset_inst=dataset_inst, **kwargs)

```

ap/config/analysis_st.py
```diff
@@ -94,6 +94,9 @@
 # default calibrator, selector, producer and ml_model
 config_2018.set_aux("default_calibrator", "test")
 config_2018.set_aux("default_selector", "test")
+config_2018.set_aux("default_selector_steps", {
+    "test": ("Lepton", "Jet", "Deepjet"),
+})
```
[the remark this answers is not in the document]
mafrahm: writing all steps in the command line sounds quite tedious to me (depending on the number of steps defined by the selector). However, I'm assuming that one could simply define a selector_steps_groups auxiliary in the config. Using such a ground in the command line sounds acceptable to me

ap/selection/test.py
mafrahm: Yes, for the simple cutflow plots, no additional variables are required. However, I built the histogram task such that it also creates a variable axis to allow creating ordinary plots as part of the cutflow workflow. But I should probably also define a default when no variable is stored.

ap/tasks/framework/mixins.py
mafrahm: yes, the default will be removed and instead, the usage selector_steps_groups will be enabled

PR #65, opened 2022-07-13
Add producer identifying gen top decay chains

### LINKED ISSUE #18: Explore generator information
Several gen studies, proper truth definitions for ML applications, and dataset stitching procedures require a good understanding of the event in gen level. We need some expertise in how to jump through decay chains and find (e.g.) decay products of certain particles. An example application would be to find the neutrino of the hard, leptonic W decay.

riga:
This PR adds a producer that shows how to extract generator information. It is meant as an example and starting point for further generator studies.

In particular, the producer in this PR, `gen_top_decay_products`, creates a new nested column `gen_top_decay` containing a coffea `GenParticleArray` per event and per generator top quark. Each array consists of exactly 5 `GenParticle`s in a specific order:

- top quark,
- the W boson,
- bottom quark,
- down-type quark or charged lepton from W decay, and
- up-type quark or neutrino from W decay.

Closes #18.

### DIFF
--- a/ap/config/campaign_2018.py
+++ b/ap/config/campaign_2018.py
@@ -46,6 +46,7 @@ with uniqueness_context(campaign_2018.name):
         name="st_tchannel_t",
         id=1,
         processes=[procs.process_st_tchannel_t],
+        aux={"has_top": True},
         info=dict(
             nominal=DatasetInfo(
                 keys=[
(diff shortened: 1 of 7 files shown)


riga:
For some reason, the PR is broken and I'll reopen.

PR #67, opened 2022-07-14
Externalize plotting

### LINKED ISSUE #21: Externalize plotting functions

mafrahm:
This PR externalizes the plotting logic in tasks into separate functions. Two files are added:

- In `plotter.py` the generalized plotting functions are located. The `plot_all` function can be called to create a variety of plots based on the input that is given. The inputs are:
    - `plot_config`, which expects dictionaries with keys "method" (e.g. `draw_from_stack`), "hist", "kwargs" (to override defaults from "method") and "ratio_kwargs" (to override defaults from "method" when drawing the ratio plot)
    - `style_config`, which overrides the defaults from the axes (`ax_cfg` and `rax_cfg`), the legend (`legend_cfg`) and the CMS label (`CMS_label_cfg`)
- in `variables.py`, plotting functions for different types of plots are defined to call the `plot_all` function. These functions are used in the plotting tasks.

Commands for testing:
- PlotVariables:
`law run PlotVariables --version test --variables jet1_pt --processes tt_sl,st_tchannel --categories incl --workers 6`
- PlotShiftedVariables:
`law run PlotShiftedVariables --version test --variables jet1_pt --processes tt_sl --categories incl --shift-sources minbias_xs --workers 6`
PlotCutflow:
`law run PlotCutflow --version test --processes tt_sl,st_tchannel --categories incl --selector-steps Jet,Lepton,Deepjet --workers 6`

Closes #21

### DIFF
--- a/ap/plotting/plotter.py
+++ b/ap/plotting/plotter.py
@@ -4,7 +4,9 @@
 Generalized plotting functions to create plots from hist histograms
 """
 
-from ap.util import maybe_import
+from typing import Optional, Sequence
+
+from ap.util import maybe_import, test_float
 
 np = maybe_import("numpy")
 hist = maybe_import("hist")
@@ -12,7 +14,9 @@ plt = maybe_import("matplotlib.pyplot")
 mplhep = maybe_import("mplhep")
 
 
-def draw_error_stairs(ax, h, kwargs={}):
+def draw_error_stairs(ax: plt.Axes, h: hist.Hist, kwargs: Optional[dict] = None) -> None:
+    if kwargs is None:
+        kwargs = {}
     norm = kwargs.pop("norm", 1)
     values = h.values() / norm
     error = np.sqrt(h.variances()) / norm
@@ -29,14 +33,18 @@ def draw_error_stairs(ax, h, kwargs={}):
     ax.stairs(**defaults)
 
 
-def draw_from_stack(ax, h, kwargs={}):
+def draw_from_stack(ax: plt.Axes, h: hist.Stack, kwargs: Optional[dict] = None) -> None:
+    if kwargs is None:
+        kwargs = {}
+
     norm = kwargs.pop("norm", 1)
 
     # check if norm is a number
-    try:
-        int(norm)
+    if test_float(norm):
         h = hist.Stack(*[i / norm for i in h])
-    except:
+    else:
+        if not isinstance(norm, Sequence) and not isinstance(norm, np.ndarray):
+            raise TypeError(f"norm must be either a number, sequence or np.ndarray, not a {type(norm)}")
         norm = np.array(norm)
         # 1 normalization factor/array per histogram
         if len(norm) == len(h):
@@ -56,7 +64,10 @@ def draw_from_stack(ax, h, kwargs={}):
     h.plot(**defaults)
 
 
-def draw_from_hist(ax, h, kwargs={}):
+def draw_from_hist(ax: plt.Axes, h: hist.Hist, kwargs: Optional[dict] = None) -> None:
+    if kwargs is None:
+        kwargs = {}
+
     norm = kwargs.pop("norm", 1)
     h = h / norm
     defaults = {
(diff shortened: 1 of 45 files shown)


ap/plotting/plotter.py
```diff
@@ -0,0 +1,181 @@
...
+
+from ap.util import maybe_import
+
+np = maybe_import("numpy")
+hist = maybe_import("hist")
+plt = maybe_import("matplotlib.pyplot")
+mplhep = maybe_import("mplhep")
+
+
+def draw_error_stairs(ax, h, kwargs={}):
```
riga, line 15: Could you rather use `kwargs=None` as a default and the set it to an empty dict when None?

```python
def draw_error_stairs(ax, h, kwargs=None):
    if kwargs is None:
        kwargs = {}
```

Mutable default arguments can become [nasty](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) 👀

ap/plotting/plotter.py
```diff
@@ -0,0 +1,181 @@
...
+        "hatch": "///",
+        "facecolor": "none",
+        "linewidth": 0,
+        "color": "black",
+    }
+    defaults.update(kwargs)
+    ax.stairs(**defaults)
+
+
+def draw_from_stack(ax, h, kwargs={}):
```
riga, line 32: Same as above

ap/plotting/plotter.py
```diff
@@ -0,0 +1,181 @@
...
+    norm = kwargs.pop("norm", 1)
+
+    # check if norm is a number
+    try:
+        int(norm)
+        h = hist.Stack(*[i / norm for i in h])
+    except:
```
riga, line 39: To avoid unforseen exceptions from being falsely handled here, could you catch exact exception types?

ap/plotting/plotter.py
```diff
@@ -0,0 +1,181 @@
...
+    defaults = {
+        "ax": ax,
+        "stack": True,
+        "histtype": "fill",
+    }
+    defaults.update(kwargs)
+    h.plot(**defaults)
+
+
+def draw_from_hist(ax, h, kwargs={}):
```
riga, line 59: Same as above

ap/plotting/plotter.py
```diff
@@ -0,0 +1,181 @@
...
+    defaults = {
+        "ax": ax,
+        "stack": False,
+        "histtype": "step",
+    }
+    defaults.update(kwargs)
+    h.plot1d(**defaults)
+
+
+def draw_errorbars(ax, h, kwargs={}):
```
riga, line 71: Same as above

ap/plotting/plotter.py
```diff
@@ -0,0 +1,181 @@
...
+        cfg = plot_config[key]
+        if "method" not in cfg:
+            raise ValueError("No method given in plot_cfg entry {key}")
+        method = cfg["method"]
+
+        if "hist" not in cfg:
+            raise ValueError("No histogram(s) given in plot_cfg entry {key}")
+        hist = cfg["hist"]
+        kwargs = cfg.get("kwargs", {})
+        globals()[method](ax, hist, kwargs)
```
riga, line 132: Maybe it would be nice to have a mapping somewhere in this file (probably at the end) that explicitly maps method names to methods. Sth along the lines of 

```python
plot_methods = {
    func.__name__: func
    for func in [list of methods here]
}
```

ap/plotting/plotter.py
```diff
@@ -0,0 +1,181 @@
...
+    # legend
+    legend_kwargs = {
+        "title": "Processes",
+        "ncol": 1,
+        "loc": "upper right",
+    }
+    legend_kwargs.update(style_config.get("legend_cfg", {}))
+    ax.legend(**legend_kwargs)
+
+    CMS_label_kwargs = {
```
riga, line 171: We should use all lower-case variable names  :)

riga (changes requested):
Nice :)

I have just a couple smaller comments inline, plus this one.

Could you maybe add type hints to signatures and return values of the plot functions? I think this might help esp. newcomers a lot.

ap/plotting/plotter.py
mafrahm: sure, will do

ap/plotting/plotter.py
mafrahm: Yes, I'll add the changes to the next commit

ap/plotting/plotter.py
mafrahm: ah okay, I misunderstood the comment. I'll implement that.

riga (approved):
LGTM!

PR #71, opened 2022-07-19
Store and apply event weight sums per process.

riga:
This is a rather small PR that fixes how sums of event weights are stored in the selection and applied when normalization weights are created.

Before, sums were stored as a single number per dataset. Now, they are stored per dataset and process id. Also, there a custom selector function now that handles the filing of the `stats` dictionary.

When creating the normalization weights, the sum of event weights is now obtain per event based on its process id, using a lookup table just like for the cross sections.

### DIFF
--- a/ap/production/normalization.py
+++ b/ap/production/normalization.py
@@ -36,10 +36,14 @@ def normalization_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Arra
     lumi = self.config_inst.x.luminosity.nominal
 
     # read the cross section per process from the lookup table
-    xs = np.array(self.xs_table[0, np.asarray(events.process_id)].todense())[0]
+    process_id = np.asarray(events.process_id)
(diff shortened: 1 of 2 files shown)


ap/production/normalization.py
```diff
@@ -69,8 +73,10 @@ def normalization_weights_setup(self: Producer, inputs: dict) -> None:
     attributes:
 
         - py:attr:`selection_stats`: The stats dict loaded from the output of MergeSelectionsStats.
-        - py:attr:`xs_table`: A sparse array serving as a lookup table for all processes known to
-          the config of the task, with keys being process ids.
+        - py:attr:`sum_weights_table`: A sprase array serving as a lookup table for the sum of event
```
mafrahm, line 76: *sparse array

mafrahm (approved):

ap/production/normalization.py
riga: :)

PR #73, opened 2022-07-20
Feature/cutflow variables

mafrahm:
This rather small PR adds a new task `PlotCutflowVariables`, which allows plotting variables using the output from `CreateCutflowHistograms`. The `CreateCutflowHistograms` task is now also a `VariablesMixin` and returns one output file per variable.
The variables which one wants to plot need to be saved in `SelectEvents` via `set_ak_column` in the selector function and added to the `keep_columns` aux in the config.
The `PlotCutflowVariables` adds one new parameter, `plot_steps`, which is a boolean to handle whether to create plots after each individual selection step or only after the final selection step.

Command for testing:
`law run PlotCutflowVariables --version test --variables cf_ht,cf_n_jet --processes tt_sl,st_tchannel, --categories incl --selector-steps Jet,Lepton,Deepjet --plot-steps True --workers 6`

### DIFF
--- a/ap/config/analysis_st.py
+++ b/ap/config/analysis_st.py
@@ -314,7 +314,7 @@ config_2018.set_aux("keep_columns", DotDict.wrap({
         "category_ids", "deterministic_seed",
     },
     "MergeSelectionMasks": {
-        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids", "ht", "n_jet", "jet1_pt",
+        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids", "cutflow.*",
     },
 }))
 
--- a/ap/config/analysis_st.py
+++ b/ap/config/analysis_st.py
@@ -314,7 +314,7 @@ config_2018.set_aux("keep_columns", DotDict.wrap({
         "category_ids", "deterministic_seed",
     },
     "MergeSelectionMasks": {
-        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids",
+        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids", "ht", "n_jet", "jet1_pt",
     },
 }))
 
--- a/ap/config/analysis_st.py
+++ b/ap/config/analysis_st.py
@@ -74,7 +74,7 @@ for dataset_name in dataset_names:
 # default calibrator, selector, producer, ml model and inference model
 config_2018.set_aux("default_calibrator", "test")
 config_2018.set_aux("default_selector", "test")
-config_2018.set_aux("default_producer", "variables")
+config_2018.set_aux("default_producer", "features")
 config_2018.set_aux("default_ml_model", None)
 config_2018.set_aux("default_inference_model", "test")
 
--- a/ap/selection/test.py
+++ b/ap/selection/test.py
@@ -13,6 +13,8 @@ from ap.util import maybe_import
 from ap.columnar_util import set_ak_column, Route
 from ap.production.processes import process_ids
 
+from ap.columnar_util import EMPTY_FLOAT
+
 np = maybe_import("numpy")
 ak = maybe_import("awkward")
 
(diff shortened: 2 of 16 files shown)


ap/selection/test.py
```diff
@@ -92,6 +92,17 @@ def var_HT(self: Selector, events: ak.Array, **kwargs) -> ak.Array:
...
 
+@selector(
+    uses={var_nJet, var_HT, "Jet.pt"},
+    produces={"n_jet", "ht", "jet1_pt"},
+    exposed=True,
+)
+def cutflow_features(self: Selector, events: ak.Array, **kwargs) -> None:
+    set_ak_column(events, "n_jet", self[var_nJet](events))
+    set_ak_column(events, "ht", self[var_HT](events))
+    set_ak_column(events, "jet1_pt", Route("Jet.pt[:,0]").apply(events, -999))
```
riga, line 103: Nice.

btw, could you use the EMPTY_FLOAT from columnar util here instead of the -999?

ap/config/analysis_st.py
```diff
@@ -314,7 +314,7 @@ def make_jme_filenames(jme_aux, sample_type, names, era=None):
         "category_ids", "deterministic_seed",
     },
     "MergeSelectionMasks": {
-        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids",
+        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids", "ht", "n_jet", "jet1_pt",
```
riga, line 317: Only thinking out loud ... but should we maybe add a parent column for variables that are only meant to be produced for the cutflows? E.g. the variables could be named `"cutflow.ht", "cutflow.n_jet", "cutflow.jet1_pt"`, and then we would only need to put "cutflow" into the set of variables to keep.

ap/tasks/cutflow.py
```diff
@@ -146,7 +147,8 @@ def requires(self):
         }
 
     def output(self):
-        return self.local_target("cutflow_hist.pickle")
+        print(self.variables)
```
riga, line 150: stray debug line

riga (changes requested):
Looks good to me. There are just a couple comments inline.

I have one question though. The MergeSelection task now has multiple outputs (one per variable). In the other plottting workflow we store histograms for many variables in a single dict that is saved once. Should we maybe do it the same way here first, until we checked which approach we should follow globally?

Also with the updated MergeSelectionMasks in this PR, all variables are recreated even if a couple of them do already exist. The task is just not re-run when all of them exist.

ap/selection/test.py
mafrahm: good point, done

mafrahm:
> I have one question though. The MergeSelection task now has multiple outputs (one per variable). In the other plottting workflow we store histograms for many variables in a single dict that is saved once. Should we maybe do it the same way here first, until we checked which approach we should follow globally?

I could of course change it back to '1 output per task', but I feel like the '1 output per variable' is the superior way, as long as we only need one variable per task in the following plotting tasks, so I'd like to keep it like that for the moment.

> Also with the updated MergeSelectionMasks in this PR, all variables are recreated even if a couple of them do already exist. The task is just not re-run when all of them exist.

But this is of course not optimal. I propose that I change `CreateCutflowHistograms` such that it only requires `MergeSelectionMasks` for the missing variables

ap/config/analysis_st.py
mafrahm: Good idea, I'll try that out

riga: For compute intensive variables, that are already there, one could also add a simple alias which would not even copy the column :)

riga:
> But this is of course not optimal. I propose that I change `CreateCutflowHistograms` such that it only requires `MergeSelectionMasks` for the missing variables

Yes, I think if we find a good solution for that, we could also solve the "other" plotting workflow (I started in a private branch already, but stopped exactly at that point).

ap/config/analysis_st.py
```diff
@@ -314,7 +314,7 @@ def make_jme_filenames(jme_aux, sample_type, names, era=None):
         "category_ids", "deterministic_seed",
     },
     "MergeSelectionMasks": {
-        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids", "ht", "n_jet", "jet1_pt",
+        "LHEWeight.originalXWGTUP", "normalization_weight", "process_id", "category_ids", "cutflow.*",
```
riga, line 317: You can just add `cutflow` here.

riga (changes requested):

ap/config/analysis_st.py
riga: Sorry, turns out that is not (yet?) implemented.

riga (approved):

PR #74, opened 2022-07-20
Adapt plots to mpl version in venv_columnar.

### LINKED ISSUE #26: Move cmssw env in plotting task to venv

riga:
This PR adapts one part in the plotting code, `ax.stairs`, to be compatible with the mpl version in the current `venv_columnar`. `ax.bars` can do the same, though with a different parametrization of the stat. error bars.

In addition, this PR changes the plot functions from expecting `kwargs` in a single dict, to dynamic `**kwargs`. This way, defaults can be defined as `def func(my_value_with_deefault=123, **kwargs)`, and this avoids the need for having to copy the kwargs to prevent results from in-place operations to show up in calling contexts as well.

Closes #26.

### DIFF
--- a/ap/plotting/plotter.py
+++ b/ap/plotting/plotter.py
@@ -14,7 +14,7 @@ plt = maybe_import("matplotlib.pyplot")
 mplhep = maybe_import("mplhep")
 
 
-def draw_error_stairs(ax: plt.Axes, h: hist.Hist, norm: float = 1.0, **kwargs) -> None:
+def draw_error_bands(ax: plt.Axes, h: hist.Hist, norm: float = 1.0, **kwargs) -> None:
     # compute relative errors
     rel_error = h.variances()**0.5 / h.values()
     rel_error[np.isnan(rel_error)] = 0.0
@@ -120,7 +120,7 @@ def plot_all(plot_config: dict, style_config: dict, ratio: bool = True) -> plt.F
     # available plot methods mapped to their names
     plot_methods = {  # noqa
         func.__name__: func
-        for func in [draw_error_stairs, draw_from_stack, draw_from_hist, draw_errorbars]
+        for func in [draw_error_bands, draw_from_stack, draw_from_hist, draw_errorbars]
     }
 
     plt.style.use(mplhep.style.CMS)
--- a/ap/plotting/plotter.py
+++ b/ap/plotting/plotter.py
@@ -4,7 +4,7 @@
 Generalized plotting functions to create plots from hist histograms
 """
 
-from typing import Optional, Sequence
+from typing import Sequence
 
 from ap.util import maybe_import, test_float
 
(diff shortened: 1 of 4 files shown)


ap/plotting/plotter.py
```diff
@@ -188,7 +182,7 @@ def plot_all(plot_config: dict, style_config: dict, ratio: bool = True) -> plt.F
 
     cms_label_kwargs = {
         "ax": ax,
-        "label": "Work in Progress",
+        "llabel": "Work in progress",
```
mafrahm, line 185: typo

ap/plotting/plotter.py
```diff
@@ -14,45 +14,46 @@
 mplhep = maybe_import("mplhep")
 
 
-def draw_error_stairs(ax: plt.Axes, h: hist.Hist, kwargs: Optional[dict] = None) -> None:
-    if kwargs is None:
-        kwargs = {}
-    norm = kwargs.pop("norm", 1)
-    values = h.values() / norm
-    error = np.sqrt(h.variances()) / norm
+def draw_error_stairs(ax: plt.Axes, h: hist.Hist, norm: float = 1.0, **kwargs) -> None:
```
mafrahm, line 17: might be a good point to choose a better name since we are not using `ax.stairs` anymore. How about `draw_error_bands`?

ap/plotting/plotter.py
```diff
@@ -14,45 +14,46 @@
...
+    # compute relative errors
+    rel_error = h.variances()**0.5 / h.values()
+    rel_error[np.isnan(rel_error)] = 0.0
```
mafrahm, line 20: This function is not only used for the relative errors in the lower panel but also for the Total syst. uncert. in the upper panel. Does that also work?

mafrahm: Ah I see, should in principle work since `height` and `bottom` are multiplied by the `baseline`. However, when running the task myself, I don't see any error bands. Do you have the same issue?

mafrahm: ok, turns out the hatch is just not drawn for some reason. The computation of the errors looks correct.

ap/plotting/plotter.py
```diff
@@ -14,45 +14,46 @@
...
-        "baseline": values - error,
-        "values": values + error,
+        "x": h.axes[0].centers,
+        "width": h.axes[0].edges[1:] - h.axes[0].edges[:-1],
+        "height": baseline * 2 * rel_error,
+        "bottom": baseline * (1 - rel_error),
         "hatch": "///",
         "facecolor": "none",
         "linewidth": 0,
         "color": "black",
```
mafrahm, line 36: The hatch is not drawn for some reason. Changing `"facecolor: "black"` would at least make the error bands visible, but that's not really the style we are looking for.

mafrahm (changes requested):

ap/plotting/plotter.py
riga: I also had to look twice, but the version of mplhep in the venv uses `llabel` (for left label).

ap/plotting/plotter.py
riga: Good point! Changed.

ap/plotting/plotter.py
riga: I checked and the statistical error is just to small (it's mostly in the 0.5 - 1.2% range)

ap/plotting/plotter.py
riga: Yup, see above :)

mafrahm (approved):

PR #76, opened 2022-07-22
Feature/plotting parameters

mafrahm:
This PR adds 5 new parameters to the `PlotMixin`:

- `skip_ratio` removes the lower panel
- `shape_norm` normalized the upper panel on it's integral (not used in PlotCutflow)
- `skip_legend` removes the legend
- `skip_cms` removes the CMS logo but keeps lumi (TODO: does not work due to the mplhep version)
- `yscale` allows setting the y scale to whatever you desire (default: whatever the config of the variable says)

In addition, some defaults for the ylim values are defined based on the chosen scale ("log" or "linear")

Example command for testing:
`law run PlotVariables --version test --processes tt_sl,st_tchannel --variables jet1_pt,ht --workers 6 --skip-ratio True --skip-legend True --skip-cms True --yscale log --shape-norm True`

### DIFF
--- a/ap/config/variables.py
+++ b/ap/config/variables.py
@@ -76,6 +76,7 @@ def add_variables(config: od.Config) -> None:
         binning=(40, 0., 400.),
         unit="GeV",
         x_title=r"Jet 2 $p_{T}$",
+        log_y=True,
     )
     config.add_variable(
         name="jet3_pt",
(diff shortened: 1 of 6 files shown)


riga (changes requested):
Looks good to me !

I had a few inline comments, but to save time I prepared a PR with the suggestions: https://github.com/uhh-cms/analysis_playground/pull/78 (just minor things)

riga (approved):

PR #77, opened 2022-07-22
Improve histogram merging

riga:
This PR adds new features to the `MergeHistograms` task.

- The output histograms are now split for different variables which should help with large file sizes in the future.
- There is a new flag `--only-missing` which only requires `CreateHistograms` and merges files for variables that have not been processed yet.
- There is a new flag `--remove-previous` which leads to the input files (outputs of `CreateHistograms`) being removed after they have been successfully merged.

### DIFF
--- a/ap/tasks/histograms.py
+++ b/ap/tasks/histograms.py
@@ -219,14 +219,16 @@ class MergeHistograms(
         if only_super:
             return reqs
 
-        reqs["hists"] = self.requires_from_branch()
+        reqs["hists"] = self.as_branch().requires()
 
         return reqs
 
     def requires(self):
         # optional dynamic behavior: determine not yet created variables and require only those
+        prefer_cli = {"variables"}
         variables = self.variables
         if self.only_missing:
+            prefer_cli.clear()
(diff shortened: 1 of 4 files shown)


ap/tasks/histograms.py
```diff
@@ -187,62 +188,89 @@ class MergeHistograms(
...
-            for i in range(start_leaf, end_leaf)
+    def run(self):
+        # preare inputs and outputs
+        inputs = self.input()["collection"]
+        outputs = self.output()
+
+        # load input histograms
+        hists = [
+            inp.load(formatter="pickle")
+            for inp in self.iter_progress(inputs.targets.values(), len(inputs), reach=(0, 50))
```
mafrahm, line 259: what is the 'reach' parameter doing?

mafrahm (approved):

ap/tasks/histograms.py
riga: There is a progress bar in the central scheduler gui that is increased from reach[0] to reach[1] (defaulting to 0 and 100) during the iteration. Since there are two iterations (input + output), each progress iteration is just filling half of the progress.

PR #88, opened 2022-08-19
Fix argsort

mafrahm:
This is just a small fix to correctly select jets after sorting. There might be a more elegant way.

Example: `pt = events.Jet.pt = [100, 40, 25, 35]`

Before: the selection would remove the 35 GeV Jet and keep the 25 GeV Jet.
```
mask = pt > 30 = [True, True, False, True]
ak.argsort(pt, ...) = [0, 1, 3, 2]
jet_indices = ak.argsort(pt, ...)[mask] = [0, 1, 2]
pt[jet_indices] = [100, 40, 25]
```

Now:
```
mask = pt > 30 = [True, True, False, True]
jet_indices = ak.argsort(pt, ...) = [0, 1, 3, 2]
masked_jet_indices = jet_indices[mask[jet_indices]] = jet_indices[True, True, True, False] = [0, 1, 3]
pt[masked_jet_indices] = [100, 40, 35]
```


riga (approved):
LGTM!

PR #90, opened 2022-08-22
added process_lines parameter to plotting tasks

mafrahm:
This PR adds the `process_lines` parameter to plotting tasks, which removes selected processes from the stack and plots them as individual lines instead.

Example command:
```
law run cf.PlotVariables \                                                                                                                                       
    --version v1 \
    --datasets st_tchannel_t,tt_sl \
    --producers example \
    --variables jet1_pt \
    --categories 1e \
    --branch 0 \
    --process-lines st_tchannel_t
```

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -786,6 +786,12 @@ class PlotMixin(AnalysisTask):
         description="string parameter to define the y-axis scale of the plot in the upper panel; "
         "choices: NO_STR,linear,log; no default",
     )
+    process_lines = law.CSVParameter(
+        default=(),
+        description="comma-separated process names or patterns to plot as individual lines; "
+        "can also be the key of a mapping defined in the 'process_groups' auxiliary data of "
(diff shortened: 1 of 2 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -786,6 +786,12 @@ class PlotMixin(AnalysisTask):
         description="string parameter to define the y-axis scale of the plot in the upper panel; "
         "choices: NO_STR,linear,log; no default",
     )
+    process_lines = law.CSVParameter(
```
riga, line 789: Just a note:

At some point we should remove the parameters again from the basic plot mixin, but they should rather be attached to a more specific plot tasks. E.g. there's the "skip_ratio" parameter (or now "process_lines") that are good to have for 1D variable plots, but the PlotMixin should also serve as a base for completely different tasks that might not have these features. I will open an issue to document this.

But for now, I think it's ok since these rather specific parameters already exist.

riga (approved):
I only have one inline comment, but it does not require any action for now.

PR #92, opened 2022-08-23
Add mc_weight producer, favor over lhe weight

riga:
This PR adds a new producer `mc_weight` which should be used early on during calibration to create a column with the true MC weight to be used throughout the analysis.

### Background

There are multiple columns that might carry generator weight information:

- `genWeight`
- `Generator.weight`
- `LHEWeight.originalXWGTUP`

According to https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookNanoAOD?rev=99#Weigths the two former columns are identical and this seems true in the 5 different datasets I tested. The next *claim* is that `genWeight` is the product of `LHEWeight.originalXWGTUP` and potential additional weights from the generator chain. However, this is not true for all samples. In some madgraph samples (e.g. `hh_ggf_bbtautau_madgraph`) the lhe weight is some O(1e-2) value, while the `genWeight` is always 1.

To be robust against these bugs in MC, I added a producer that makes a selection between the two options, favoring `genWeight` when it's present and not 1, using the LHE otherwise when existing.

### Changes

- Added the producer
- Adjusted the remaining code to use the `mc_weight` column instead of `LHEWeight.originalXWGTUP`

This should be considered in any downstream analysis.

### DIFF
--- a/columnflow/production/mc_weight.py
+++ b/columnflow/production/mc_weight.py
@@ -1,0 +1,53 @@
+# coding: utf-8
+
+"""
+Methods for dealing with MC weights.
+"""
+
+from columnflow.production import Producer, producer
+from columnflow.util import maybe_import
+from columnflow.columnar_util import set_ak_column, has_ak_column
+
+np = maybe_import("numpy")
+ak = maybe_import("awkward")
+
+
+@producer(
+    uses={"genWeight", "LHEWeight.originalXWGTUP"},
+    produces={"mc_weight"},
+)
+def mc_weight(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
+    """
+    Reads the genWeight and LHEWeight columns and makes a decision about which one to save. This
+    should have been configured centrally [1] and stored in genWeight, but there are some samples
+    where this failed.
+
+    Strategy:
+      1. When LHEWeight is missing, use genWeight.
+      2. When both exist and they are identical, use genWeight.
+      3. When both exist and genWeight is always one, use LHEWeight.
+      4. Raise an exception in all other cases.
+
+    [1] https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookNanoAOD?rev=99#Weigths
+    """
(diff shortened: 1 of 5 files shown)


columnflow/production/mc_weight.py
```diff
@@ -0,0 +1,53 @@
...
+        mc_weight = events.genWeight
+    elif ak.all(events.genWeight == events.LHEWeight.originalXWGTUP):
+        mc_weight = events.genWeight
+    elif ak.all(events.genWeight == 1.0):
+        mc_weight = events.LHEWeight.originalXWGTUP
+    else:
+        raise Exception(
+            f"cannot determine mc_weight from genWeight {events.genWeight} and "
+            f"LHEWeight {events.LHEWeight.originalXWGTUP}",
+        )
```
mafrahm, line 48: the `events.genWeight` and `events.LHEWeight.originalXWGTUP` vector is probably quite large, should we maybe just print a subset, e.g. `events.genWeight[:10]`?

mafrahm: or even specifically select a sub-set of weights where the conditions are not met

mafrahm: Something like that maybe?

```
else:
    mask = (events.genWeight != events.LHEWeight.originalXWGTUP) & (events.genWeight != 1)
    raise Exception(
        f"cannot determine mc_weight from genWeight {events.genWeight[mask]} and "
        f"LHEWeight {events.LHEWeight.originalXWGTUP[mask]}",
    )
```

mafrahm (approved):
Looks good to me, just one small inline comment.

columnflow/production/mc_weight.py
riga: Ah, I was relying on awkward's automatic truncation when printing arrays. Do you think that would be sufficient? :)

mafrahm: yes, this is probably also sufficient.

PR #93, opened 2022-08-23
Make all array operations return views.

riga:
This PR reverts an assumption that we made a couple weeks back.

We assumed that it should be possible to define all ak array operations to work in-place (adding new columns, overwriting columns, removing columns, adding aliases, etc). With that assumption we could simplify a lot of the array function logic in that we could implement caching (without the need to return a value when a cached function is called), handle return values of exposed selectors (which would only return a SelectionResult), and so on.

However, in one of our central helpers, `set_ak_column`, this assumption breaks. When a column should be added that already exists and has some behavior attached to it (which is very often the case for us), the overwriting fails with a cryptic (but surely sensical) awkward exception and the behavior breaks. I even noted this a while back and then had to add the last line shown here:

https://github.com/uhh-cms/columnflow/blob/63071c692b3322631a9f36c3885d6d4a99ed1d93/columnflow/columnar_util.py#L553-L583

Since one would always need to remove potentially existing columns, this removal should be part of `set_ak_column`, which can potentially return a view.

This maybe or maybe-not behavior of return views or original, in-place-changed arrays is not very clean, which is why this PR reverts our initial simplifications, purely relying on and expecting that views are returned by every function in the columnar utils. However, incorporating some of the features / changes added since then, it's actually not too bad now 👀


mafrahm (approved):
The changes look good to me.

PR #95, opened 2022-08-25
Generalize job submission, add SlurmWorkflow

riga:
This PR adds a SlurmWorkflow as well as a common `RemoteWorkflow` that inherits from both `SlurmWorkflow` and `HTCondorWorkflow`. The inheritance of all existing workflows is update so that they can decide whether they use HTCondor or slurm.

In addition, some generalizations were added to make this possible. Overall, the changes mostly affect the job bootstrap script.


mafrahm (approved):
Looks good to me

PR #97, opened 2022-09-01
Allow aux data being passed to SelectionResult.

riga:
This PR contains 3 different types of changes:

1. SelectionResult now allows for keeping auxiliary data that can be passed around between selectors and that is not serialized in `to_ak`.
2. All CSV-based parameter support shell-style brace expansion.
3. Various minor tweaks (mostly wordings in exceptions texts).

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -512,7 +512,7 @@ class CategoriesMixin(ConfigTask):
         default=("1mu",),
         description="comma-separated category names or patterns to select; can also be the key of "
         "a mapping defined in 'category_groups' auxiliary data of the config; default: ('1mu',)",
-        brace_expand=True
+        brace_expand=True,
     )
 
     allow_empty_categories = False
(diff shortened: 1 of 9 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -415,7 +418,7 @@ def events_used_in_training(self, dataset_inst: od.Dataset, shift_inst: od.Shift
         # shall be used in the training
         return (
             dataset_inst in self.ml_model_inst.used_datasets and
-            not shift_inst.x("disjoint_from_nominal", False)
+            not shift_inst.has_tag("disjoint_from_nominal")
```
mafrahm, line 421: This changed from `not disjoint_from_nominal = False` to `not disjoint_from_nominal` so the logic changed. Is that correct?

mafrahm (approved):
Looks good, just a single inline comment

columnflow/tasks/framework/mixins.py
riga: That `False` was just the default when the aux entry did not exist, but the logic still was "if not disjoint from nominal".

PR #100, opened 2022-09-15
Implement MET propagation.

riga:
This PR adds the propagation of shifted jet four vectors due to JEC and JER to the MET "vector".

There is no selection yet on the jets whose updates shifts are propagated.
@dsavoiu Do you think that's ok?

(reopened from #99 which fixed the CI instead)

### DIFF
--- a/columnflow/calibration/jets.py
+++ b/columnflow/calibration/jets.py
@@ -5,11 +5,13 @@ Calibration methods for jets
 """
 
 import os
+from typing import Tuple
 
 import law
 
 from columnflow.util import maybe_import, memoize
 from columnflow.calibration import Calibrator, calibrator
+from columnflow.production.util import attach_coffea_behavior
 from columnflow.columnar_util import set_ak_column
 
 np = maybe_import("numpy")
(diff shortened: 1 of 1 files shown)


columnflow/calibration/jets.py
```diff
@@ -173,11 +210,29 @@ def jec(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
         Rho=events.fixedGridRhoFastjetAll,
     )
 
+    # store pt and phi of the full jet system
+    jetsum = ak.sum(events.Jet, axis=1)
```
dsavoiu, line 214: AFAIK there is a pT threshold on the jets that enter the Type-I MET correction (maybe 15 GeV?), otherwise one could introduce a bias on the MET. Any plans to implement this? (Should be enough to just mask/filter the jets in the sum here based on some configurable threshold.)

dsavoiu (commented):
This looks good! The code is clear and the `prop_met` function is very useful.

What indeed is still missing is the selection of the jets entering the MET correction, in particular the application of a pT threshold. Of course the jets in nanoAOD already have a threshold set, but for MET I vaguely recall that an even higher threshold was used (15 GeV?), so it would be good to add the option to do that.

Another issue I see is that the full set of JECs is propagated to the MET, including the L1 correction, whereas IIRC the recommendation is to only propagate the JECs starting with L2Relative (See for example the TWiki on [METType1Type2Formulae](https://twiki.cern.ch/twiki/bin/viewauth/CMS/METType1Type2Formulae) - the page is old and the numbers are outdated, but I think the formulas still apply.)

This might not be relevant for Run 3 with PUPPI jets (which don't have an L1 correction IIRC), but should be implemented anyway to support Run 2 analyses.

columnflow/calibration/jets.py
```diff
@@ -393,10 +469,38 @@ def jer(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
...
+    # (note: apply variations first since they refer to the original pt)
+    events = set_ak_column(events, "Jet.pt_jer_up", events.Jet.pt * smear_factors[:, :, 1])
+    events = set_ak_column(events, "Jet.mass_jer_up", events.Jet.mass * smear_factors[:, :, 1])
+    events = set_ak_column(events, "Jet.pt_jer_down", events.Jet.pt * smear_factors[:, :, 2])
+    events = set_ak_column(events, "Jet.mass_jer_down", events.Jet.mass * smear_factors[:, :, 2])
+    events = set_ak_column(events, "Jet.pt", events.Jet.pt * smear_factors[:, :, 0])
+    events = set_ak_column(events, "Jet.mass", events.Jet.mass * smear_factors[:, :, 0])
+
+    # recover coffea behavior
+    events = self[attach_coffea_behavior](events, colletions=["Jet"], **kwargs)
```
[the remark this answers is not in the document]
riga: Fixed.

columnflow/calibration/jets.py
```diff
@@ -173,11 +210,29 @@ def jec(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
...
+    jet_pt1 = jetsum.pt
+    jet_phi1 = jetsum.phi
+
     # store corrected pt, mass and recalculate rawFactor
     events = set_ak_column(events, "Jet.pt", events.Jet.pt_raw * jec_factors)
     events = set_ak_column(events, "Jet.mass", events.Jet.mass_raw * jec_factors)
     events = set_ak_column(events, "Jet.rawFactor", (1 - events.Jet.pt_raw / events.Jet.pt))
 
+    # recover coffea behavior
+    events = self[attach_coffea_behavior](events, colletions=["Jet"], **kwargs)
```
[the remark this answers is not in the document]
riga: Fixed.

columnflow/calibration/jets.py
riga: Yep, I'm going to add the two cuts mentioned in the [https://twiki.cern.ch/twiki/bin/view/CMS/JECAnalysesRecommendations?rev=19#Minimum_jet_selection_cuts](JECAnalysesRecommendations).

riga: Added :)

dsavoiu (approved):
Thanks for the update @riga, everything looks OK from what I can tell so this can probably be merged now.

PR #101, opened 2022-09-19
Restructuring of Plotting base tasks and adding process_settings parameter to plotting

mafrahm:
This PR moves most plotting parameters from the `PlotMixin` into different base plotting tasks. The structure is (atm) setup as follows:
`PlotMixin` -> `PlotBase` ->`PlotBase1D` ->`ProcessPlotBase` and
`PlotMxin` -> `PlotBase` -> `PlotBase2D` (dummy since not yet used in any tasks)

This PR also allows setting the style of individual processes in plots on the command line (or as default in the config)
Possible parameters that are supported at the moment are: `label`, `color` (matplotlib default colors or hex code), `unstack` and `scale`, but can be implemented in the plotting function independently of the task.
An exemplary command would look like
```
law run cf.PlotVariables --version v1 --process-settings tt,label="tt custom label",color=red:st,unstack,scale=10
```
Or alternatively
`law run cf.PlotVariables --version v1 --process-settings default` with
```
process_settings_groups = {
    "default" : [["tt", "label=tt custom label", "color=red"], ["st", "unstack", "scale=10"]],
}
```

Other changes:
- you can now change the xticklabels of the Cutflow plots via setting up a `selector_step_labels` dictionary in the config (mapping the name of a selector step to a label).
- The `plot_suffix` parameter was added to allow customizing of plot output names. (For completeness I could also add a `plot_prefix` parameter)
- some minor changes in plotting functions

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -781,52 +781,6 @@ class PlotMixin(AnalysisTask):
         description="a command to execute after the task has run to visualize plots right in the "
         "terminal; no default",
     )
-    skip_ratio = luigi.BoolParameter(
-        default=False,
-        significant=False,
-        description="when True, no ratio (usually Data/Bkg ratio) is drawn in the lower panel; "
-        "default: False",
-    )
-    shape_norm = luigi.BoolParameter(
-        default=False,
-        significant=False,
-        description="when True, each process is normalized on it's integral in the upper panel; "
-        "default: False",
-    )
-    skip_legend = luigi.BoolParameter(
-        default=False,
-        significant=False,
-        description="when True, no legend is drawn; default: False",
-    )
-    skip_cms = luigi.BoolParameter(
-        default=False,
-        significant=False,
-        description="when True, no CMS logo is drawn; default: False",
-    )
-    y_scale = luigi.ChoiceParameter(
-        choices=(law.NO_STR, "linear", "log"),
-        default=law.NO_STR,
-        significant=False,
-        description="string parameter to define the y-axis scale of the plot in the upper panel; "
-        "choices: NO_STR,linear,log; no default",
-    )
-    process_lines = law.CSVParameter(
-        default=(),
-        description="comma-separated process names or patterns to plot as individual lines; "
-        "can also be the key of a mapping defined in the 'process_groups' auxiliary data of "
-        "the config; uses no process when empty; empty default",
-    )
-
-    def get_plot_parameters(self):
-        # convert parameters to usable values during plotting
-        return {
-            "skip_ratio": self.skip_ratio,
-            "shape_norm": self.shape_norm,
-            "skip_legend": self.skip_legend,
-            "skip_cms": self.skip_cms,
-            "yscale": None if self.y_scale == law.NO_STR else self.y_scale,
-            "process_lines": self.process_lines,
-        }
 
     def get_plot_func(self, func_name: str) -> Callable:
         """
--- a/columnflow/tasks/framework/plotting.py
+++ b/columnflow/tasks/framework/plotting.py
@@ -1,0 +1,132 @@
+# coding: utf-8
+
+"""
+Base tasks for different types of plotting tasks
+"""
+
+import law
+import luigi
+
+from columnflow.tasks.framework.mixins import PlotMixin, CategoriesMixin, DatasetsProcessesMixin
+
+
+class PlotBase(PlotMixin):
+    """
+    Base class for all plotting tasks.
+    """
+
+    skip_legend = luigi.BoolParameter(
+        default=False,
+        significant=False,
+        description="when True, no legend is drawn; default: False",
+    )
+    skip_cms = luigi.BoolParameter(
+        default=False,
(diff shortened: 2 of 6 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -781,52 +781,6 @@ class PlotMixin(AnalysisTask):
...
-    def get_plot_parameters(self):
-        # convert parameters to usable values during plotting
-        return {
-            "skip_ratio": self.skip_ratio,
-            "shape_norm": self.shape_norm,
-            "skip_legend": self.skip_legend,
-            "skip_cms": self.skip_cms,
-            "yscale": None if self.y_scale == law.NO_STR else self.y_scale,
-            "process_lines": self.process_lines,
-        }
```
riga, line 829: Maybe we define a base method here to express that this feature exists? Also, in case we have generic parameters that can be defined in this main mixin, we would be able to inject those here. I.e.

```python
def get_plot_parameters(self):
    return {}
```

And then in `PlotBase` (inheriting from this mixin) and below, we start with `parameters = super().get_plot_parameters()` and update `parameters` (edit: like you already do actually :) but we could start at this mixin)

columnflow/tasks/plotting.py
```diff
@@ -191,6 +178,27 @@ class PlotShiftedVariables(
...
+        significant=True,
+        description="when True, one plot per process is produced; default: False",
+    )
+
+    legtitle = luigi.Parameter(
+        default=law.NO_STR,
+        significant=False,
+        description="sets the title of the legend; when empty and only one process is present in "
+        "the plot, the process_inst label is used; empty default",
+    )
```
riga, line 192: For consistency we should maybe name this `legend_title`? It could help to make this a little more verbose on the command line.

columnflow/tasks/framework/mixins.py
riga: Thinking about this again ... do we still need the mixin when all plots anyway inherit from the new `PlotBase`, i.e., would there at some point be any task that inherits from the mixin but not the base tasks?

columnflow/tasks/framework/plotting.py
```diff
@@ -0,0 +1,166 @@
...
+    Base class for plotting tasks creating 1-dimensional plots.
+    """
+
+    skip_ratio = luigi.BoolParameter(
+        default=False,
+        significant=False,
+        description="when True, no ratio (usually Data/Bkg ratio) is drawn in the lower panel; "
+        "default: False",
+    )
+    yscale = luigi.ChoiceParameter(
```
riga, line 54: For consistency, I would still vote for `y_scale`. Below you used `z_scale` already :)

columnflow/tasks/plotting.py
```diff
@@ -226,7 +234,17 @@ def requires(self):
 
     def output(self):
         b = self.branch_data
-        return self.target(f"plot__cat_{b.category}__var_{b.variable}.pdf")
+        suffix = f"__{self.plot_suffix}" if self.plot_suffix else ""
+        if self.per_process:
+            # one output per process
+            return law.SiblingFileCollection({
+                p: self.target(f"plot__proc_{p}__unc_{b.shift_source}__cat_"
+                               f"{b.category}__var_{b.variable}{suffix}.pdf")
```
riga, line 242:
```suggestion
                p: self.target(
                    f"plot__proc_{p}__unc_{b.shift_source}__cat_{b.category}"
                    f"__var_{b.variable}{suffix}.pdf"
                )
```

riga (changes requested):
Looks very good! A few minor things below ...

columnflow/tasks/plotting.py
mafrahm: Sure. I think this parameter could also be part of `PlotBase`, right? I'd just need to remove the plotting function dependent part of the description.

columnflow/tasks/framework/plotting.py
mafrahm: We could use y_scale as parameter for the command line and directly switch to `yscale` here:
https://github.com/uhh-cms/columnflow/blob/feature/PlotBase/columnflow/tasks/framework/plotting.py#L73
But I feel like this is much more confusing than just staying with yscale consistently.

columnflow/tasks/plotting.py
riga: Yeah, I think this makes sense in PlotBase.

columnflow/tasks/framework/mixins.py
mafrahm: You're probably right. And even if the `PlotBase` is not the true base task for all plots, we simply have to modify it or add another PlotBase task somewhere.

columnflow/tasks/framework/plotting.py
riga: I think it should be either `yscale` and `zscale`, or `y_scale` and `z_scale`, and I'd have a slight tendency towards the latter with other parameters in mind that might be added in the future, such as "x_range", "x_min", "x_max", ....

riga (approved):
LGTM!

PR #102, opened 2022-09-21
Fix jec/jer after ReduceEvents

mafrahm:
Changes:

- added `selection_dependent` tag (feel free to suggest a better name) for shifts such that in tasks after ReduceEvents it is not attempted to use aliases for columns such as Jet.pt
- added `jecdummy` shifts to the example config to enable testing such uncertainties without setting up an analysis.
- changed default category from `1mu` to `incl`
- minor changes/fixes to example selector


riga (approved):
Looks good to me!

PR #103, opened 2022-09-23
Support for multiple file types in plots.

riga:
This small PRs adds an additional parameter to plot tasks that let's users select the file types of plots. Example:

`law run cf.PlotVariables ... --file-types png,pdf --plot-suffix foo`

will create two output targets, png and pdf. Internally, there is a function now (`get_plot_names`) that loops over file types and also adds the plot postfix at a central spot.


mafrahm (approved):
Looks good, no comments from my side.

PR #109, opened 2022-10-06
Update code style and venv requirements for Python 3.9.

riga:
With the new conda-based setup in #107, we can update large parts of our code base to use some newer python features as well as update some of the requirements in the base venvs.

This PR adds exactly these changes but should only be merged when #107 is included.
Until then, there are some minor TODOs left:

- [x] Change sandboxes of plotting tasks to the default columnar env.
- [x] Merge master back in.

### DIFF
--- a/columnflow/calibration/__init__.py
+++ b/columnflow/calibration/__init__.py
@@ -4,7 +4,9 @@
 Object and event calibration tools.
 """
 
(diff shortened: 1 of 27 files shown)


riga:
As coordinated offline earlier, since #107 is in the master and *this* PR only contains style and format changes, I'm going ahead and merge.

PR #112, opened 2022-10-17
Add default variables and categories to config.

### LINKED ISSUE #91: Move residual analysis specific code
There are some left-over parts that seem too analysis specific and contain choices that maybe should not be part of columnflow. This issue is meant to document them for later action.

- [x] As mentioned in https://github.com/uhh-cms/columnflow/pull/90#discussion_r951242368, the PlotMixin contains some parameters that are somewhat too specific for the main base plot task. They should be moved to a less generic base task for (e.g.) 1D plots. However, even this choice seems somewhat too exclusive as plotting is always highly custom, so that such more specific plot base tasks should be rather analysis specific.
- [ ] There's a "1mu" category in the mixins.

@mafrahm

riga:
This PR removes the last analysis specific pieces and closes #91.

There was a residual specification of a default category being defined in the mixins. With this PR, the `CategoryMixin` and `VariablesMixin` don't define defaults on their own, but read values from the analysis specific config.

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -550,7 +550,7 @@ class CategoriesMixin(ConfigTask):
             if not params["categories"] and config_inst.x("default_categories", ()):
                 params["categories"] = tuple(config_inst.x.default_categories)
 
-            # when still empty and default variables are defined, use them instead
+            # when still empty and default categories are defined, use them instead
             if not params["categories"] and cls.default_categories:
                 params["categories"] = tuple(cls.default_categories)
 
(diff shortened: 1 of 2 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -544,6 +546,15 @@ def modify_param_values(cls, params):
 
         # resolve categories
         if "categories" in params:
+            # when empty, use the config default
+            if not params["categories"] and config_inst.x("default_categories", ()):
+                params["categories"] = tuple(config_inst.x.default_categories)
+
+            # when still empty and default variables are defined, use them instead
```
mafrahm, line 553: default categories*

mafrahm (approved):

columnflow/tasks/framework/mixins.py
riga: Fixed 👍

PR #114, opened 2022-11-17
Feature/plot2dplotter

mafrahm:
This PR restructures the plotting tasks as shown in the comment below.
The major changes are that the `PlotVariables` task is now separated in two callable tasks,  `PlotVariables1d` and `PlotVariables2d`, which have the same base class but inherit from different base classes to provide different sets of parameters.

The new plotting function is implemented here [1] and uses all plotting parameters enabled for 2d plotting tasks except the `process-settings`.

Example command for `PlotVariables2d`:
```
law run cf.PlotVariables2d --version v1 --processes tt_sl,st_tchannel --variables n_jet_vs_jet1_pt --zscale log --shape-norm True
```
`PlotCutflowVariables2d` is also implemented and working, producing one plot per process and selector-step.

TODOs (for a next PR?):
- we should implement a `per-process` parameter for the 2d tasks to enable switching between producing 1 plot per process or 1 plot with all processes added together. This could also be a part of the `ProcessPlotSettingMixin`.
- use the process_settings in `plot_2d`. Changing plotting label and process scale could be used.
- The plot_function_names in `PlotCutflowVariables1d` are still hard-coded.

[1]

### DIFF
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -12,9 +12,9 @@ import law
 
 from columnflow.tasks.framework.base import AnalysisTask, DatasetTask, ShiftTask, wrapper_factory
 from columnflow.tasks.framework.mixins import (
-    CalibratorsMixin, SelectorStepsMixin, VariablesMixin, ChunkedReaderMixin,
+    CalibratorsMixin, SelectorStepsMixin, VariablesMixin, CategoriesMixin, ChunkedReaderMixin,
 )
-from columnflow.tasks.framework.plotting import PlotBase, ProcessPlotBase
+from columnflow.tasks.framework.plotting import PlotBase, ProcessPlotSettingMixin
 from columnflow.tasks.framework.remote import RemoteWorkflow
 from columnflow.tasks.selection import MergeSelectionMasks
 from columnflow.util import dev_sandbox, DotDict
@@ -174,7 +174,8 @@ class PlotCutflow(
     ShiftTask,
     SelectorStepsMixin,
     CalibratorsMixin,
-    ProcessPlotBase,
+    CategoriesMixin,
+    ProcessPlotSettingMixin,
     law.LocalWorkflow,
     RemoteWorkflow,
 ):
@@ -338,7 +339,7 @@ class PlotCutflowVariables(
     VariablesMixin,
     SelectorStepsMixin,
     CalibratorsMixin,
-    ProcessPlotBase,
+    ProcessPlotSettingMixin,
     law.LocalWorkflow,
     RemoteWorkflow,
 ):
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -349,6 +349,7 @@ class PlotCutflowVariablesBase(
     law.LocalWorkflow,
     RemoteWorkflow,
 ):
+    exclude_index = True
 
     only_final_step = luigi.BoolParameter(
         default=False,
@@ -607,6 +608,8 @@ class PlotCutflowVariables2d(
         outputs = self.output()
 
         for step in self.chosen_steps:
+            # TODO: implement a 'per_process' parameter similar to PlotShiftedVariables
+            #       (maybe as parameter in ProcessPlotSettingMixin, since we could also use this in PlotVariables)
             for process_inst, h in hists.items():
                 h_step = {process_inst: h[{"step": hist.loc(step)}]}
 
@@ -616,6 +619,7 @@ class PlotCutflowVariables2d(
                     hists=h_step,
                     config_inst=self.config_inst,
                     variable_insts=variable_insts,
+                    style_config={"legend_cfg": {"title": f"Step '{step}'"}},
                     **self.get_plot_parameters(),
                 )
 
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -185,7 +185,7 @@ class PlotCutflow(
 
     shifts = set(CreateCutflowHistograms.shifts)
 
-    plot_function_name_1d = luigi.Parameter(
+    plot_function_1d = luigi.Parameter(
         default="columnflow.plotting.example.plot_cutflow",
         significant=False,
         description="name of the 1d plot function; default: 'columnflow.plotting.example.plot_cutflow'",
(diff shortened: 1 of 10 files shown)


columnflow/tasks/cutflow.py
```diff
@@ -190,7 +200,7 @@ def get_plot_parameters(self):
         params = super().get_plot_parameters()
 
         # no ratio plot implemented: disable ratio plot for `plot_cutflow`
-        if self.plot_function_name == "columnflow.plotting.example.plot_cutflow":
+        if self.plot_function_name_1d == "columnflow.plotting.example.plot_cutflow":
```
riga, line 203: Maybe `skip_ratio` should rather become a parameter? Having plot function dependent settings somehow works around the function plug-in mechanism and might not scale too well.

columnflow/util.py
```diff
@@ -530,6 +530,16 @@ def __getattr__(self, attr: str) -> Any:
...
+    def cautious_update(self, other: dict) -> None:
+        """
+        Updates DotDict, but only if no value is overwritten. Throws error otherwise.
+        """
+        key_replicas = set(self.keys()) & set(other.keys())
+        if not key_replicas:
+            self.update(other)
+        else:
+            raise KeyError(f"'{self.__class__.__name__}' object already has attributes {key_replicas}")
+
```
riga, line 542: 2 things here:

- For the DotDict I think it would be better to keep the amount of "real" attributes as minimal as possible and thus rather move such a functionality into a util function that accepts any dict subtypes.
- Maybe more importantly: I think this way of doing "cautious" updates can be dangerous. IIUC, if any of the keys is already existing in the dict, then none of the other dict items are added. But then, one always needs to know whether certain keys are definitely not present, which is quite implicit and might depend on the situation. As an alternative, I would explicitly check for the presence of a certain key (at the point where the update should be performed), and then add new items depending on the response (see e.g. above)

columnflow/tasks/framework/plotting.py
```diff
@@ -178,21 +183,22 @@ class PlotBase1D(PlotBase):
...
-        params["skip_ratio"] = self.skip_ratio
-        params["yscale"] = None if self.yscale == law.NO_STR else self.yscale
-        params["shape_norm"] = self.shape_norm
-
+        params_to_add = {
+            "skip_ratio": self.skip_ratio,
+            "yscale": None if self.yscale == law.NO_STR else self.yscale,
+            "shape_norm": self.shape_norm,
+        }
+        params.cautious_update(params_to_add)
```
riga, line 195: See the last review comment first :)

The `cautious_update` can be dangerous here in case (e.g.) `yscale` is, say, already present and set to `None`. This would be fully consistent (since `None` would also be added through `params_to_add`), but still, `skip_ratio` and `shape_norm` would not be added to `params`.

Instead, maybe this is what one would like to read here as it is more explicit:

```python
... removed ...
```

... actually, I don't see a dependence among the three `params_to_add`, so I'm not sure why one would want to add all of them, but only if none of them is already present 🤔

riga (changes requested):
Nice work 🎉 I'm eager to see the 2D plot style.

However, I couple comments inline, mainly related to the `cautious_update` which I find too implicit and therefore dangerous.

columnflow/tasks/cutflow.py
riga, line 203: Maybe `skip_ratio` should rather become a parameter? Having plot function dependent settings somehow works around the function plug-in mechanism and might not scale too well.

riga (changes requested):

columnflow/tasks/cutflow.py
mafrahm: `skip_ratio` already is a parameter, I just set it to True here, because this plot function does not provide a ratio plot. But I think, the ratio plot would not be produced event when `skip_ratio=False`, so I could also just remove this line.

riga: 👍

mafrahm:
There were still some issues with handling of line colors of stacks. I fixed this with another commit, where I also moved some duplicated code into this file:
https://github.com/uhh-cms/columnflow/blob/feature/plot2dplotter/columnflow/plotting/plot_util.py

columnflow/tasks/cutflow.py
mafrahm: turns out I can't just remove this because then an empty ratio plot is drawn, but I simply set `skip_ratio=True` in the plot_function itself

columnflow/util.py
dsavoiu: I think the idea here was to have a convenience function that works as a regular `dict.update`, but have it fail if any of the keys to be added is already present in the dictionary. The main benefit of this is to prevent accidentally overwriting a value that was set by a previous function operating on the same `DotDict` (e.g. to prevent style changes from getting lost/clobbered when multiple plot functions are called successively). I'm not sure I would consider it dangerous, it would just fail instead of overwriting, so in a sense it functions as an assertion.

This could also be a util function instead of a member of `DotDict`, or one could avoid a function completely and check the presence of a key explicitly whenever a key is added, but having a function would avoid make the calling code less verbose.

It's true that the **values** are not checked. One could of course extend the implementation to fail not just if a key is already present, but only if the value for that key would also change through the update.

columnflow/tasks/framework/plotting.py
dsavoiu: To handle the case where (part of) a `cautious_update` is consistent with the existing values one could explicitly check if these change (see above).

dsavoiu (approved):
Very nice, this is definitely useful! I had a few comments about the `cautious_update` function, but it seems like `dict_add_strict` fits in quite nicely as an alternative. Looks good to me otherwise, so I'm fine with merging these changes.

columnflow/util.py
mafrahm: I added the value check to the new `dict_add_strict` function :+1:

riga (changes requested):
Just some minor left-over comment.

columnflow/tasks/framework/plotting.py
riga: Indeed. With the strict add method, this should be fine.

columnflow/tasks/framework/plotting.py
```diff
@@ -257,23 +259,33 @@ def store_parts(self):
         return parts
 
 
-class PlotBase2D(PlotBase):
+class PlotBase2d(PlotBase):
     """
     Base class for plotting tasks creating 2-dimensional plots.
     """
 
+    plot_function_name_2d = luigi.Parameter(
```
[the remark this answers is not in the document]
mafrahm: I agree, this change is pushed

riga (approved):

PR #115, opened 2022-11-23
Add btag and muon weight producers.

riga:
This PR adds two new producers for btag SF and muon weights. Both producers are based on correctionlib values. In particular:

- The `btag_weight` producer implements the shape calibrating SFs from the iterative fit method. However, no renormalization is done on this end which is probably too analysis specific. The producer is just meant to produce the bare per-event weight as a product of per-jet SFs, including systematic variations `["hf", "lf", "hfstats1", "hfstats2", "lfstats1", "lfstats2"]`.
- The `muon_weight` prodicer (just like the already existing `electron_weight`) producer covers all effects except for trigger efficiency discrepancies. The application is straight forward.

Both producers require certain settings in the central `config_inst`, but this is better documented in the doc strings of the producer functions.

### DIFF
--- a/columnflow/production/btag.py
+++ b/columnflow/production/btag.py
@@ -20,7 +20,7 @@ ak = maybe_import("awkward")
     },
     # produced columns are defined in the init function below
 )
-def btag_weight(
+def btag_weights(
     self: Producer,
     events: ak.Array,
     jet_mask: ak.Array | type(Ellipsis) = Ellipsis,
@@ -50,9 +50,6 @@ def btag_weight(
     if self.dataset_inst.is_data:
         return events
 
-    # get the total number of jets in the chunk
-    n_jets_all = len(flat_np_view(events.Jet.pt, axis=1))
-
     # get flat inputs, evaluated at jet_mask
     flavor = flat_np_view(events.Jet.hadronFlavour[jet_mask], axis=1)
     abs_eta = flat_np_view(abs(events.Jet.eta[jet_mask]), axis=1)
@@ -78,7 +75,7 @@ def btag_weight(
         )
 
         # insert them into an array of ones whose length corresponds to the total number of jets
-        sf_flat_all = np.ones(n_jets_all, dtype=np.float32)
+        sf_flat_all = np.ones_like(flavor, dtype=np.float32)
         if jet_mask is Ellipsis:
             indices = flavor_mask
         else:
@@ -121,8 +118,8 @@ def btag_weight(
     return events
 
 
-@btag_weight.init
-def btag_weight_init(self: Producer) -> None:
+@btag_weights.init
+def btag_weights_init(self: Producer) -> None:
     # depending on the requested shift_inst, there are three cases to handle:
     #   1. when a JEC uncertainty is requested whose propagation to btag weights is known, the
     #      producer should only produce that specific weight column
@@ -172,8 +169,8 @@ def btag_weight_init(self: Producer) -> None:
         self.produces.add("btag_weight")
 
 
-@btag_weight.requires
-def btag_weight_requires(self: Producer, reqs: dict) -> None:
+@btag_weights.requires
+def btag_weights_requires(self: Producer, reqs: dict) -> None:
     if self.dataset_inst.is_data or "external_files" in reqs:
         return
 
@@ -181,8 +178,8 @@ def btag_weight_requires(self: Producer, reqs: dict) -> None:
     reqs["external_files"] = BundleExternalFiles.req(self.task)
 
 
-@btag_weight.setup
-def btag_weight_setup(self: Producer, reqs: dict, inputs: dict) -> None:
+@btag_weights.setup
+def btag_weights_setup(self: Producer, reqs: dict, inputs: dict) -> None:
     if self.dataset_inst.is_data:
         self.btag_sf_corrector = None
         return
--- a/columnflow/production/btag.py
+++ b/columnflow/production/btag.py
@@ -1,0 +1,197 @@
+# coding: utf-8
+
+"""
+Producers for btag scale factor weights.
+"""
+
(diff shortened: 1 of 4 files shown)


riga:
Adding @nprouvost and @Bogdan-Wiederspan who also worked on the actual implementations.

columnflow/production/muon.py
```diff
@@ -0,0 +1,96 @@
...
+
+    .. code-block:: python
+
+        cfg.x.muon_sf_names = ("NUM_TightRelIso_DEN_TightIDandIPCut", "2017_UL")
+    """
+    if self.dataset_inst.is_data:
+        return events
+
+    # get year string
+    year = self.config_inst.x.muon_sf_names[1]
```
mafrahm, line 46: maybe change the name to something else (e.g. `trig_year`) to make clear that this is (potentially) different to the campaign year?

columnflow/production/muon.py
```diff
@@ -0,0 +1,96 @@
...
+    abs_eta = ak.flatten(abs(events.Muon.eta), axis=1)
+    pt = ak.flatten(events.Muon.pt, axis=1)
+
+    # loop over systematics
+    for syst, postfix in [
+        ("sf", ""),
+        ("systup", "_up"),
+        ("systdown", "_down"),
+    ]:
+        sf_flat = self.muon_sf_corrector.evaluate(year, abs_eta, pt, syst).astype(np.float32)
```
mafrahm, line 58: I assume this operation can only be performed with flattened arrays?

columnflow/production/btag.py
```diff
@@ -0,0 +1,197 @@
...
+        sf_flat = self.btag_sf_corrector.evaluate(
+            syst_name if syst_name == "central" else f"{syst_direction}_{syst_name}",
+            flavor[flavor_mask],
+            abs_eta[flavor_mask],
+            pt[flavor_mask],
+            b_discr[flavor_mask],
+        )
+
+        # insert them into an array of ones whose length corresponds to the total number of jets
+        sf_flat_all = np.ones(n_jets_all, dtype=np.float32)
```
mafrahm, line 81: `sf_flat_all = np.ones_like(flavor, dtype=np.float32)` ? Then we don't need `n_jets_all` at all

columnflow/production/btag.py
```diff
@@ -0,0 +1,197 @@
...
+        if jet_mask is Ellipsis:
+            indices = flavor_mask
+        else:
+            indices = flat_np_view(jet_mask)
+            if flavor_mask is not Ellipsis:
+                indices = np.where(indices)[0][flavor_mask]
```
mafrahm, line 87: why not just `indices = indices & flavor_mask` ?

columnflow/production/btag.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # save the new column
+        return set_ak_column(events, column_name, ak.values_astype(weight, np.float32))
+
+    # when the uncertainty is a known jec shift, obtain the propagated effect and do not produce
+    # additional systematics
+    if self.shift_inst.is_nominal:
+        # nominal weight and those of all method intrinsic uncertainties
+        events = add_weight("central", None, "btag_weight")
+        for syst_name, col_name in self.btag_uncs.items():
```
mafrahm, line 102: Are the `btag_uncs` always the same or should we allow setting them in some config?

mafrahm (approved):
Looks good :+1: I just left some minor comments and questions, but nothing game-changing.

columnflow/production/btag.py
```diff
@@ -0,0 +1,197 @@
...
+    Resources:
+
+       - https://twiki.cern.ch/twiki/bin/view/CMS/BTagShapeCalibration?rev=26
+       - https://indico.cern.ch/event/1096988/contributions/4615134/attachments/2346047/4000529/Nov21_btaggingSFjsons.pdf
+    """
+    if self.dataset_inst.is_data:
+        return events
+
+    # get the total number of jets in the chunk
+    n_jets_all = len(flat_np_view(events.Jet.pt, axis=1))
```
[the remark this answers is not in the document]
riga: True. Added that 👍

columnflow/production/muon.py
riga: Exactly. As far as I can see, all correctionlib correctors support only flat inputs and outputs.

columnflow/production/btag.py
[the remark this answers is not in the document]
riga: ... and removed it.

columnflow/production/btag.py
riga: Ha, also true. Applied your suggestion.

columnflow/production/btag.py
riga: Then the resulting `indices` would have the same shape as before, just with fewer `True`'s, whereas above, there is an actual reduction happening to match the shape of the arrays it is applied to later on.

columnflow/production/muon.py
riga: 👍 I changed it to `sf_year`.

columnflow/production/btag.py
riga: For the shape calibrating SFs, the uncertainties have been the same since roughly 5 years, and they should not change in the future, so I thought it's ok for have them in there. Also, there are some hard-coded conditions necessary anyway, e.g.

https://github.com/uhh-cms/columnflow/blob/4dadfe46ebf9786fdc6efcf7370b36bb43c05fe3/columnflow/production/btag.py#L66-L69

columnflow/production/btag.py
mafrahm: OK, then I probably overlooked something. I'll maybe take another look into these lines of code after merging, but for now we can resolve this comment :+1:

PR #116, opened 2022-11-24
Add scale and pdf weights producer

mafrahm:
This PR adds producers for mur/muf and pdf variations

### DIFF
--- a/columnflow/production/pdf.py
+++ b/columnflow/production/pdf.py
@@ -36,7 +36,7 @@ def pdf_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
     if ak.any(N_pdfweights != 103) or ak.any(N_pdfweights != 101):
         raise Exception(
             f"Number of LHEPdfWeights ({N_pdfweights}) is not as expected (103 or 101) "
-            f"in dataset {self.dataset_inst.name}"
+            f"in dataset {self.dataset_inst.name}",
         )
 
     # first LHEPdfWeight value: nominal weight
--- a/columnflow/production/pdf.py
+++ b/columnflow/production/pdf.py
@@ -1,0 +1,53 @@
+# coding: utf-8
+
+"""
+Column production methods related to the PDF weights.
+"""
+
+
+from columnflow.production import Producer, producer
+from columnflow.util import maybe_import
+from columnflow.columnar_util import set_ak_column
+
+ak = maybe_import("awkward")
+
+
+@producer(
+    uses={"LHEPdfWeight"},
+    produces={
+        "pdf_weight", "pdf_weight_up", "pdf_weight_down",
+    },
+)
+def pdf_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
+    """
+    determines the pdf up and down shifts
+    Documentation: https://arxiv.org/pdf/1510.03865.pdf
+    """
+
+    # stop here for data
+    if self.dataset_inst.is_data:
+        return events
+
+    N_pdfweights = ak.num(events.LHEPdfWeight, axis=1)
+    if ak.any(N_pdfweights != 103) or ak.any(N_pdfweights != 101):
+        raise Exception(f"Number of LHEPdfWeights ({N_pdfweights}) is not "
+                        f"as expected (103 or 101) in dataset {self.dataset_inst.name}")
+
+    # first 101 LHEPdfWeight values: pdf variations
+    # NOTE: we should find out which weight corresponds to the nominal weight (should always be 1)
+    pdfweights = events.LHEPdfWeight[:, :101]
+    pdfweights = ak.sort(pdfweights)
+
+    # PDF uncertainty by calculating the variance (underestimates uncertainty?)
+    # var = ak.var(events.LHEPdfWeight, axis=1)
+    # PDF uncertainty as 68% CL
+    var = (pdfweights[:, 83] - pdfweights[:, 15]) / 2
+
+    mean = ak.mean(events.LHEPdfWeight, axis=1)
+
+    # NOTE: use mean value as nominal pdf weight? or remove the necessity of adding this nominal weight?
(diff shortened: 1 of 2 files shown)


columnflow/production/pdf.py
```diff
@@ -0,0 +1,59 @@
...
+    pdfweights = events.LHEPdfWeight[:, 1:101] / pdf_weight_nominal
+    pdfweights = ak.sort(pdfweights)
+
+    # PDF uncertainty as 68% CL
+    var = (pdfweights[:, 83] - pdfweights[:, 15]) / 2
+
+    # NOTE: use mean value as nominal pdf weight? or remove the necessity of adding this nominal weight?
+    events = set_ak_column(events, "pdf_weight", pdf_weight_nominal)
+    events = set_ak_column(events, "pdf_weight_down", pdf_weight_nominal + var)
+    events = set_ak_column(events, "pdf_weight_up", pdf_weight_nominal - var)
```
riga, line 57: Mh, if I see correctly, then `var` (or `std` with the above comment), should be scaled first by `pdf_weight_nominal` again. As it is right now, `var` is the standard deviation of the weights that were previously scaled to nominal.

This means that if nominal is != 1, you are adding / subtracting a normalized value (sth around 1) to / from a value that can be (e.g.) 0.0001.

I think the columns in lines 56/57 should be `pdf_weight_nominal * (1 +- var)`.

columnflow/production/scale.py
```diff
@@ -0,0 +1,96 @@
...
+    Resources:
+
+       - https://cms-nanoaod-integration.web.cern.ch/integration/master/mc94X_doc.html
+    """
+
+    # stop here for data
+    if self.dataset_inst.is_data:
+        raise ValueError("attempt to read out mur/muf weights in data")
+
+    N_scaleweights = ak.num(events.LHEScaleWeight, axis=1)
```
riga, line 36: `n_weights`

columnflow/production/scale.py
```diff
@@ -0,0 +1,96 @@
...
+    events = set_ak_column(events, "muf_weight", events.LHEScaleWeight[:, 4])
+    events = set_ak_column(events, "muf_weight_up", events.LHEScaleWeight[:, 5])
+    events = set_ak_column(events, "muf_weight_down", events.LHEScaleWeight[:, 3])
+
+    return events
+
+
+@producer(
+    uses={"LHEScaleWeight"},
+    produces={"scale_weight", "scale_weight_up", "scale_weight_down"},
```
riga, line 59: We should call the produced columns sth like `murf_envelope_weight{,_up,_down}`.

columnflow/production/scale.py
```diff
@@ -0,0 +1,96 @@
...
+
+    if ak.any(events.LHEScaleWeight[:, 4] != 1):
+        raise Exception("The nominal LHEScaleWeight should always be 1")
+
+    events = set_ak_column(events, "scale_weight", events.LHEScaleWeight[:, 4])
+
+    # for the up/down variations, take the max/min value of all possible combinations
+    # except mur=2, muf=0.5 (index 2) and mur=0.5, muf=2 (index 6) into account
+    idx_mask = (ak.local_index(events.LHEScaleWeight) != 2) & (ak.local_index(events.LHEScaleWeight) != 6)
+    considered_scale_weights = events.LHEScaleWeight[idx_mask]
```
riga, line 92: `local_index` can be quite expensive. Should ak arrays support advanced indexing with lists? In this case this would simply become

```python
considered_scale_weights = events.LHEScaleWeight[..., [0, 1, 3, 4, 5, 7, 8]]
```

riga (changes requested):
Looks good overall, just

- a couple naming suggestions,
- one possible performance improvement, and
- a potential logical error regarding pdf weight normalization.

columnflow/production/pdf.py
```diff
@@ -0,0 +1,59 @@
...
+    pdf_weight_nominal = events.LHEPdfWeight[:, 0]
+    if ak.any(pdf_weight_nominal != 1):
+        print("The first entry of the LHEPdfWeight is not 1 but will be used as the nominal weight.")
+
+    # the following 100 LHEPdfWeight values: pdf variations
+    pdfweights = events.LHEPdfWeight[:, 1:101] / pdf_weight_nominal
+    pdfweights = ak.sort(pdfweights)
+
+    # PDF uncertainty as 68% CL
+    var = (pdfweights[:, 83] - pdfweights[:, 15]) / 2
```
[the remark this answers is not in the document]
mafrahm: Yes, this only works for a set of 100 weights, but it's also the recommended procedure given here: https://arxiv.org/pdf/1510.03865.pdf (page 50) and an exception is raised if the number of pdfweights is not as expected. I propose we try to generalize this when running across a dataset with a different number of pdfweights.

columnflow/production/pdf.py
mafrahm: you're right, or I just skip dividing by the nominal weight to begin with

columnflow/production/pdf.py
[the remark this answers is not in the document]
riga: Ok, sounds good!

columnflow/production/pdf.py
riga: But then the "weight" would not be usable anymore as a multiplicative weight out of the box, right?

mafrahm: yes, this is the case when all weights are normalised to one entry w_0 we know of, W_i = w_i / w_0. However, in some samples, the "w_0" is actually not the first entry in the list of pdf weights. So, assuming all weights are normalized to some w_N (where we don't know which weight this is or even if N is the same for all events), we cannot identify which weight the nominal weight is. So in my case, I then simply redefine the weight w_0 as the nominal weight by multiplying with W_0 = w_0 / w_N. When N=0, this should not change anything and when N!=0 we now use a different weight as the nominal one.

riga (approved):

PR #120, opened 2022-12-01
Rotate cutflow bin labels to prevent overlap.

dsavoiu:
The x axis labels in cutflow plots can quickly overlap if the selection step labels get too long, making them difficult to read. This PR proposes a quick fix by drawing the labels at an angle of 45 degrees.

### DIFF
--- a/columnflow/plotting/example.py
+++ b/columnflow/plotting/example.py
@@ -467,4 +467,8 @@ def plot_cutflow(
     # ratio plot not used here; set `skip_ratio` to True
     kwargs["skip_ratio"] = True
 
-    return plot_all(plot_config, style_config, **kwargs)
+    p = plot_all(plot_config, style_config, **kwargs)
+
+    plt.gca().set_xticklabels(xticklabels, rotation=45, ha="right")
+
+    return p


riga:
Hey there! Good idea, but just two quick questions:

- Are the labels only rotated when there is not enough horizontal space, or are they rotated regardless of space issues?
- `plt.gca` seems to use the global plot objects to access the *current* axis. I just discussed with @mafrahm and maybe all plot functions should return `fig, (ax1, ax2, ...)` (`fig, (ax1,)` in the 1D case) so that one can modify plot axes more easily. Would you agree?

dsavoiu:
Hi! To answer your questions:
* the labels are always rotated. I find in practice this is almost always desirable since some an overlap happens quite often, but of course this depends on how verbose the labels are or how many selection steps are plotted. In principle, one could check if the axis is overcrowded and do the rotation depending on that, but to do it properly one would need either a good heuristic and/or some way to check the label overlap in matplotlib. Personally, I would avoid the latter if possible, since it could get a bit hacky and also might yield different results for different backends).

* `plot.gca()` is used as a workaround here, but it would indeed be better to have access to the `axes` object(s) directly. I would be in favor of making plot methods return `(fig, axes)` or similar.

riga (approved):

PR #124, opened 2022-12-03
Apply JEC/JER smearing via correctionlib

dsavoiu:
This PR implements new calibrators for applying the jet energy corrections/jet energy resolution smearing and the associated uncertainties via correctionlib.

The implementation closely follows the previous one using `coffea.jetmet_tools`, but reads in the corrections and applies them to the jets using the fast vectorized evaluators provided by `correctionlib`. The new implementation is now the default, but the old one is kept in `calibration/jets_coffea.py` for reference and validation purposes.

To use the new implementation, an external file containing the correction tables needs to be added to the config. All JEC levels and uncertainties, as well as the JER values, scale factors, and scale factor variations in correctionlib format, are now provided centrally via a single gzipped JSON file, which must be supplied via a new `jet_jerc` entry in `cfg.x.external_files`. The `jec`, `junc`, `jer` and `jersf` external file dicts are not used by the new implementation and can be removed.

The `jec` and `jer` auxiliary config entries can be kept mostly as they are, but the `levels` key can now be a list with a compound correction level (typically `L1L2L3Res`) as its only element instead of the full list of correction levels. This compound correction applies all the required correction levels in the right order, taking care to propagate changes in the jet p<sub>T</sub> at each stage. The individual correction levels used as the starting point for the Type-I MET propagation must be configured via the `cfg.x.jec.levels_for_type1_met` key. The `source` key is no longer needed and can be removed.

To validate the new implementation, I compared some of the corrected jet and MET quantities to the old implementation and found only negligible differences:
```python
events_old = self[jets_coffea](events, **kwargs)
events_new = self[jets](events, **kwargs)

def compare(route, comp_func=lambda a, b: abs(a - b)):
    from columnflow.columnar_util import Route
    route = Route(route)
    old = route.apply(events_old)
    new = route.apply(events_new)
    comp = comp_func(old, new)
    return {
        "min":  round(ak.min(comp),  5),
        "mean": round(ak.mean(comp), 5),
        "max":  round(ak.max(comp),  5),
    }

>>> compare("Jet.pt")
... {'min': 0.0, 'mean': 0.0, 'max': 5e-05}
>>> compare("MET.pt")
... {'min': 0.0, 'mean': 2e-05, 'max': 0.00023}
>>> compare("MET.phi")
... {'min': 0.0, 'mean': 0.0, 'max': 0.0001}

>>> compare("Jet.pt_jec_Total_up")
... {'min': 0.0, 'mean': 1e-05, 'max': 0.00017}
>>> compare("MET.pt_jec_Total_up")
... {'min': 0.0, 'mean': 2e-05, 'max': 0.00032}

>>> compare("Jet.pt_jer_up")
... {'min': 0.0, 'mean': 0.0, 'max': 5e-05}
>>> compare("MET.pt_jer_up")
... {'min': 0.0, 'mean': 1e-05, 'max': 7e-05}
```

The correctionlib implementation appears to be about 2.7 (!) times faster than the one using `coffea.jetmet_tools`:

```bash
# coffea.jetmet_tools
iterate through 197113 events in 4 chunks ...
handling chunk 3
calibrator 'jets_coffea' took 14.107 seconds  # first chunk also loads the correction files
handling chunk 0
calibrator 'jets_coffea' took 7.374 seconds
handling chunk 1
calibrator 'jets_coffea' took 7.303 seconds
handling chunk 2
calibrator 'jets_coffea' took 7.183 seconds
done (took 38 seconds)
event processing in loop body took 36.04 seconds

# correctionlib
iterate through 197113 events in 4 chunks ...
handling chunk 3
calibrator 'jets_correctionlib' took 2.680 seconds
handling chunk 0
calibrator 'jets_correctionlib' took 2.732 seconds
handling chunk 1
calibrator 'jets_correctionlib' took 2.671 seconds
handling chunk 2
calibrator 'jets_correctionlib' took 2.662 seconds
done (took 12.69 seconds)
event processing in loop body took 10.82 seconds
```

### DIFF
--- a/columnflow/calibration/jets.py
+++ b/columnflow/calibration/jets.py
@@ -10,6 +10,7 @@ import law
 
 from columnflow.util import maybe_import, memoize
 from columnflow.calibration import Calibrator, calibrator
+from columnflow.calibration.util import propagate_met, ak_random
 from columnflow.production.util import attach_coffea_behavior
 from columnflow.columnar_util import set_ak_column
 
@@ -35,31 +36,6 @@ def get_basenames(struct):
         lambda p: os.path.splitext(os.path.basename(p[0] if isinstance(p, tuple) else p))[0],
         struct,
     )
-
-
-# https://github.com/scikit-hep/awkward/issues/489\#issuecomment-711090923
-def ak_random(*args, rand_func):
-    """
-    Return an awkward array filled with random numbers. The *args* must be broadcastable
-    awkward arrays and will be passed as positional arguments to *rand_func* to obtain the
-    random numbers.
-    """
-    args = ak.broadcast_arrays(*args)
-
-    if hasattr(args[0].layout, "offsets"):
-        # convert to flat numpy arrays
-        np_args = [np.asarray(a.layout.content) for a in args]
-
-        # pass flat arrays to random function and get random values
-        np_randvals = rand_func(*np_args)
-
-        # convert back to awkward array
-        return ak.Array(ak.layout.ListOffsetArray64(args[0].layout.offsets, ak.layout.NumpyArray(np_randvals)))
-
-    # pass args directly (this may fail for some array types)
-    # TODO: make function more general
-    np_randvals = rand_func(*args)
-    return ak.from_numpy(np_randvals)
 
 
 @memoize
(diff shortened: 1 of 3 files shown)


columnflow/calibration/jets_coffea.py
```diff
@@ -0,0 +1,646 @@
...
+from columnflow.production.util import attach_coffea_behavior
+from columnflow.columnar_util import set_ak_column
+
+np = maybe_import("numpy")
+ak = maybe_import("awkward")
+coffea = maybe_import("coffea")
+
+coffea_extractor = maybe_import("coffea.lookup_tools.extractor")
+coffea_jetmet_tools = maybe_import("coffea.jetmet_tools")
+coffea_txt_converters = maybe_import("coffea.lookup_tools.txt_converters")
```
riga, line 23: Is the main coffea package required here? If not, we could remove that single line, but if it is needed for the following three subpackage imports to work, we could remove the assignment, i.e.,

```python
maybe_import("coffea")
```

columnflow/calibration/jets.py
```diff
@@ -192,63 +114,103 @@ def jec(
...
+            # determine correct inputs (change depending on corrector)
+            inputs = [
+                variable_map[inp.name]
+                for inp in corrector.inputs
+            ]
+            correction = ak_evaluate(corrector, *inputs)
+            # update pt for subsequent correctors
+            variable_map.update({
+                "JetPt": variable_map["JetPt"] * correction,
+            })
```
riga, line 174: Maybe just

```python
variable_map["JetPt"] = variable_map["JetPt"] * correction
```

?

riga (changes requested):
Awesome, this looks extremely clean 😎

I only have to superficial syntax comments, no show stoppers or anything. I think we will move to the correctionlib-based version right away

columnflow/calibration/jets_coffea.py
dsavoiu: Good point. It doesn't look like the main package is required explicitly, so I removed the line.

riga (approved):

dsavoiu:
Thanks for the review and quick merge, glad there were no big issues :)

PR #126, opened 2022-12-05
Various improvements for plotting tasks/functions

mafrahm:
Changes in this PR:
- restructuring of plotting tasks
    - PlotVariables and PlotShiftedVariables tasks now inherit from the same base (PlotVariablesBase), using the same run method
    - PlotCutflow and PlotCutflowVariables also share a base (PlotCutflowBase)
    - reduction of the shift axis is now part of plotting functions for all tasks
- some `PerProcess` wrapper tasks are now included to produce 1 plot per process (for 2d plotting tasks and `PlotShiftedVariables`
- a `SettingsParameter` was implemented for the `process_settings` and the new `variable_settings` parameter
- `variable_settings` allows rebinding of individual variables and changing of the label associated to the variable
- plot functions now return the figure and all axes
- minor style changesChanges in this PR:
- restructuring of plotting tasks
    - PlotVariables and PlotShiftedVariables tasks now inherit from the same base (PlotVariablesBase), using the same run method
    - PlotCutflow and PlotCutflowVariables also share a base (PlotCutflowBase)
    - reduction of the shift axis is now part of plotting functions for all tasks
- some `PerProcess` wrapper tasks are now included to produce 1 plot per process (for 2d plotting tasks and `PlotShiftedVariables`
- a `SettingsParameter` was implemented for the `process_settings` and the new `variable_settings` parameter
- `variable_settings` allows rebinding of individual variables and changing of the label associated to the variable
- plot functions now return the figure and all axes
- minor style changes

### DIFF
--- a/columnflow/plotting/plot_util.py
+++ b/columnflow/plotting/plot_util.py
@@ -18,23 +18,26 @@ od = maybe_import("order")
 
 
 def apply_variable_settings(hists: dict, variable_settings: dict | None = None) -> dict:
+    """
+    applies settings from `variable_settings` dictionary to all histograms in `hists`.
+    """
     # check if there are variable settings to apply
     if not variable_settings:
         return hists
-    relevant_variables = set(list(hists.values())[0].axes.name) & set(variable_settings.keys())
+    relevant_variables = set(list(hists.values())[0].axes.name).intersection(variable_settings.keys())
     if not relevant_variables:
         return hists
 
     # apply all settings
     for var in relevant_variables:
         var_settings = variable_settings[var]
-        for key, h in hists.items():
+        for key, h in list(hists.items()):
             # apply rebinning setting
             rebin_factor = int(var_settings.get("rebin", 1))
             h = h[{var: hist.rebin(rebin_factor)}]
 
             # apply label setting
-            var_label = var_settings.get("label", None)
+            var_label = var_settings.get("label")
             if var_label:
                 h.axes[var].label = var_label
 
--- a/columnflow/plotting/plot_util.py
+++ b/columnflow/plotting/plot_util.py
@@ -17,6 +17,23 @@ mplhep = maybe_import("mplhep")
 od = maybe_import("order")
 
 
+def remove_residual_axis(hists: dict, ax_name: str, max_bins: int=1) -> dict:
+    """
+    removes axis named 'ax_name' if existing and there is only a single bin in the axis;
+    raises Exception otherwise
+    """
+    for key, hist in list(hists.items()):
+        if ax_name in hist.axes.name:
+            n_bins = len(hist.axes[ax_name])
+            if n_bins > max_bins:
+                raise Exception(
+                    f"{ax_name} axis of histogram for key {key} has {n_bins} values whereas at most "
+                    "{max_bins} is expected",
+                )
+            hists[key] = hist[{ax_name: sum}]
+
+    return hists
+
 def prepare_plot_config(
     hists: OrderedDict,
     shape_norm: bool | None = False,
--- a/columnflow/plotting/plot_util.py
+++ b/columnflow/plotting/plot_util.py
@@ -17,7 +17,34 @@ mplhep = maybe_import("mplhep")
 od = maybe_import("order")
 
 
-def remove_residual_axis(hists: dict, ax_name: str, max_bins: int=1) -> dict:
+def apply_variable_settings(hists: dict, variable_settings: dict | None = None) -> dict:
+    # check if there are variable settings to apply
+    if not variable_settings:
+        return hists
+    relevant_variables = set(list(hists.values())[0].axes.name) & set(variable_settings.keys())
+    if not relevant_variables:
+        return hists
+
+    # apply all settings
+    for var in relevant_variables:
+        var_settings = variable_settings[var]
+        for key, h in hists.items():
+            # apply rebinning setting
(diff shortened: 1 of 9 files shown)


columnflow/plotting/plot_util.py
```diff
@@ -17,6 +17,51 @@
 od = maybe_import("order")
 
 
+def apply_variable_settings(hists: dict, variable_settings: dict | None = None) -> dict:
```
riga, line 20: Could you perhaps add a docstring? Just to make it easier to dive in.

columnflow/plotting/plot_util.py
```diff
@@ -17,6 +17,51 @@
...
+    # check if there are variable settings to apply
+    if not variable_settings:
+        return hists
+    relevant_variables = set(list(hists.values())[0].axes.name) & set(variable_settings.keys())
```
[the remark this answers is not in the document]
riga: True, and I would even remove the `keys()` at the end ;) But could be ok to treat the structure a bit more verbose here ...

columnflow/plotting/plot_util.py
```diff
@@ -17,6 +17,51 @@
...
+    if not relevant_variables:
+        return hists
+
+    # apply all settings
+    for var in relevant_variables:
+        var_settings = variable_settings[var]
+        for key, h in hists.items():
+            # apply rebinning setting
+            rebin_factor = int(var_settings.get("rebin", 1))
```
riga, line 33: Just a question: can there be cases where `rebin` is present in `var_settings`, but `None`?

columnflow/plotting/plot_util.py
```diff
@@ -17,6 +17,51 @@
...
+            h = h[{var: hist.rebin(rebin_factor)}]
+
+            # apply label setting
+            var_label = var_settings.get("label", None)
```
riga, line 37: `var_label = var_settings.get("label")`

columnflow/plotting/plot_util.py
```diff
@@ -17,6 +17,51 @@
...
+    if not variable_settings:
+        return hists
+    relevant_variables = set(list(hists.values())[0].axes.name) & set(variable_settings.keys())
+    if not relevant_variables:
+        return hists
+
+    # apply all settings
+    for var in relevant_variables:
+        var_settings = variable_settings[var]
+        for key, h in hists.items():
```
riga, line 31: Could you use `list(hists.items())` here? Just to be sure that no unwanted behavior occurs since you are overwriting the keys a few lines below while iterating (this is most probably ok in this instance since you are reusing the same keys, but it might be even a bit more expressive here with `list` as in "we are going to change the dict while iterating").

columnflow/tasks/framework/parameters.py
```diff
@@ -0,0 +1,84 @@
...
+        value = float(value)
+    elif value.lower() == "true":
+        value = True
+    elif value.lower() == "false":
+        value = False
+    return (key, value)
+
+
+def serialize_setting(name, value):
+    return f"{name}={value}"
```
riga, line 24: Could you adapt them to be `classmethod`'s of `SettingsParameter`?

riga (changes requested):
Looks good to me! I only have a few minor comments below, which should be fairly easy to include.

columnflow/plotting/plot_util.py
mafrahm: The settings are parsed to either a bool, float or string, so None values should not be possible.

columnflow/plotting/plot_util.py
mafrahm: sure!

columnflow/tasks/framework/parameters.py
```diff
@@ -4,27 +4,34 @@
...
-    return f"{name}={value}"
-
-
-class SettingsParameter(law.CSVParameter):
+class SettingsParameterBase():
+    """
+    Base class to implement class methods for different types of setting parameters.
+    """
+    @classmethod
+    def parse_setting(self, setting: str) -> tuple[str, Union[float, bool, str]]:
```
riga, line 18: `self` → `cls`

columnflow/tasks/framework/parameters.py
```diff
@@ -4,27 +4,34 @@
...
+        if test_float(value):
+            value = float(value)
+        elif value.lower() == "true":
+            value = True
+        elif value.lower() == "false":
+            value = False
+        return (key, value)
+
+    @classmethod
+    def serialize_setting(self, name: str, value: str) -> str:
```
riga, line 30: `self` → `cls`

riga (changes requested):
Just two minor additional comments.

columnflow/tasks/framework/parameters.py
```diff
@@ -0,0 +1,91 @@
...
+import law
+from columnflow.util import test_float
+
+
+class SettingsParameterBase():
+    """
+    Base class to implement class methods for different types of setting parameters.
+    """
+    @classmethod
+    def parse_setting(self, setting: str) -> tuple[str, Union[float, bool, str]]:
```
riga, line 18: and `tuple[str, float | bool | str]` should work here.

riga (changes requested):

columnflow/tasks/framework/parameters.py
mafrahm: seems to not work here, not sure why

riga (approved):

PR #130, opened 2022-12-16
Generalize plot function handling.

### LINKED ISSUE #122: Generalize PlotVariablesBase to nD plots
The current implementation of PlotVariablesBase [1] is too strict regarding the implementation of plot functions, e.g. the implementation of a 1D plot that requires 2 variables (plotting an efficiency that takes as inputs the numerator and denominator) is not possible. This should be generalized at some point.

[1] https://github.com/uhh-cms/columnflow/blob/43db794483e2529e3858d6a45ef662220a3b72d5/columnflow/tasks/plotting.py#L155-L158

riga:
This PR generalizes the use of the `--plot-function` parameter. So far, there have been `*_1d` and `*_2d` versions of the same parameter and there was a switch in the generic PlotVariablesBase task that acted somewhat hardcoded, depending on the number of variables passed:

https://github.com/uhh-cms/columnflow/blob/43db794483e2529e3858d6a45ef662220a3b72d5/columnflow/tasks/plotting.py#L155-L158

If I understand correctly, this was intended for situations where both 1D and 2D variables were plotted at the same time. However, now that we have dedicated `PlotVariables{1,2}D` tasks (and for cutflows plots alike), this is no longer required. Therefore, it seems like merging the parameters into a single `--plot-function` parameter is a logical step forward (but correct me if I'm wrong @mafrahm).

Closes #122.

### DIFF
--- a/columnflow/tasks/framework/parameters.py
+++ b/columnflow/tasks/framework/parameters.py
@@ -6,33 +6,9 @@ Custom luigi parameters.
 
 from __future__ import annotations
 
-import copy
-
 import law
-import luigi
 
 from columnflow.util import test_float, DotDict
-
-
-class PlotFunctionParameter(luigi.Parameter):
-    """
-    Plain parameter subclass that provides a convenience method for copying an instance, assigning
-    a different default value and optionally changing the description text.
-    """
-
-    def with_default(
-        self,
-        default: str,
-        description: str | None = None,
-        amend_description: bool = False,
-    ) -> PlotFunctionParameter:
-        inst = copy.copy(self)
-        inst._default = default
-        if description is not None:
-            inst.description = description
-        elif amend_description:
-            inst.description += f"; default: {default}"
-        return inst
 
 
 class SettingsParameter(law.CSVParameter):
--- a/columnflow/tasks/framework/parameters.py
+++ b/columnflow/tasks/framework/parameters.py
@@ -25,7 +25,7 @@ class PlotFunctionParameter(luigi.Parameter):
         default: str,
         description: str | None = None,
         amend_description: bool = False,
-    ) -> luigi.Parameter:
+    ) -> PlotFunctionParameter:
         inst = copy.copy(self)
         inst._default = default
         if description is not None:
--- a/columnflow/tasks/framework/parameters.py
+++ b/columnflow/tasks/framework/parameters.py
@@ -4,11 +4,35 @@
 Custom luigi parameters.
 """
 
-from typing import Union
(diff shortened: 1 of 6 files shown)


columnflow/tasks/framework/parameters.py
```diff
@@ -4,13 +4,37 @@
...
+    ) -> luigi.Parameter:
+        inst = copy.copy(self)
+        inst._default = default
+        if description is not None:
+            inst.description = description
+        elif amend_description:
+            inst.description += f"; default: {default}"
+        return inst
+
+
```
riga, line 37: This pattern for copying parameter instances could be generalized, too, if it proves effective.

columnflow/tasks/framework/parameters.py
```diff
@@ -28,7 +52,7 @@ class SettingsParameter(law.CSVParameter):
     """
 
     @classmethod
-    def parse_setting(cls, setting: str) -> tuple[str, Union[float, bool, str]]:
+    def parse_setting(cls, setting: str) -> tuple[str, float | bool | str]:
```
riga, line 55: Turns out, we forgot to import annotations ;)

columnflow/tasks/plotting.py
```diff
@@ -220,14 +209,20 @@ class PlotVariables1D(
     PlotVariablesBaseSingleShift,
     PlotBase1D,
 ):
-    pass
+    plot_function = PlotBase.plot_function.copy(
+        default="columnflow.plotting.example.plot_variable_per_process",
+        add_default_to_description=True,
```
mafrahm, line 214: where is the implementation of the `add_default_to_description` functionality?

mafrahm (approved):
Looks good to me!

Only for `PlotCutflowVariables1D`, the plot parameters are still hard-coded (based on the `per_plot` parameter), I'm not sure how to remove these.

columnflow/tasks/plotting.py
riga: It's in law at https://github.com/riga/law/blob/02dc3744f87fa0a6c2202ff0dc8b80ca885f443b/law/patches.py#L369-L399

I'm going to open a PR directly to luigi, but in the meantime we can use that "patch".

riga:
> Only for PlotCutflowVariables1D, the plot parameters are still hard-coded (based on the per_plot parameter), I'm not sure how to remove these.

I guess this would require a dedicated PR to disentangle the two types of plots being produced, whether to show certain steps or certain processes.

PR #133, opened 2022-12-19
Add CoalesceColumns task.

riga:
This PR adds an additional task that horizontally unites / coalesces columns (unlike the `Merge*` tasks that work vertically) from

- MergeReducedEvents
- ProduceColumns
- MLEvaluation

to have a point in the workflow where we can save the state of columns used for creating histograms also into parquet files. Use cases are (e.g.)

- giving combined files to students
- passing events to other groups for synchronization purposes.

The implementation is straight forward and resembles that of other tasks.

### DIFF
--- a/columnflow/tasks/coalesce.py
+++ b/columnflow/tasks/coalesce.py
@@ -83,7 +83,7 @@ class CoalesceColumns(
 
     @MergeReducedEventsUser.maybe_dummy
     def output(self):
-        return self.target(f"columns__{self.branch}.parquet")
+        return self.target(f"data_{self.branch}.parquet")
 
     @law.decorator.log
     @law.decorator.localize(input=True, output=False)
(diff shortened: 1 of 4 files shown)


columnflow/tasks/coalesce.py
```diff
@@ -0,0 +1,152 @@
...
+        if self.ml_models:
+            files.extend([inp.path for inp in inputs["ml"]])
+        for (events, *columns), pos in self.iter_chunked_io(
+            files,
+            source_type=len(files) * ["awkward_parquet"],
+            # TODO: not working yet since parquet columns are nested
+            # open_options=[{"columns": load_columns}] + (len(files) - 1) * [None],
+        ):
+            # add additional columns
+            events = update_ak_array(events, *columns)
```
mafrahm, line 121: Just to clarify: if the same column exist in multiple files, the priority is:

MLEvaluation > ProduceColumns > ReduceEvents ?

mafrahm (approved):
Looks good!

PR #142, opened 2023-01-12
Add analysis template

BalduinLetzer:
Adds an analysis template for newcomers and a curlable bash script to download the template and initialize it as a git repo.
After merge, the checkout link in the script to download the template needs to be changed to point to the correct branch.

PR #148, opened 2023-01-26
Move cms related array functions and tasks to cms/ subdirectories.

### LINKED ISSUE #145: Highlight cms-dependent tasks & producers
Most of the columnflow code basis is experiment agnostic and could in principle also be used by others. For the purpose of clarity & documentation, we could highlight those parts of the project (in terms of tasks and producers) that are in fact cms-based.

In the long term, this could even improve the structure of the project (of course, up to the point where a generalization would impose a significant inconvenience).

Required changes:

- **Structural**
  - Move selectors / calibrators / producers into "cms" subdirectories, e.g. `from columnflow.production.btag import btag_weights` would become `from columnflow.production.cms.btag import btag_weights`.
  - Move certain tasks to "cms" subdirectories. Example: `CreatePileupWeights`.
- **Task logic**
  - The LFN lookup in `GetDatasetLFNs` is perhaps cms-centric right now, but other experiments might use a similar mechanism. A switch of the method does already exist but we should check how things are done e.g. in Atlas.
  - We should implement the `WritePyhfWorkspace` task.
- **Setup**
  - Some CMS-related variables are queried during the setup.
- **Plotting**
  - Minor changes in the default plotting methods, mostly regarding the "CMS" label, but I'd consider this to be trivial (@mafrahm).

riga:
This PR introduces a better distinction between generic, experiment-agnostic parts of columnflow and CMS-related ones. In particular,

- some calibrators & producers are moved into a cms/ subdirectory,
- the datacard writer is moved into `inference/cms/datacard.py`, and
- two tasks are moved into tasks/cms, namely
  - `CreatePileupWeights` in `tasks/cms/external.py`
  - `CreateDatacards` in `tasks/cms/inference.py`.

If merged, analyses only might need to adapt some import statements.

For the code review: The general functionality is fully preserved and there are no code changes. Still, we should have a quick review process.

Closes #145.


columnflow/production/example.py
```diff
@@ -6,7 +6,6 @@
 
 from columnflow.production import Producer, producer
 from columnflow.production.normalization import normalization_weights
-from columnflow.production.pileup import pu_weight
```
pkausw, line 9: I understand why you removed the CMS specific modules from the example producer. It might make sense to create a CMS specific example though, just to make sure this knowlegde doesn't get lost!

columnflow/selection/example.py
```diff
@@ -7,8 +7,6 @@
 from collections import defaultdict
 
 from columnflow.selection import Selector, SelectionResult, selector
-from columnflow.production.mc_weight import mc_weight
-from columnflow.production.seeds import deterministic_seeds
```
pkausw, line 11: As stated in my other comment, I don't really see a reason to remove these modules from a general example. But maybe I'm missing something

columnflow/tasks/cms/external.py
```diff
@@ -0,0 +1,151 @@
...
+        self.output().dump(weights, formatter="json")
+
+    @classmethod
+    def read_mc_profile_from_cfg(
+        cls,
+        pu_config_target: law.FileSystemTarget,
+    ) -> list[float]:
+        """
+        Takes a mc pileup configuration file stored in *pu_config_target*, parses its content and
+        return the pu profile as a list of float probabilities.
```
pkausw, line 87:
```suggestion
        returns the pu profile as a list of float probabilities.
```

columnflow/tasks/cms/external.py
```diff
@@ -0,0 +1,151 @@
...
+        Takes the pileup profile in data read stored in *pu_file_target*, which should have been
+        produced when processing data, and a *minbias_mexs* value in mb (milli), builds the ratio to
+        mc and returns the weights in a list for 99 bins (recommended number).
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def normalize_values(cls, values: Sequence[float]) -> list[float]:
+        _sum = sum(values, 0.0)
+        return [value / _sum for value in values]
```
pkausw, line 143: I agree this is a very safe way to handle general `Sequence`s of floats. But the `for` loop isn't very vectorized, and the input values could also just be a numpy/awkward array, right?

columnflow/tasks/cms/external.py
```diff
@@ -0,0 +1,151 @@
...
+    def read_data_profile_from_hist(
+        cls,
+        pu_hist_target: law.FileSystemTarget,
+    ) -> list[float]:
+        """
+        Takes the pileup profile in data preproducd by the lumi pog and stored in *pu_hist_target*,
+        builds the ratio to mc and returns the weights in a list.
+        """
+        hist_file = pu_hist_target.load(formatter="uproot")
+        probs = hist_file["pileup"].values().tolist()
```
pkausw, line 124: Why `to_list()`? The return type of `uproot.Histogram.values()` is a `numpy.ndarray`, right? Shouldn't this also suffice here (and also profit from numpy's power in general)?

law.cfg
```diff
@@ -10,7 +10,6 @@ columnflow.tasks.ml
 columnflow.tasks.union
 columnflow.tasks.histograms
 columnflow.tasks.plotting
-columnflow.tasks.inference
```
pkausw, line 13: Not an expert on `law` configs, sorry :D Why is this just gone? Does this mean that the CMS inference task isn't available anymore?

pkausw (commented):
I'm not sure about the following moves:
- [ ] `gen_top_decay.py`: I read through the module, and it doesn't seem to me like this is _cms_ specific per say. The way I read this, this module is more specific to a generator for top physics, but that could be the same between experiments, e.g. ATLAS and CMS, right? Though this would probably be not as important to other experiments, e.g. LHCb... Still it doesn't seem particularly CMS specific to me
- [ ] `mc_weight.py`: similar to the `gen_top_decay` module, this seems experiment-agnostic to me
- [ ] `pdf.py`: This also seems more PDF-/MC-generator-dependent to me. A certain set of PDF (used in a MC generator) can also be used be different experiments, right?
- [ ] `scale.py`: similar comments as before. The LHE weights and their treatment are similar (if not the same) between different experiments since it's a definition from the MC developers, right?
- [ ] `seeds.py`: The definition for the random seeds seem to depend on _objects_ like e.g. electrons and muons, but those objects should be (to some extent) experiment-agnostic, right? I think the module should also be usable out of the box in ATLAS (provided the input .root files have the correct field of course)

Of course, it's always possible that I missed something, so feel free to call me out on that! I create the bullets as clickable check points so we can tick them of as we discuss them in this comment thread

law.cfg
mafrahm: As far as I understand, the inference tasks could not be called from the default cf setup, but you can simply add that module to the `[modules]` in your own analysis-specific law.cfg

riga: Yup, that would be the way to go :)

columnflow/tasks/cms/external.py
riga: This is just to be in line with what `read_mc_profile_from_cfg` does, i.e., returning a list of floats that are passed to `normalize_values`. In total, there are 100 values, in the hists, so there shouldn't be too big of a performance loss :)

columnflow/tasks/cms/external.py
riga: True. But anyway we deal with a maximum of 100 values here, so this should be too significant.

columnflow/production/example.py
riga: With #142 opened, I think would anyway drop the example files as we could have all example snippets (also in e.g. a cms-specific flavor) in a central space (the "analysis_template" subdirectory). This is of course up to discussion, but I think it would be favorable to have only a single place for these type of examples.

In this case, we should of course adapt #142 to remove the `example.py`'s.

columnflow/selection/example.py
riga: See above :)

pkausw (commented):
Looks good to me. I just want to follow up on the general producers I was wondering about in my first review - but I can't find the comment anymore :/

pkausw:
Since I really can't find my original comment with the list of producers I wasn't sure about moving, I'll try it again with this separate one. I was wondering whether it really makes sense to move the following list of producers into the cms directory:

- [ ] `mc_weights`
- [ ] `gen_top_decay`
- [ ] `pdf`
- [ ] `scale`
- [ ] `seed`

Apologies if you already answered this, but for whatever reason, I really cannot find the thread again 😅

riga:
All good :)

I was wondering the same thing about the producers you listed, since they are really rather generic. However, in the end this impression might be biased and I decided based on the names of the columns they `use` which are quite cms-specific I think.

pkausw:
> All good :)
> 
> I was wondering the same thing about the producers you listed, since they are really rather generic. However, in the end this impression might be biased and I decided based on the names of the columns they `use` which are quite specific and reliant on CMS NanoAOD naming schemes.

Ah! Hm, that is of course true... Do you think it might make sense to create a base class or something like it for this? Though I guess this would be difficult since the column names are used by name in the function of the respective modules.

One thing that comes to my mind whould be to generate a rather abstract class, where the required columns are member objects. So instead of e.g. calling `events.LHEScaleWeight`, we would first initialize a member variable with the (list of) column names and then use `events[self.column_name]` in the functions of the respective producers. It should be rather straight forward to derive CMS- or ATLAS-specific producers with this base producer, without actually duplicating lines of code. However, I can see how this might also be overkill. What do you think?

riga:
> It should be rather straight forward to derive CMS- or ATLAS-specific producers with this base producer, without actually duplicating lines of code. However, I can see how this might also be overkill. What do you think?

I think in the long term, once we know a bit better how things at e.g. ATLAS work, we could consider this. Actually, the `Route` class does already what you are proposing, e.g.

```python
r = Route("some.nested.column")
some_nested_column = r.apply(events)
...
```

However, I guess it's best to wait for the point where this type of column / producer generalization becomes necessary before implementing it.

pkausw:
> > It should be rather straight forward to derive CMS- or ATLAS-specific producers with this base producer, without actually duplicating lines of code. However, I can see how this might also be overkill. What do you think?
> 
> I think in the long term, once we know a bit better how things at e.g. ATLAS work, we could consider this. Actually, the `Route` class does already what you are proposing, e.g.
> 
> ```python
> r = Route("some.nested.column")
> some_nested_column = r.apply(events)
> ...
> ```
> 
> However, I guess it's best to wait for the point where this type of column / producer generalization becomes necessary before implementing it.

Fine by me. Since this is already quite modular, I think it probably won't get too tangled up in the long run, so I agree that a generalization at a later point in time should also be possible w/o complications :+1:

pkausw (approved):

PR #149, opened 2023-01-27
Restructuring of plotting functions

mafrahm:
This PR contains:

- plotting tasks now create copies of variable_insts, category_insts and process_insts such that tasks can change these instances without interfering with other tasks running in parallel
- the application of plotting parameters in the plotting functions has been generalized by applying the changes directly to variable_insts, process_insts or histograms
- the plot function files are renamed

New features:
- the `discrete_x` tag from a variable_inst is now used in plot functions
- there is a new `density` parameter, which scales the bin values by the bin width


riga (approved):
Looks all good to me 🎉

PR #150, opened 2023-01-27
Take `shape_norm` into account for data/MC ratio.

dsavoiu:
This PR changes how the data/MC ratio in the lower pane of the plots is calculated when `shape_norm` is enabled.

Since with `shape_norm` the upper pane shows the data points and MC stack normalized to their respective integrals, I think the natural expectation here is to see a ratio between those two (normalized) values in the lower pane. That is, for bin $i$ with data $D_i$ and the sum of all stacked MC processes $M_i$, the ratio shown should be calculated as
$$\frac{D_i/\sum_{j}{D_j}}{M_i/\sum_{k}{M_k}}$$ rather than $\frac{D_i}{M_i}$.


riga (approved):
I never created normalized histograms with data involvement, but thinking about it, you are probably right.

I'd let @mafrahm comment on whether there is perhaps another place where this change needs to be applied.

mafrahm:
I agree with your change and will merge it, but we should keep in mind to check what happens when no processes are included in the stack (then the task will most likely fail), but this is not a regular use case anyway.
Regarding the other functions, we should also have a look into the ratio plots of `plot_variable_variants` (shows the selection efficiency) and `plot_shifted_variable` (shows the ratio between nominal and up/down variation) functions. Could be useful to remove rate differences in the ratio there aswell.

PR #151, opened 2023-01-30
Share prepared ML events between ML models.

riga:
So far, each ML model requires its own minimal data preprocessing pipeline consisting of

```mermaid
graph TD
    PrepareMLEvents(PrepareMLEvents)
    MergeMLEvents(MergeMLEvents)
    MLTraining(MLTraining)

    PrepareMLEvents ==folds==> MergeMLEvents
    MergeMLEvents --folds--> MLTraining
```

However, two or more models that could require the same inputs strictly trigger different pipelines which is extremely redundant.

This PR adds a slightly different mixin for the `{Prepare,Merge}MLEvents` task that does not add the model name to the store path, but rather a configurable `store_name`. When *None* (the default), it falls back to the model name (`cls_name`) though.

The additional mixin is needed since for the two task classes mentioned above, the model itself is actually not significant but just its `store_name` attribute. This is achieved through a hidden parameter (as also done in many other places) that ensures significance in the luigi context.

### DIFF
--- a/columnflow/ml/__init__.py
+++ b/columnflow/ml/__init__.py
@@ -76,6 +76,10 @@ class MLModel(Derivable):
     # default number of folds
     folds = 2
 
+    # default name for storing e.g. input data
+    # falls back to cls_name if None
+    store_name = None
+
     def __init__(
         self,
         config_inst: od.Config,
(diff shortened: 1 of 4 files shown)


mafrahm (approved):
Looks good to me. I'll also check if it works right away

PR #159, opened 2023-02-08
Add first draft for config_utils.

### LINKED ISSUE #152: Provide config_util.py
Some parts of columnflow, such as producers/selectors and tasks expect the analysis config to have certain fields in a well-defined structure. For newcomers, this structure is not necessarily trivial to a) understand and b) replicate.

We should provide a `config_util.py` that contains a couple small functions that help creating these objects. Examples:

- shift aliases
- external files
- event weight definitions
- sandbox bundles

riga:
This PR adds `config_util.py` which, for now, contains three functions that are helpful when setting up an `order.Config` object in a structure that columnflow needs. See #152 for more info.

**Note:** The `get_root_processes_from_campaign` is moved over from `util.py` which might cause an import error in existing analysis.

Closes #152.


mafrahm (approved):
Looks good to me!

PR #160, opened 2023-02-08
Allow missing 'objects' in selection results.

dsavoiu:
This PR handles the case where no `objects` are specified in the `SelectionResult` returned by an exposed selector.

### DIFF
--- a/columnflow/tasks/reduction.py
+++ b/columnflow/tasks/reduction.py
@@ -111,13 +111,14 @@ class ReduceEvents(
         read_sel_columns.add(Route("steps.*" if self.selector_steps else "event"))
         # add object masks, depending on the columns to write
         # (as object masks are dynamic and deeply nested, preload the meta info to access fields)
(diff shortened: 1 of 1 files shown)


columnflow/tasks/reduction.py
```diff
@@ -172,18 +173,19 @@ def run(self):
...
-                dst_names = list(sel["objects", src_name].fields)
-                if src_name in dst_names:
-                    # move to the end
-                    dst_names.remove(src_name)
-                    dst_names.append(src_name)
-                for dst_name in dst_names:
-                    object_mask = sel.objects[src_name, dst_name][event_mask]
-                    dst_collection = events[src_name][object_mask]
-                    events = set_ak_column(events, dst_name, dst_collection)
+            if hasattr(sel, "objects"):
```
riga, line 176:
```suggestion
            if "objects" in sel.fields:
```

riga (changes requested):
Nice :)
Just a clickable suggestion to use the `fields` of the `sel` array.

riga (approved):

PR #162, opened 2023-02-08
Handling of empty chunks

### LINKED ISSUE #161: Handling arrays with zero rows/entries
The array reduction performed during `cf.ReduceEvents` may result in an empty set of selected events. In the current implementation (revision [8b41c5](https://github.com/uhh-cms/columnflow/tree/8b41c5a13686fd6c539d2848f79780b1c8b50840)), this results in an error when writing to Parquet. The problem arises because the separation of an array into chunks yields an empty list if the array has zero entries:
```
Traceback (most recent call last):
  ...
  File "modules/columnflow/columnflow/tasks/production.py", line 117, in run
    law.pyarrow.merge_parquet_task(self, sorted_chunks, output, local=True)
  File "modules/columnflow/modules/law/law/contrib/pyarrow/util.py", line 132, in merge_parquet_task
    merge(inputs, output)
  File "modules/columnflow/modules/law/law/contrib/pyarrow/util.py", line 115, in merge
    merge_parquet_files(
  File "modules/columnflow/modules/law/law/contrib/pyarrow/util.py", line 35, in merge_parquet_files
    raise Exception("cannot merge empty list of parquet files")
Exception: cannot merge empty list of parquet files
```

For some tighter selections, having zero-length arrays after `ReduceEvents` can happen occasionally, so it would make sense to support this use case in the long run.

The most straightforward way would be to simply write the zero-length array to Parquet, making sure that column/schema information is preserved. Until recently, reading such empty arrays from Parquet was not implemented in Awkward but this has been added starting with version [2.0.7](https://github.com/scikit-hep/awkward/releases/tag/v2.0.7).

Such an implementation would also require either short-circuiting the chunked processing and writing out the empty array directly, or otherwise adapting the chunking logic to always produce at least one chunk, which would be empty in the zero-length case.

There also remain some limitations when handling zero-length arrays with Awkward + Arrow/Parquet (in particular regarding nullability for nested/ragged columns), for which workarounds would be needed.

An alternative option would be to produce no output file for zero-length arrays, but this would probably interfere with the workflow management, which relies on the presence of output files to check task completion.

dsavoiu:
This PR proposes an adaptation of the chunking logic to handle zero-length arrays (see also issue #161).

It would ensure that at least one (empty) chunk is produced in the zero-length case, but it may have other unintended side-effects for non-empty arrays, which should be checked. The necessity and implications of skipping the nullability workaround for empty arrays should also be evaluated.

Still a work in progress, so input is very welcome. It may also not be needed at all if we decide to handle empty arrays differently.

Fixes #161.


riga (approved):

riga:
I added the required changes to some downstream tasks that required changes, most notably

- `ProduceColumns`, and
- `CreateHistograms`.

@dsavoiu It seems that I can't request your review since you opened the PR, but could you still have a look?

dsavoiu:
@riga Thanks for the review! I hadn't considered all the implications for the downstream tasks, so it's great to see the zero-length case is now being handled there as well. After running some tests with the current version it looks like everything behaves as expected, so I'd say this is ready to be merged.

PR #179, opened 2023-02-09
change MLTraining to workflow with parameter *folds*

mafrahm:


riga:
- Removed `folds` parameter, as the number of folds is already determined by the ml model.
- Added support for scheduler messages potentially sent while in running state -> they can be helpful during the training loop to interact with training parameters.
- Added a method to the ml model interface for simple parsing of messages and the option to respond back to the scheduler.

riga (approved):

PR #180, opened 2023-02-10
Fix shift resolution.

riga:
This PR fixes a rather complex issue related to the shift resolution of tasks like `cf.PlotVariables1D` that don't have a particular dataset, but rather loop over several datasets.

When e.g. requested with shift `tune_up` and the resolved `--datasets` have at least one dataset that implements the shift, the entire task should accept `tune_up`. This was not the case before and `cf.PlotVariables1D` always fell back to `nominal`.

The actual change is in `DatasetsProcessesMixin` as the mixin that introduces the `--datasets` parameter. However, to make things easier, there are a couple of generalizations within `base.py` and `mixins.py` regarding the resolution of parameters as well as shifts.

Here, the most important changes are:

#### Parameters

- `--shift` is - as before - the requested shift.
- There is an internal parameter called `--task-shift` now that carries information on the resolved shifts (more info below), obtained within the `modify_param_values(cls, params)` class method, over to the task initialization. It should **not be used from the cli** and never be set manually (in fact, it's hidden and setting it would have no effect anyway).

#### Shift resolution

- When a shift is requested (through `--shift`), a decision is made in `modify_param_values(cls, params)` based on the shifts implemented by the task itself as well as shifts implemented by upstream tasks.
  A. When a shift is *implemented by a task*, then the task itself is meant to change some of it's behavior to *apply the shift*, either in the form of adding column aliases, changing selection cuts, changing weight computations, etc.
  B. When an upstream task implements a shift, but the task in question does not, it's input - and hence, it's output, too - still depends on this shift, although the task behavior does not change.
  C. There can be cases where a shift is implemented both by the task itself **and** at least one of it's upstream dependencies. In this case, both the input to the task changes, as well as the task behavior.

#### Task attributes

To distinguish these cases, two objects are accessible by the task: `shift_inst` and `effective_shift_inst`.

- `shift_inst` refers to the nominal shift in case the task **does not** implement the requested shift ifself (cases **A** or **C**), and to the requested shift **if it does**.
- `effective_shift_inst` refers to the requested shift if the task itself **or** an upstream dependency implements it (any of the cases **A**, **B** or **C**), and to the nominal shift otherwise.
- In short, when the task output will differ in any way from that of the nominal one (again, either through a varied input due to an upstream task implementing the shift, or the task itself), `effective_shift_inst` will refer to that shift, whereas `shift_inst` might not. Consequently, the construction of the store path of the tasks output is based on the name of the `effective_shift_inst`.


columnflow/tasks/framework/base.py
```diff
@@ -508,59 +518,59 @@ def modify_param_values(cls, params):
...
                 params["shift"] = "nominal"
-                params["effective_shift"] = "nominal"
+                params["task_shift"] = "nominal"
 
-        # store references
-        params["requested_shift_inst"] = config_inst.get_shift(params["shift"])
-        params["effective_shift_inst"] = config_inst.get_shift(params["effective_shift"])
+        # store references (not the meaning of effective and task shifts here)
+        params["effective_shift_inst"] = config_inst.get_shift(params["shift"])
+        params["shift_inst"] = config_inst.get_shift(params["task_shift"])
```
mafrahm, line 537: One minor thing: the nomenclature here is a bit confusing, since `shift_inst` does not correspond to the `shift`parameter. Maybe we could rename the `shift_inst`to `task_shift_inst`?

mafrahm (approved):
Looks good to me! Just one minor thing regarding nomenclature.

columnflow/tasks/framework/base.py
riga: Thanks for the review!

This exact question is also still on my mind. I actually called it `task_shift_inst` first (and had a `@property` named `shift_inst` as an alias), but then `self.task_shift_inst` or even `self.task.task_shift_inst` in (e.g.) producers or the ml model seemed a bit redundant. What do you think?

riga: Also @dsavoiu  :)

riga:
As communicated privately, I changed the names in the last commit.

- `--shift` remains the main parameter to configure the shift via the interface.
- `--local-shift` is the internal parameter that carries information of the shift resolution to the task instantiation. It is not to be used manually and is hidden from the interface.
- `global_shift_inst` (previously `effective_shift_inst`) refers to the shift that is *either* implemented by the task, *or* any task upstream.
- `local_shift_inst` (previously `shift_inst`) refers to the shift that is implemented by the task.
- `local_shift_inst` is forwarded to task array function instances (producers, etc) as `shift_inst`.

@mafrahm @dsavoiu

riga:
An additional note: the separation of shift aliases into selection dependent and independent parts is no longer needed. It will be removed in the next PR.

mafrahm (approved):

dsavoiu:
@riga Great, thanks for implementing the changes! I think the names are more intuitive now using `local` and `global` than they were before with `effective` etc. OK to be merged from my side.

PR #181, opened 2023-02-13
add htcondor_memory parameter

mafrahm:


riga:
Good idea. Maybe you even want to use the `BytesParameter` (e.g. used here: https://github.com/uhh-cms/nano_merger/blob/master/nano_merger/tasks.py#L73) that provides some convenience related to parsing.

riga (approved):

PR #182, opened 2023-02-14
Top-to-bottom ordering of stacked processes.

dsavoiu:
This PR changes the order in which processes are plotted as part of a stacked histogram, so that their order from top to bottom corresponds to the one given via the `processes` task parameter.

In the current implementation, the order of the stacked processes appears to be reversed. While this ordering technically makes sense when considering that the stack is constructed from the bottom up, it seems to me somewhat counter-intuitive, as I would expect the first named process to appear at the top.

Another related issue addressed by this PR is the ordering of the legend handles, which I think should be sorted so that they visually match the process stack.

### DIFF
--- a/columnflow/plotting/plot_util.py
+++ b/columnflow/plotting/plot_util.py
@@ -214,7 +214,9 @@ def prepare_plot_config(
         h_data = sum(data_hists[1:], data_hists[0].copy())
     if mc_hists:
         h_mc = sum(mc_hists[1:], mc_hists[0].copy())
-        h_mc_stack = hist.Stack(*mc_hists)
+        # reverse hists when building MC stack so that the
+        # first process is on top
+        h_mc_stack = hist.Stack(*mc_hists[::-1])
 
     # setup plotting configs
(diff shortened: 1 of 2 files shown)


columnflow/plotting/plot_util.py
```diff
@@ -227,10 +229,10 @@ def prepare_plot_config(
...
             "kwargs": {
                 "norm": mc_norm,
-                "label": mc_labels,
-                "color": mc_colors,
-                "edgecolor": mc_edgecolors,
-                "linewidth": [(0 if c is None else 1) for c in line_colors],
+                "label": mc_labels[::-1],
+                "color": mc_colors[::-1],
+                "edgecolor": mc_edgecolors[::-1],
+                "linewidth": [(0 if c is None else 1) for c in mc_colors[::-1]],
```
dsavoiu, line 235: Not strictly related to this PR, but it looks like this should actually be `mc_colors` instead of `line_colors`?

mafrahm (approved):
Looks good to me!

PR #186, opened 2023-02-17
Add config_util to create combinations of categories.

riga:
This PR adds a small helper to our config_util that creates the product of categories and creates subcategories for all combinations.

The help text is actually describing it already:

https://github.com/uhh-cms/columnflow/blob/9b16cfb87ec421f3fcf6d1d944f1c57fa072d040/columnflow/config_util.py#L93-L131


mafrahm (approved):
Looks good to me!

PR #187, opened 2023-02-20
Make plot tasks resolve categories after initializing dependent TaskArrayFunctions.

dsavoiu:
This PR modifies the method resolution order of plotting tasks so that the category lookup is done after all dependent calibrators, selectors and producers have been initialized.

This is done to support the use case where categories are added dynamically by the `init` functions of these objects.


dsavoiu:
I just noticed that this was already partially addressed in #186 (commit [4a6e3e](https://github.com/uhh-cms/columnflow/commit/4a6e3e5729ce9062439b1f438b0e76c01df3f5ee)). This PR would then add the same fix for the `PlotCutflow` task as well.

riga:
Thanks for opening this!

Mh, on Friday there was a similar issue in the non-cutflow pipeline (https://github.com/uhh-cms/columnflow/commit/4a6e3e5729ce9062439b1f438b0e76c01df3f5ee), but here there was the need to move the `CategoriesMixin` *in front* of e.g. the `CalibratorsMixin`.

I think the reason for this difference is they have different base tasks, or to be more specific, the first common base task they share is at a different depth, which is why the resolution hits the `CategoriesMixin` later in your case.

I will try to draw the MRO again later to see if we have an MRO issue elsewhere as well, in which case the wrong order in `PlotCutflowBase` would just be a symptom.

dsavoiu:
@riga Thanks for the quick reply! Looks like we stumbled onto the same issue independently ;)

Just to clear any confusion: I think the `CategoryMixin` should indeed also appear in front of/before the `CalibratorMixin` etc. in the inheritance list of both `PlotVariablesBase` and `PlotCutflow`, but I agree the actual order of the calls should be checked if the inheritance structure leads to a different MRO than expected.

In any case, the `PlotVariables1D` fix was certainly more important than the one for `PlotCutflow`, since the latter depends on only on calibrators which should in principle not need to add categories dynamically. Still, for the sake of consistency we should try to understand for which tasks a similar fix would also be needed in principle.

riga:
Looking into this again with more time, I think I confused sth my last comment. Of course you are absolutely right, and your fix is doing exactly what was done in https://github.com/uhh-cms/columnflow/commit/4a6e3e5729ce9062439b1f438b0e76c01df3f5ee.

For future reference, here's the MRO with your fix:

```mermaid
graph TD
    PlotCutflowBase(PlotCutflowBase 1)
    SelectorStepsMixin(SelectorStepsMixin 2)
    CategoriesMixin(CategoriesMixin 4)
    CalibratorsMixin(CalibratorsMixin 5)
    PlotBase(PlotBase 6)
    ShiftTask(ShiftTask 7)
    SelectorMixin(SelectorMixin 3)
    ConfigTask(ConfigTask 8)

    ConfigTask --> ShiftTask
    ConfigTask --> SelectorMixin
    ConfigTask --> CalibratorsMixin
    ConfigTask --> CategoriesMixin
    ConfigTask --> PlotBase
    SelectorMixin --> SelectorStepsMixin
    SelectorStepsMixin --> PlotCutflowBase
    CategoriesMixin --> PlotCutflowBase
    CalibratorsMixin --> PlotCutflowBase
    PlotBase --> PlotCutflowBase
    ShiftTask --> PlotCutflowBase
```

The numbers indicate the MRO, and since all `resolve_param_values()` methods instantly call their respective `super()` before evaluating their own function bodies, classes with higher numbers are resolved first. And now that `CalibratorsMixin > CategoriesMixin`, we can merge :) Sorry for the delay.

riga (approved):

PR #188, opened 2023-02-20
Support hiding error bars for unstacked processes.

dsavoiu:
This PR implements support for a `draw_errors` flag, which can be set to `False` on a process-by-process basis to prevent error bars being drawn for unstacked processes. The aim is to prevent overcrowded plots when the statistical uncertainty is large.


mafrahm (changes requested):
Good idea! However, it might be nice to change the actual parameter to `skip_error` or so. One reason is consistency (I think all our bool parameters have the default `False`) but it's also shorter to add on command line (`--process-settings proc,draw_errors=False` vs `--process-settings proc,skip_error`).

We could also add this parameter to the other two plot functions `plot_shifted_variable` and `plot_variable_variants` but I can also do that at some point.

dsavoiu:
That's a good point! I would suggest to keep the plural form `errors` though, since there are typically multiple errorbars per plot and it may be confusing to refer to them collectively as *the* error. Also, `skip` seems to be used for the keyword arguments passed to the plotting functions rather than the process properties that control the plotting, so that might again create a bit of confusion.

Maybe we can call it something like `hide_errors` or `no_errors` (or even `no_errorbars`) instead? Let me know what you think.

> We could also add this parameter to the other two plot functions plot_shifted_variable and plot_variable_variants but I can also do that at some point.

Right, I didn't consider the other plot functions, but I think it makes sense to implement this for them as well. I'm happy to add this to the PR once we settle on a name for the setting.

riga:
+1 for `hide-errors`. In case we need the opposite (also for other plot objects that we might want to toggle), `show-*` would be intuitive. For `no-errors`, I wouldn't know what to pick here.

dsavoiu:
Thanks for the feedback! Let's go with `hide_errors` then.

It occurs to me that this could also be used to control showing not just the error bars on unstacked MC processes, but also the error bars for data and the hatched uncertainty band on the MC stack. Moreover, as suggested by @mafrahm, this could also be interesting for tasks other than `PlotVariables1D`, which do not show different processes, but things like shifts or selection steps.

To make this more flexible, I implemented this in two places:
* as a parameter `--hide-errors` in `PlotBase1D`, which controls if errors should be shown in general for any plot task, and 
* as an attribute of `Process` objects, which can be set individually via the `process-settings` to override the general behavior for one particular process (e.g. those drawn as a line outside of the MC stack in `PlotVariables1D`)

Not passing the parameter value (or if the `Process` attribute is missing) yields a default value of `None`, which can be handled according to the need of each task (e.g. for `PlotVariables1D`, the default is to always show the errors).
Passing `True` or `False` will force showing or hiding the errors, with the `Process` attribute taking precedence over the parameter task for unstacked processes.

PR #189, opened 2023-02-22
Fix duplicate entries in `category_ids`.

dsavoiu:
This PR ensures that no duplicate entries are made to the `category_ids` column if multiple instances of the same category are returned by `order.Category.get_leaf_categories`.


mafrahm:
This is a good point and will most likely be relevant at other spots swell (e.g. in plotting tasks), however this is most likely a bug in Order (the output of `get_leaf_categories` should already provide unique categories)

riga:
Indeed, or maybe a residual issue with the combinatorics builder in `config_util`.

riga:
Yup, it was a bug / missing feature in order which is now added in https://github.com/riga/order/commit/2af864bd0b9262620225db002d8c59d9c0758959. I took the liberty to update this feature branch.

PR #198, opened 2023-02-24
New selector for applying data certification filter (golden JSON).

dsavoiu:
This PR implements a selector for filtering events in recorded data based on a JSON-file containing data certification information (a.k.a. the "golden JSON"). This file provides a list of luminosity block ranges per run that are considered valid for physics analysis.

The (vectorized) lookup implementation is based on a sparse boolean matrix (analogous to what is done for cross section lookup when computing the normalization weights). The matrix rows (columns) correspond to the runs (luminosity blocks) covered by a golden JSON file, with `True` entries indicating that a particular (`run`, `luminosityBlock`) combination is certified. The JSON file itself is supplied via the `external_files` config entry.

### DIFF
--- a/columnflow/selection/cms/__init__.py
+++ b/columnflow/selection/cms/__init__.py
@@ -1,0 +1,1 @@
+# coding: utf-8
--- a/columnflow/selection/cms/json_filter.py
+++ b/columnflow/selection/cms/json_filter.py
@@ -1,0 +1,77 @@
+# coding: utf-8
+
+"""
+Selectors for applying golden JSON in data.
+"""
+
+from __future__ import annotations
+
+from columnflow.selection import Selector, selector
+from columnflow.util import maybe_import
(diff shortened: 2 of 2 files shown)


riga (approved):
Nice! Looks all good to me.

Maybe it's a minor thing, but in the CMS context, this type of data filtering is always referred to as "json filtering", right? I thought about suggesting a different name, but then again, if the majority of people in CMS would expect this exact behavior for a selector with that name, we should go with it.

dsavoiu:
Great, thanks for the quick review! Yeah, I think "json filtering" is pretty much the standard way people refer to this in CMS, so I think it's reasonable to keep the selector name as `json_filter`. And the risk of it being misunderstood as something else is probably also pretty low, as I don't think we use JSONs specifically for *filtering* events in another context.

PR #199, opened 2023-02-24
New selector for applying MET filters.

dsavoiu:
This PR adds a selector for applying a configurable subset of MET filters. These are filters designed to reject noisy events with typically large anomalous MET contributions (from detector noise, beam halo effects, etc.). The filter decisions are pre-computed and stored NanoAOD files as part of the `Flag` table.

Since many analyses apply MET filters, it would be good to have something like this centrally provided in `columnflow`.

### DIFF
--- a/columnflow/selection/cms/met_filters.py
+++ b/columnflow/selection/cms/met_filters.py
@@ -46,11 +46,9 @@ def met_filters(
 
     Returns a bool array containing the logical AND of all input columns.
     """
-    met_filters = set(self.config_inst.x.met_filters)
-
     result = ak.ones_like(events.event, dtype=bool)
 
-    for route in met_filters:
+    for route in self.met_filters:
         # interpret column values as booleans
         vals = Route(route).apply(events)
         vals = ak.values_astype(vals, bool)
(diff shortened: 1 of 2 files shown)


columnflow/selection/cms/met_filters.py
```diff
@@ -0,0 +1,71 @@
...
+
+    return result
+
+
+@met_filters.init
+def met_filters_init(self: Selector) -> None:
+    """
+    Read MET filters from config and add them as input columns.
+    """
+    met_filters = set(self.config_inst.x.met_filters)
```
riga, line 70: One thing here: In a previous analysis we had different filters being applied to either data or MC. I think the `cfg.x.met_filters` should be able to be a dictionary with keys `data` and `mc`, and if that is the case, the logic at this point should pick the correct ones based on `self.dataset_inst.data_source`. Suggestion:

```python
met_filters = self.config_inst.x.met_filters
if isinstance(met_filters, dict):
    # do nothing when no dataset_inst is known
    if not getattr(self, "dataset_inst", None):
        return
    met_filters = met_filters[self.dataset_inst.data_source]

# store filters as an attribute for faster lookup
self.met_filters = set(met_filters)

self.uses |= self.met_filters
```

riga (changes requested):

riga:
Very good to have this in a central spot. We have sth similar in place for bbtautau [here](https://github.com/uhh-cms/hh2bbtautau/blob/master/hbt/selection/met.py), but we will for sure transition to this implementation once merged :+1:

dsavoiu:
Ah, missed that, sorry! The implementations seems pretty similar, except this one does an additional cast to bool. And of course doesn't distinguich data/MC, but this should indeed be easily fixed as you suggested above. I'll add that right away.

columnflow/selection/cms/met_filters.py
dsavoiu: Very good point, there needs to be a distinction here. I quite like the suggestion as well since it handles some edge cases quite nicely!

riga (approved):

PR #205, opened 2023-03-01
Allow diverging training phase space

### LINKED ISSUE #201: Different training and evaluation requirements for ML models
At the moment, all requirements resulting from passing `--calibrators X --selector Y --producers Z` to the ML pipeline are considered for both preprocessing/training tasks as well as the evaluation.

From the ML perspective, this enforces that the same phase space is used in both branches, whereas it is sometimes useful (or even required) to have a little divergence (for instance, training in a less noisy subspace).

This is not yet supported, but it clearly should.

riga:
When triggering the MLEvaluation, or anything below that task, the `--prodcers`, `--selector` and `--calibrators` settings (*) are passed upstream, influencing the behavior of all tasks that depend on them.

This, however, is also true for the ML pipeline (PrepareMLEvents → MergeMLEvents → MLTraining) which causes it to always use the exact same phase space (defined by *).

This PR adds additional hooks to the ML interface that allow for re-configuring all settings (*) in order to require a different upstream pipeline, effectively enabling the use of a different phase space w.r.t. the evaluation.

Closes #201.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ However, if you would like to join early on, contribute or just give it a spin, 
 
 ## Quickstart
 
-To create an analysis using columnflow, it is recommended to start from a predefined template (located in [analysis_templates](https://github.com/uhh-cms/columnflow/tree/master/analysis_templates).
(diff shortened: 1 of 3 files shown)


mafrahm (approved):
Looks good!

PR #206, opened 2023-03-01
Add fake weight for real data for cutflow handling.

### LINKED ISSUE #203: Cutflow-related tasks do not work for data
When running on data, the `PlotCutflow` and `PlotCutflowVariables` tasks and their upstream dependencies run into issues.

For example, the `MergeSelectionMasks` task complains about the sum of MC weights not being found.

```
Traceback (most recent call last):
  File "data/software/venvs/venv_columnar_dev/lib/python3.9/site-packages/luigi/worker.py", line 203, in run
    new_deps = self._run_get_new_deps()
  File "data/software/venvs/venv_columnar_dev/lib/python3.9/site-packages/luigi/worker.py", line 147, in _run_get_new_deps
    requires = next(task_gen)
  File "modules/columnflow/modules/law/law/contrib/tasks/__init__.py", line 555, in run
    self.merge(inputs, self.output())
  File "modules/columnflow/columnflow/tasks/selection.py", line 354, in merge
    inputs = self.zip_results_and_columns(inputs, tmp_dir)
  File "modules/columnflow/columnflow/tasks/selection.py", line 365, in zip_results_and_columns
    self.norm_weight_producer.run_setup(
  File "modules/columnflow/columnflow/columnar_util.py", line 1889, in run_setup
    self.setup_func(reqs, inputs)
  File "modules/columnflow/columnflow/production/normalization.py", line 94, in normalization_weights_setup
    for process_id, sum_weights in selection_stats["sum_mc_weight_per_process"].items():
KeyError: 'sum_mc_weight_per_process'
```

The task expects to find the key `sum_mc_weight_per_process` in the `stats` output of `SelectEvents`, but this information is not available for data. Another related issue is that the current implementation of `PlotCutflow` relies on the histogram of `mc_weight`, but this quantity is not produced in data (the corresponding producer is marked `mc_only`).

I think a good way to solve both issues would be to replace `mc_weight` consistently with a generalized `weight` variable, which would be `1.0` for data and the usual generator weight for MC. Then one could make use of the fact that `data` is its own process and store the sum of this generalized `weight` column (i.e. the raw event count) under `sum_weight_per_process` alongside the other MC weights. In this way, both `MergeSelectionMasks` and the cutflow tasks that rely on a histogram of `weight` should work transparently.

riga:
This PR adds a fake weight for real data events in the `MergeSelectionMasks` which is used to prepare values for subsequent cutflow plots.

Fixes #203.

### DIFF
--- a/columnflow/tasks/selection.py
+++ b/columnflow/tasks/selection.py
@@ -17,6 +17,7 @@ from columnflow.production import Producer
 from columnflow.util import maybe_import, ensure_proxy, dev_sandbox, safe_div
 
 
+np = maybe_import("numpy")
 ak = maybe_import("awkward")
 
 
@@ -310,29 +311,39 @@ class MergeSelectionMasks(
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        # store the normalization weight producer
-        self.norm_weight_producer = Producer.get_cls("normalization_weights")(
-            inst_dict=self.get_producer_kwargs(self),
-        )
+        # store the normalization weight producer for MC
+        self.norm_weight_producer = None
+        if self.dataset_inst.is_mc:
+            self.norm_weight_producer = Producer.get_cls("normalization_weights")(
+                inst_dict=self.get_producer_kwargs(self),
+            )
 
     def create_branch_map(self):
         # DatasetTask implements a custom branch map, but we want to use the one in ForestMerge
         return law.tasks.ForestMerge.create_branch_map(self)
 
     def merge_workflow_requires(self):
-        return {
+        reqs = {
             "selection": self.reqs.SelectEvents.req(self, _exclude={"branches"}),
-            "normalization": self.norm_weight_producer.run_requires(),
         }
 
+        if self.dataset_inst.is_mc:
+            reqs["normalization"] = self.norm_weight_producer.run_requires()
+
+        return reqs
+
     def merge_requires(self, start_branch, end_branch):
-        return {
+        reqs = {
             "selection": [
                 self.reqs.SelectEvents.req(self, branch=b)
                 for b in range(start_branch, end_branch)
             ],
-            "normalization": self.norm_weight_producer.run_requires(),
         }
+
+        if self.dataset_inst.is_mc:
+            reqs["normalization"] = self.norm_weight_producer.run_requires()
+
+        return reqs
 
     def trace_merge_workflow_inputs(self, inputs):
         return super().trace_merge_workflow_inputs(inputs["selection"])
@@ -354,16 +365,16 @@ class MergeSelectionMasks(
         law.pyarrow.merge_parquet_task(self, inputs, output)
 
     def zip_results_and_columns(self, inputs, tmp_dir):
-        import awkward as ak
-        from columnflow.columnar_util import RouteFilter, sorted_ak_to_parquet
+        from columnflow.columnar_util import RouteFilter, sorted_ak_to_parquet, set_ak_column
 
         chunks = []
 
         # setup the normalization weights producer
-        self.norm_weight_producer.run_setup(
-            self.requires()["forest_merge"]["normalization"],
-            self.input()["forest_merge"]["normalization"],
-        )
+        if self.dataset_inst.is_mc:
+            self.norm_weight_producer.run_setup(
+                self.requires()["forest_merge"]["normalization"],
+                self.input()["forest_merge"]["normalization"],
+            )
 
         # define columns that will be written
         write_columns = set(self.config_inst.x.keep_columns[self.task_family])
@@ -373,8 +384,11 @@ class MergeSelectionMasks(
             events = inp["columns"].load(formatter="awkward")
             steps = inp["results"].load(formatter="awkward").steps
 
-            # add normalization weight
-            events = self.norm_weight_producer(events)
+            # add normalization weight, create a fake weight for data
+            if self.dataset_inst.is_mc:
+                events = self.norm_weight_producer(events)
+            else:
+                events = set_ak_column(events, "mc_weight", np.ones(len(events), dtype=np.float32))
 
             # remove columns
             events = route_filter(events)


mafrahm (commented):
On a similar note, it is also a bit annoying that you need to add the `mc_weight`variable into your config to run the cutflow tasks, because you can not run e.g. `PlotVariables1D` with data and without the `--variables` parameter because it will try to plot the `mc_weight`, which (usually) does not exist for data. Is there maybe some other dummy variable we could use for cutflow plots which exist for data and mc per default?

dsavoiu (commented):
Thanks for the quick fix @riga. Looks like this solves the issue with `MergeSelectionMasks`. 

Maybe just one thing: the output column is still called `mc_weight` in data, which I find a bit confusing. In #203 I was suggesting to rename it to just `weight` so it applies to both data and MC, do you think this could be implemented here?

I realize that the name change could break existing workflows, but it's probably worthwhile to have in the long run. Of course this can also be handled in a separate issue/PR.

riga:
@mafrahm @dsavoiu I think we can solve both issues (need to pass through `mc_weight` & confusing name `mc_weight` for data) by just using the `event` number. This is the most generic variable and should always be present. We also know it's extent and could use a custom `od.Variable` with reasonable options for @mafrahm 's point.

dsavoiu:
We could for sure add something like a dummy "counter" `od.Variable` like `event`, but at the moment this would then only be used for cutflows, right? Also, what would you propose to take as an extent/binning for the event number? One could define a single bin covering the whole range of the respective integer data type, but this seems a bit hacky.

One should also be aware that there is a difference between using `mc_weight` and `event` for the MC cutflow b/c the latter only tallies the raw event numbers.

dsavoiu:
BTW, I noticed the MC-only `normalization_weight` was still being used for data in `CreateCutflowHistograms` and it seemed an easy enough fix, so I added it to this PR [here](https://github.com/uhh-cms/columnflow/pull/206/commits/6bd3b764f87910f482b0b987d693b72c0f201188).

riga:
> BTW, I noticed the MC-only normalization_weight was still being used for data in CreateCutflowHistograms and it seemed an easy enough fix, so I added it to this PR [here](https://github.com/uhh-cms/columnflow/pull/206/commits/6bd3b764f87910f482b0b987d693b72c0f201188).

Added almost the same and merged now :)

> We could for sure add something like a dummy "counter" od.Variable like event, but at the moment this would then only be used for cutflows, right? Also, what would you propose to take as an extent/binning for the event number? One could define a single bin covering the whole range of the respective integer data type, but this seems a bit hacky.

The variable binning itself is actually not needed if I see correctly, but it's rather important to just *have* the variable present in the config.
- I think at first we wanted to pick `mc_weight` as a placeholder since it could also be used for (e.g.) normalization checks. However, now we have the full normalization weight for that, so this seems a bit redundant.
- Thinking about this again, having the event number being saved as a column for the cutflow pipeline could indeed be a good thing to have, for instance to check if a certain event (number) passes some selection steps or not.

I just tried (and pushed) the implementation with that, but if you think we could use a different "counter" variable we could also include that instead :)

While checking the plotting I also stumbled upon a couple minor issues in the plot methods. I added comments below so that we can keep track of them, but imho they could be solved in a different PR.

columnflow/plotting/plot_util.py
```diff
@@ -172,7 +174,8 @@ def prepare_style_config(
 
     # disable minor ticks based on variable_inst
     if variable_inst.discrete_x:
-        style_config["ax_cfg"]["xticks"] = range(int(xlim[0]), int(xlim[1]) + 1)
```
riga, line 175: I had a variable with binning `(100, 0.0, 1.0e10)` which caused `range` to create 10B integers in memory. Changed this to `bin_edges` for now, but this isn't optimal either.

columnflow/plotting/plot_util.py
```diff
@@ -279,12 +282,12 @@ def prepare_plot_config(
             "kwargs": {
                 "norm": data_norm,
                 "label": "Data",
-                "yerr": False if data_hide_errors[i] else None,
-            },
-            "ratio_kwargs": {
-                "norm": h_mc.values() * data_norm / mc_norm,
-                "yerr": False if data_hide_errors[i] else None,
```
riga, line 286: In this block, neither `i` nor `h_mc` are necessarily existing.

columnflow/plotting/plot_util.py
```diff
@@ -279,12 +282,12 @@ def prepare_plot_config(
...
+                "yerr": False if data_hide_errors[0] else None,
```
mafrahm, line 285: How about "skip year als long as some data process requested it?"
```suggestion
                "yerr": False if any(data_hide_errors) else None,
```

columnflow/plotting/plot_util.py
```diff
@@ -279,12 +282,12 @@ def prepare_plot_config(
...
             },
+            # "ratio_kwargs": {
+            #     "norm": h_mc.values() * data_norm / mc_norm,
```
mafrahm, line 288: the "norm" parameter is quite important for the ratio plot, so we should included it back rather sooner than later. Otherwise, there will be no visible data points in the data/mc ratio. Suggestion, which adds it back as long as some stack is present.
```suggestion
            if h_mc:
                plot_config["ratio_kwargs"]["norm"] = h_mc.values() * data_norm / mc_norm
```

columnflow/tasks/framework/mixins.py
```diff
@@ -254,9 +254,9 @@ def store_parts(self):
 class SelectorStepsMixin(SelectorMixin):
 
     selector_steps = law.CSVParameter(
-        default=(),
+        default=("default",),
         description="a subset of steps of the selector to apply; uses all steps when empty; "
-        "empty default",
+        "default: (default,)",
```
mafrahm, line 259: does it really make sense to add a default here? The empty default also uses the "main" selection when calling ReduceEvents

mafrahm: Also, I think we wantd to use the `selector-steps` parameter after the reduction to perform an additional last-minute selection before e.g. plotting, right?

riga: I forgot about that, you're right! Reverted this one.

columnflow/plotting/plot_functions_1d.py
```diff
@@ -145,12 +144,12 @@ def plot_shifted_variable(
...
     **kwargs,
 ) -> plt.Figure:
-
-    hists = apply_variable_settings(hists, variable_insts, variable_settings)
+    """
+    TODO.
+    """
     variable_inst = variable_insts[0]
-
+    hists = apply_variable_settings(hists, variable_insts, variable_settings)
```
mafrahm, line 151: `variable_inst = variable_insts[0]` just creates a pointer, right? Otherwise, the order here would be important since we also want to change `variable_inst`with the `apply_variable_settings` function.

mafrahm (approved):
Just one last inline comment. Otherwise, the PR can be merged from my point of view.

columnflow/plotting/plot_util.py
riga: Resolving this for now, should be followed up on in #207.

columnflow/plotting/plot_functions_1d.py
riga: I see. Changed this :+1:

columnflow/tasks/cutflow.py
```diff
@@ -338,7 +342,7 @@ def run(self):
                     }]
 
                     # axis reductions
-                    h = h[{"process": sum, "category": sum, "mc_weight": sum}]
+                    h = h[{"process": sum, "category": sum, "event": sum}]
```
dsavoiu, line 345: Does the sum over `event` include potential `flow` entries? Likely only a problem if the range of the `event` `od.Variable` is too small (see my other comment on the binning).

columnflow/tasks/selection.py
```diff
@@ -354,27 +365,28 @@ def merge(self, inputs, output):
...
-        # define columns that will be written
-        write_columns = set(self.config_inst.x.keep_columns[self.task_family])
+        if self.dataset_inst.is_mc:
+            self.norm_weight_producer.run_setup(
+                self.requires()["forest_merge"]["normalization"],
+                self.input()["forest_merge"]["normalization"],
+            )
+
+        # define columns that will be written, including the event number as required for cutflows
+        write_columns = set(self.config_inst.x.keep_columns[self.task_family]) | {"event"}
```
dsavoiu, line 380: AFAIK event IDs are not unique, so usually one needs the full set of `run`, `luminosityBlock` and `event` to unambiguously identify an event. Should we maybe write out all three here?

columnflow/plotting/plot_util.py
```diff
@@ -279,13 +282,14 @@ def prepare_plot_config(
...
                 "label": "Data",
-                "yerr": False if data_hide_errors[i] else None,
-            },
-            "ratio_kwargs": {
-                "norm": h_mc.values() * data_norm / mc_norm,
-                "yerr": False if data_hide_errors[i] else None,
+                "yerr": False if any(data_hide_errors) else None,
             },
         }
+        if h_mc:
```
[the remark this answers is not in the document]
riga: True. Added that 👍

analysis_templates/cms_minimal/__cf_module_name__/config/analysis___cf_short_name_lc__.py
```diff
@@ -261,6 +260,14 @@
 )
 
 # add variables
+# (the "event" variable is required for some cutflow plotting task)
+cfg.add_variable(
+    name="event",
+    expression="event",
+    binning=(100, 0.0, 1.0e9),
```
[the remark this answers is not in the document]
riga: True. I used a single bin for now.

Maybe we can continue the discussion on the choice of the counter variable(s) in https://github.com/uhh-cms/columnflow/pull/206#discussion_r1123606712 below :)

columnflow/tasks/cutflow.py
dsavoiu: Great, thanks for checking! I also think it's probably better to include `flow` bins in `sum` reductions, but it's good to be aware of this behavior in case this is ever not wanted.

columnflow/tasks/selection.py
riga: Yes, we should actually include all for unambiguous event book keeping during selection synchronizations. I added it and included a parameter for the choice of the counter variable.

dsavoiu: Nice! Then I guess this PR is good to go.

dsavoiu (approved):

PR #209, opened 2023-03-06
Fix empty setup mode for remote jobs.

dsavoiu:
This PR proposes a fix for the install mode check for remote jobs. Following the changes introduced by [`75f013`](https://github.com/uhh-cms/columnflow/commit/75f0136527ece4fcb4ec5d6001ea039b17744cea), the empty `mode` is being resolved before checking if it is empty for remote jobs, which always leads to an error.


riga (approved):
True! Needs to go in right away. Thank you for fixing that.

PR #212, opened 2023-03-10
Ensure unique sandbox install paths.

### LINKED ISSUE #211: Handle sandbox name collisions
Right now, sandboxes (venv, cmssw) are installed in directories that only contain the name of the respective sandbox. However, when there are identically named sandboxes, for instance defined once in CF and once in the analysis, they currently use the same install directory, which we probably do not want.

riga:
This PR fixes a naming collision between different sandboxes that would currently be installed into the same directory.

Right now, sandboxes are stored e.g. in `$CF_(VENV|CMSSW)_BASE/<sandbox_name>/`, where the `sandbox_name` can be defined by the sandbox itself. However, when there are multiple sandboxes with the same name, the install paths are identical, but they clearly shouldn't.

This PR changes the install paths to `$CF_(VENV|CMSSW)_BASE/<sandbox_name>_<sandbox_hash>/` where the hash is a unique identifier based on the location of the sandbox file. There are two complications that render this PR rather large:

- The shell environment (in which sandboxes are built) and python tools (where bundling for remote jobs takes place) have to agree on the hash definition.
- The hash is based on the sandbox file location, but one can't use the absolute path here. Instead, names of environment variables must be considered (e.g. `\$CF_BASE/sandboxes/venv_columnar.sh` or `\$MY_ANALYSIS_BASE/sandboxes/my_venv.sh`) so that the local env and remote job envs also agree on the same hashes.

Fixes #211.

### DIFF
--- a/columnflow/tasks/framework/remote.py
+++ b/columnflow/tasks/framework/remote.py
@@ -48,8 +48,7 @@ class BundleRepo(AnalysisTask, law.git.BundleGitRepository, law.tasks.TransferLo
         self.bundle(bundle)
 
         # log the size
-        size, unit = law.util.human_bytes(bundle.stat().st_size)
-        self.publish_message(f"size is {size:.2f} {unit}")
+        self.publish_message(f"size is {law.util.human_bytes(bundle.stat().st_size, fmt=True)}")
 
         # transfer the bundle
         self.transfer(bundle)
@@ -151,9 +150,12 @@ class BundleBashSandbox(AnalysisTask, law.tasks.TransferLocalFile):
         super().__init__(*args, **kwargs)
 
         # get the name and install path of the sandbox
+        from cf_sandbox_file_hash import create_sandbox_file_hash
         sandbox_file = os.path.expandvars(os.path.expanduser(self.sandbox_file))
+        self.sandbox_file_hash = create_sandbox_file_hash(sandbox_file)
         self.venv_name = os.path.splitext(os.path.basename(sandbox_file))[0]
-        self.venv_path = os.path.join(os.environ["CF_VENV_BASE"], self.venv_name)
+        self.venv_name_hashed = f"{self.venv_name}_{self.sandbox_file_hash}"
+        self.venv_path = os.path.join(os.environ["CF_VENV_BASE"], self.venv_name_hashed)
 
         # checksum cache
         self._checksum = None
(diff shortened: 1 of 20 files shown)


columnflow/tasks/framework/remote.py
```diff
@@ -227,38 +228,41 @@ class BundleCMSSWSandbox(AnalysisTask, law.cms.BundleCMSSW, law.tasks.TransferLo
...
         return self.requires().sandbox_inst.env["CMSSW_BASE"]
 
-    def get_file_pattern(self):
-        path = os.path.expandvars(os.path.expanduser(self.single_output().path))
-        return self.get_replicated_path(path, i=None if self.replicas <= 0 else r"[^\.]+")
-
     def single_output(self):
         cmssw_path = os.path.basename(self.get_cmssw_path())
-        return self.target(f"{cmssw_path}.{self.checksum}.tgz")
+        return self.target(f"{self.cmssw_env_name_hashed}_{cmssw_path}.{self.checksum}.tgz")
```
dsavoiu, line 249: Given that the tarball name now includes the hash, could the `cmsw_path` part actually be dropped here?

dsavoiu (approved):
This PR goes into quite some depth, but in general I see the need for the added complexity. It's definitely useful to have unique paths for different-but-identically-named sandboxes, although I'm not entirely sure what the use-case is for this (maybe this can be discussed a bit on #211).

Codewise I don't see any obvious issues. The idea to canonicalize in-repo paths to start with an env-var like `\$CF_BASE` or the repo alias, and to do the hash calculation on that string via a shell-executable Python script looks like a robust solution to me!

I only have one minor comments below, but otherwise I think this can be merged.

columnflow/tasks/framework/remote.py
riga: The installation directory for cmssw is e.g. `cmssw_default_b6e834cc/CMSSW_12_6_2`, driven by the way `cmsrel` checks out releases.Aand it could be that the same sandbox file (same hash) switches the CMSSW version. This would also need to be covered by the target path above (not 100% cleam, but cmssw compatible) :+1:

riga:
Thanks for checking this PR in such depth!
I think we can merge then :)

columnflow/tasks/framework/remote.py
dsavoiu: Ah, I see, thanks for the explanation! Then of course it makes sense to keep the CMSSW version as part of the tarball name 👍

PR #214, opened 2023-03-16
Fix default linting test config

danielhundhausen:


analysis_templates/cms_minimal/tests/run_linting
```diff
@@ -10,7 +10,7 @@ action() {
 
     (
         cd "${__cf_short_name_lc___dir}" && \
-        flake8 __cf_module_name__ tests
+        flake8 --config modules/columnflow/.flake8 __cf_module_name__ tests
```
riga, line 13: There's actually a flake8 config file shipped with the template in [`__cf_short_name_lc___dir`](https://github.com/uhh-cms/columnflow/blob/master/analysis_templates/cms_minimal/.flake8), so we probably shouldn't forward to the rules in cf itself.

Or did sth not work as expected?

riga (changes requested):
Thanks for fixing the link! 👍

Just one question related to the flake8 config inline.

analysis_templates/cms_minimal/tests/run_linting
danielhundhausen: Ah, okay. If I understand correctly, the file you pointed out is supposed to be copied [here](https://github.com/uhh-cms/columnflow/blob/6bd7017d18d79251ad7248f2ae5b39f8803a127b/create_analysis.sh#L179) in debug mode and otherwise moved from the downloaded tarball in the else blcok below.
I think this fails because apparently `*` does not "see" the hidden files both when used with `cp` as well as `mv`.

```
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/Foonalysis% cd ..
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad% mkdir foo
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad% cd foo 
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/foo% mkdir bar
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/foo% touch bar/file1
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/foo% touch bar/.file2
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/foo% cp -r bar/* .
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/foo% ll
total 2
drwxr-xr-x  3 hundhad af-cms 4096 Mar 17 08:51 .
drwxr-x--- 23 hundhad af-cms 4096 Mar 17 08:51 ..
drwxr-xr-x  2 hundhad af-cms 4096 Mar 17 08:51 bar
-rw-r--r--  1 hundhad af-cms    0 Mar 17 08:51 file1
```

But [this](https://superuser.com/questions/61611/how-to-copy-with-cp-to-include-hidden-files-and-hidden-directories-and-their-con) should work for copying.
```
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/foo% cp -r bar/. .
[hundhad@naf-cms12]/nfs/dust/cms/user/hundhad/foo% ll
total 2
drwxr-xr-x  3 hundhad af-cms 4096 Mar 17 08:57 .
drwxr-x--- 23 hundhad af-cms 4096 Mar 17 08:51 ..
-rw-r--r--  1 hundhad af-cms    0 Mar 17 08:57 .file2
drwxr-xr-x  2 hundhad af-cms 4096 Mar 17 08:51 bar
-rw-r--r--  1 hundhad af-cms    0 Mar 17 08:57 file1
```

And replacing `*` with `{.[!.],}*` seems to fix `mv`.

riga: Since this is a zsh-bash thing, I opened https://github.com/danielhundhausen/columnflow/pull/1 .

riga (approved):

PR #217, opened 2023-03-29
Feature/array function timing

### LINKED ISSUE #215: Automatic runtime timing of TaskArrayFunctions
There should be a quick and easy way to enable timing all individual TaskArrayFunctions that are called as part of a Selector/Producer/Calibrator

mafrahm:
Closes #215


riga:
I move the functionality up to the general `ArrayFunction` as this could also be a helpful feature elsewhere :+1:

I also added an option to configure the runtime logging on a per-object basis and used `time.perfcounter()` for the exact measurement.

riga (approved):

PR #218, opened 2023-03-29
Speed up gen top decay producer.

riga:
This PR updates the `gen_top_decay_products` producer and speeds it up significantly. The change makes use of the new "distinctChildrenDeep" feature, which does exactly what we want when looking for decay products to collect distinct child particles.

In addition, the default `skip_func` definitions for producers/calibrators/selectors received syntax updates to make it easier to debug the skip decision process.


mafrahm (approved):
Looks good, the `distinctChildrenDeep` makes that way easier :+1:

Issue #196, opened 2023-02-23
Add tasks to create yield tables


riga:
We need a task that reads the histograms from either `MergeHistograms` or `MergeShiftedHistograms` and produces yield tables.

Development already started in [feature/yield_tables](https://github.com/uhh-cms/columnflow/tree/feature/yield_tables).

riga:
Added in #219.

PR #219, opened 2023-03-30
Implementation of CreateYieldTable task

mafrahm:
Example task:
```
law run cf.CreateYieldTable --version v1 --processes tt_sl,st_tchannel_t --categories incl,1mu
```
Parameters that are already implemented are `table_format` (changing the tabulate table style), `number_format` (changing the scinum.number string format) and `skip_uncert` (to not display the uncertainties in the table).

An option to normalise entries in the table is still missing.

### DIFF
--- a/columnflow/tasks/yields.py
+++ b/columnflow/tasks/yields.py
@@ -4,9 +4,8 @@
 Tasks to produce yield tables
 """
 
-from collections import defaultdict, OrderedDict
-
 import math
+from collections import defaultdict, OrderedDict
 
 import law
 import luigi
@@ -18,7 +17,6 @@ from columnflow.tasks.framework.mixins import (
     DatasetsProcessesMixin, CategoriesMixin,
 )
 from columnflow.tasks.framework.remote import RemoteWorkflow
-
 from columnflow.tasks.histograms import MergeHistograms
 from columnflow.util import dev_sandbox, test_int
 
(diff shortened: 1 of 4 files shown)


riga:
Closes #196.

columnflow/tasks/yields.py
```diff
@@ -0,0 +1,191 @@
...
+    CategoriesMixin,
+    ProducersMixin,
+    SelectorStepsMixin,
+    CalibratorsMixin,
+    law.LocalWorkflow,
+    RemoteWorkflow,
+):
+    sandbox = dev_sandbox("bash::$CF_BASE/sandboxes/venv_columnar.sh")
+
+    dep_MergeHistograms = MergeHistograms
```
riga, line 37: That one can be removed I think.

columnflow/tasks/yields.py
```diff
@@ -0,0 +1,191 @@
...
+    number_format = luigi.Parameter(
+        default="pdg",
+        significant=False,
+        description=(
+            "format of each number in the yield table, takes all formats taken by "
+            "Number.str, e.g. 'pdg', 'publication' or '%.1f'; default: pdg"
+        ),
+    )
+
+    skip_uncert = luigi.BoolParameter(
```
riga, line 60: I think it's ok not to abbreviate this, i.e., `skip_uncertainties`.

columnflow/tasks/yields.py
```diff
@@ -0,0 +1,191 @@
...
+                        )
+
+                    row.append(value_str)
+
+                yields.append(row)
+
+            yield_table = tabulate(yields, headers=yield_header, tablefmt=self.table_format)
+            print(yield_table)
+
+            self.output().dump(yield_table)
```
riga, line 191: Better to use the explicit formatter, i.e., `self.output().dump(yield_table, formatter="text")`

columnflow/tasks/yields.py
```diff
@@ -0,0 +1,191 @@
...
+                            format=self.number_format,
+                            style="latex" if "latex" in self.table_format else "plain",
+                        )
+
+                    row.append(value_str)
+
+                yields.append(row)
+
+            yield_table = tabulate(yields, headers=yield_header, tablefmt=self.table_format)
+            print(yield_table)
```
riga, line 189: `self.publish_message(yield_table)`

-> this prints the table, but also forwards it to any connected central scheduler so that one can see the table in the web interface.

columnflow/tasks/yields.py
```diff
@@ -0,0 +1,191 @@
...
+                    h_cat = h[{"category": [
+                        hist.loc(c.id)
+                        for c in leaf_category_insts
+                        if c.id in h.axes["category"]
+                    ]}]
+                    h_cat = h_cat[{"category": sum}]
+
+                    value = Number(h_cat.value, math.sqrt(h_cat.variance))
+
+                    # TODO: allow normalizing per process or per category (or both?)
```
riga, line 175: both :)

columnflow/tasks/yields.py
```diff
@@ -0,0 +1,191 @@
...
+                    if self.skip_uncert:
+                        value_str = value.str(format=self.number_format).split(" +- ")[0]
```
riga, line 177: In case the uncertainty as asymmetric - for some reason that should actually not apply here - there won't be a `+-`.

Maybe go for

```python
value_str = Number(value.n).str(format=self.number_format).split(" ", 1)[0]
```

riga (changes requested):
Nice! Looks very good 🎉

Just a couple minor things inline.

riga (approved):

PR #220, opened 2023-04-03
Add category id creation to config_util.

### LINKED ISSUE #192: Add config_util for automatic, deterministic category id generation
With new features such as the combinatorial category building, we should also have a util that creates ids for categories (and maybe also other objects) in an automatic yet deterministic fashion.

Usually, one is not interested in the id themselves, but just in the fact that they are and remain unique.

riga:
This PR adds a small config_util to auto-generate deterministic category ids.

They are ensured to remain unique within the config they are added to using a hash created from config id, config name and category name.

Closes #192.

### DIFF
--- a/columnflow/config_util.py
+++ b/columnflow/config_util.py
@@ -8,13 +8,14 @@ from __future__ import annotations
 
 __all__ = [
     "get_root_processes_from_campaign", "add_shift_aliases", "get_shifts_from_sources",
-    "create_category_combinations",
+    "create_category_id", "add_category", "create_category_combinations",
 ]
 
 import re
 import itertools
 from typing import Callable, Any
 
+import law
 import order as od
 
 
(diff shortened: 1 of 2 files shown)


columnflow/config_util.py
```diff
@@ -83,6 +84,36 @@ def get_shifts_from_sources(config: od.Config, *shift_sources: str) -> list[od.S
...
+    category_name: str,
+    hash_len: int = 7,
+) -> int:
+    """
+    Creates a unique id for a :py:class:`order.Category` named *category_name* in a
+    :py:class:`order.Config` object *config* and returns it. Internally,
+    :py:func:`law.util.create_hash` is used which receives *hash_len*.
+    """
+    h = law.util.create_hash((config.name, config.id, category_name), l=hash_len)
+    return int(h, base=16)
```
mafrahm, line 98: would be nice if we could add an offset here (of 10^8?) to avoid overlap with category IDs manually set by the user

mafrahm (changes requested):
Looks good to me, just one inline comment

columnflow/config_util.py
riga: Yup, added that, plus the option to add a custom *salt* value.

columnflow/config_util.py
```diff
@@ -83,6 +84,45 @@ def get_shifts_from_sources(config: od.Config, *shift_sources: str) -> list[od.S
...
+    :py:func:`law.util.create_hash` is used which receives *hash_len*. In case of an unintentional
+    (yet unlikely) collision of two ids, there is the option to add a custom *salt* value.
+    """
+    # create the hash
+    h = law.util.create_hash((config.name, config.id, category_name, salt), l=hash_len)
+    h = int(h, base=16)
+
+    # take the last number and add it to the front to ensure that are hashes are above a threshold
+    digits = len(str(int("F" * hash_len, base=16)))
+    h += max(h % 10, 1) * 10 ** digits
```
mafrahm, line 105: I'm a bit confused by the offset you chose here. Why do you need different offsets based on the last digit of the category?
I'd simply scale each category by the same fixed value, e.g.
```suggestion
    h += 10 ** (digits + 1)
```

riga: True, we can do that too (the `+ 1` should be dropped though).

mafrahm: Yeah, skipping the `+ 1` is fine with me

Issue #223, opened 2023-04-05
Task outputs as dict


mafrahm:
The output of all tasks should be dictionaries to allow the user more flexibility when replacing tasks or redefining task outputs.

PR #224, opened 2023-04-05
change task outputs to dictionaries for calibration/selection/reduction

mafrahm:
As mentioned in #223, it might be nice to always return dictionaries as task outputs.
I implemented these changes for all tasks except external tasks (CreatePileupWeights, GetDatasetLFNS, etc.).

From the user side, there might be changes required in some producers that require tasks on their own (e.g. stats in weight renormalisation producers) and in the ML Model

### DIFF
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -142,7 +142,7 @@ class CreateCutflowHistograms(
                     histograms[var_key] = h.Weight()
 
         for arr, pos in self.iter_chunked_io(
-            inputs["masks"].path,
+            inputs["masks"]["masks"].path,
             source_type="awkward_parquet",
             read_columns=load_columns,
         ):
@@ -298,10 +298,10 @@ class PlotCutflow(
         }
 
     def output(self):
-        return [
+        return {"plots": [
             self.target(name)
             for name in self.get_plot_names(f"cutflow__cat_{self.branch_data}")
-        ]
+        ]}
 
     @law.decorator.log
     @view_output_plots
@@ -382,7 +382,7 @@ class PlotCutflow(
             )
 
             # save the plot
-            for outp in self.output():
+            for outp in self.output()["plots"]:
                 outp.dump(fig, formatter="mpl")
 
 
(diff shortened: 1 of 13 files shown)


columnflow/tasks/cutflow.py
```diff
@@ -142,7 +142,7 @@ def prepare_hists(steps):
                     histograms[var_key] = h.Weight()
 
         for arr, pos in self.iter_chunked_io(
-            inputs["masks"].path,
+            inputs["masks"]["masks"].path,
```
riga, line 145: Should the first level be "selection"?

columnflow/tasks/reduction.py
```diff
@@ -452,12 +454,12 @@ def merge_output(self):
         # use the branch_map defined in DatasetTask to compute the number of files after merging
         n_merged = len(DatasetTask.create_branch_map(self))
         return law.SiblingFileCollection([
-            self.target(f"events_{i}.parquet")
+            {"events": self.target(f"events_{i}.parquet")}
```
riga, line 457: I think I'd expect this rather to be

```python
return {
    "events": law.SiblingFileCollection([
        self.target(f"events_{i}.parquet")
        for i in range(n_merged)
    ]),
}
```

columnflow/tasks/ml.py
```diff
@@ -462,7 +462,7 @@ def requires(self):
 
     @MergeReducedEventsUser.maybe_dummy
     def output(self):
-        return self.target(f"mlcols_{self.branch}.pickle")
+        return {"mlcols": self.target(f"mlcols_{self.branch}.pickle")}
```
riga, line 465: For consistency with "events", "mlevents", and "columns", we should maybe call this "mlcolumns" (same then in the other places below).

columnflow/tasks/yields.py
```diff
@@ -222,7 +230,7 @@ def run(self):
 
                     # format yields into strings
                     yields_str[category].append((value / norm_factor).str(
-                        combine_uncs="all",
+                        # combine_uncs="all",  # TODO
```
riga, line 233: intended?

riga (changes requested):
Yup, this seems like a lot of work ^^ Nice.

Just two renaming suggestions and a minor structural thing.

columnflow/tasks/ml.py
mafrahm: Yeah, I agree. Should we then also change the file name to "mlcolumns_{branch}.pickle"?

columnflow/tasks/yields.py
mafrahm: Yes, there was an error at that line and I wasn't motivated enough to find out what the issue is.

riga: :+1: Could you open an issue referring to this line, just so that we can follow up.

columnflow/tasks/reduction.py
riga: Ok, sorry, my bad. The `ForestMerge` requires a collection output.

columnflow/tasks/yields.py
mafrahm: Opened! #229

riga (approved):

PR #225, opened 2023-04-05
fix attachment of coffea behaviors

pkausw:
For many analyses, the usage of four momenta for the calculation of variables/observables is critical. columnflow utilizes awkward array behaviors from `coffea`, which provide some very nice convenience functions (e.g. `metric_table` and `sum`) and also include the definition of operators like `+` etc. for four momenta.

In the current version of cf, this mechanism is broken in any step after the `ReduceEvents` task. It seems like the awkward arrays loose their intrinsic behavior somewhere along the way, which leads to a crash in `columnar_util.attach_behavior`.

This PR restores the behavior by providing the original `coffea` nanoaod behavior as an alternative in case the current awkward array does not have a behavior already. This fix is quite minimal, but restores full functionality for four momenta (tested with `Jet` collection).

Side note: Is there still a reason that the coffea version in the `columnar.txt` requirement is fixed to a private fork of `coffea`? I've had a (very brief) look through the history of the `awkward_v2_dev` branch of `coffea` and had the impression that the fix was already included there.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -42,6 +42,7 @@ uproot = maybe_import("uproot")
 coffea = maybe_import("coffea")
 maybe_import("coffea.nanoevents")
 maybe_import("coffea.nanoevents.methods.base")
+coffea_nanoaod = maybe_import("coffea.nanoevents.methods.nanoaod")
 pq = maybe_import("pyarrow.parquet")
 
 logger = law.logger.get_logger(__name__)
@@ -993,8 +994,9 @@ def attach_behavior(
     By default, all subfields of *ak_array* are kept. For further control, *skip_fields* can contain
     names or name patterns of fields that are filtered.
     """
+
     if behavior is None:
-        behavior = getattr(ak_array, "behavior", None)
+        behavior = getattr(ak_array, "behavior") or coffea_nanoaod.behavior
         if behavior is None:
             raise ValueError(
                 f"behavior for type '{type_name}' is not set and not existing in input array",


mafrahm (approved):
Nice!

columnflow/columnar_util.py
```diff
@@ -42,6 +42,7 @@
 coffea = maybe_import("coffea")
 maybe_import("coffea.nanoevents")
 maybe_import("coffea.nanoevents.methods.base")
+coffea_nanoaod = maybe_import("coffea.nanoevents.methods.nanoaod")
```
riga, line 45:
```suggestion
maybe_import("coffea.nanoevents.methods.nanoaod")
```

columnflow/columnar_util.py
```diff
@@ -993,8 +994,9 @@ def attach_behavior(
     By default, all subfields of *ak_array* are kept. For further control, *skip_fields* can contain
     names or name patterns of fields that are filtered.
     """
+
     if behavior is None:
-        behavior = getattr(ak_array, "behavior", None)
+        behavior = getattr(ak_array, "behavior") or coffea_nanoaod.behavior
```
riga, line 999:
```suggestion
        behavior = getattr(ak_array, "behavior", None) or coffea.nanoevents.methods.nanoaod.behavior
```

riga (changes requested):
Looks nice! Just one minor thing regarding a peculiarity of `getattr` and a subjective :) preference to keep the fill import location.

Otherwise, we can (and should) merge right away!

columnflow/columnar_util.py
riga: (just in case the behavior attribute is not even there instead of being None ... who knows)

pkausw: I actually tried this and it is indeed `None`, but I agree better safe than sorry :D

pkausw:
Thanks for the reviews @mafrahm @riga ! I implemented the suggestions as they were, this would be ready for merging from my side :+1:

@riga Do you have any comments regarding the side note in the original message above?

riga:
Ah right, I forgot :) The coffea guys are heavily pushing the 2023 release and they use a rebase/force-push combination that overwrites they git hashes every once in a while. We updated the commit references in columnar.txt about 4 or 5 times and then I figured we could use a "fixed" location until the actual release is there :)

riga (approved):

PR #232, opened 2023-04-13
Class aware decorators for task array functions

### LINKED ISSUE #230: Class-awareness of calibrator, selector and producer decorators
The default `@calibrator`, `@selector`, `@producer` decorators are hard-wired to the base `Calibrator`, `Selector` and `Producer` types which they use to create and return sub-classes.

To allow easier subclassing, the decorators should be class-properties which would automatically make then class-aware, i.e., doing 

```python
@MyCustomBaseProducer.producer
def my_function(...)
```

would create a new producer class `my_function`, but derived from `MyCustomBaseProducer`.

riga:
This PR makes the calibrator / selector / producer (CSP) decorators classmethods of the respective `Calibrator`, `Selector` and `Producer` base classes.

This does not change existing code and forces to updates to analyses, but it allows transparent subclassing. The PR also changes the two cms-related scale weight producers which actually share some attributes (i.e. the named indices of the weights). This is done by defining a common, unexposed subclass `_ScaleWeightBase` and then dynamically created subclasses via (e.g.)

```python
@_ScaleWeightBase.producer()
def ...
```

Closes #230.

### DIFF
--- a/columnflow/calibration/__init__.py
+++ b/columnflow/calibration/__init__.py
@@ -13,56 +13,63 @@ from columnflow.util import DerivableMeta
 from columnflow.columnar_util import TaskArrayFunction
 
 
-Calibrator = TaskArrayFunction.derive("Calibrator")
+class Calibrator(TaskArrayFunction):
+
+    @classmethod
+    def calibrator(
+        cls,
+        func: Callable | None = None,
+        bases=(),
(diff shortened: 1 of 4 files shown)


mafrahm (approved):
Looks good to me!

PR #233, opened 2023-04-17
Feature/allow histogramming jagged variables

mafrahm:
This PR allows creating histograms with jagged variables, e.g.
```
cfg.add_variable(
    name="jets_pt",
    expression="Jet.pt",
    binning=(40, 0.0, 400.0),
    unit="GeV",
    x_title=r"$p_{T} of all jets$",
)
```
creates a single histogram that contains the pt of all jets.

### DIFF
--- a/columnflow/tasks/histograms.py
+++ b/columnflow/tasks/histograms.py
@@ -185,7 +185,7 @@ class CreateHistograms(
                 fill_kwargs = {
                     "category": events.category_ids,
                     "process": events.process_id,
-                    "shift": self.global_shift_inst.id,
+                    "shift": ak.Array(np.zeros(len(events)) + self.global_shift_inst.id),
                     "weight": weight,
                 }
                 for variable_inst in variable_insts:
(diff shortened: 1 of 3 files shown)


columnflow/tasks/histograms.py
```diff
@@ -199,9 +199,10 @@ def expr(events, *args, **kwargs):
                             return route.apply(events, null_value=variable_inst.null_value)
                     # apply it
                     fill_kwargs[variable_inst.name] = expr(events)
+
                 # broadcast and fill
```
riga, line 203: Could you adjust the comment maybe? :)

columnflow/tasks/histograms.py
```diff
@@ -185,7 +185,7 @@ def run(self):
                 fill_kwargs = {
                     "category": events.category_ids,
                     "process": events.process_id,
-                    "shift": self.global_shift_inst.id,
+                    "shift": ak.Array(np.zeros(len(events)) + self.global_shift_inst.id),
```
riga, line 188: Is the `ak.Array` needed here? If not, then what is typically done is

```python
np.ones(len(events), dtype=np.int32) * self.global_shift_inst.id
```

Esp. the dtype could be needed here as `zeros` could default to `float64`.

riga (changes requested):
Just two minor things inline and a question:

Would this also be required for histogramming cutflow variables in the other task chain?

columnflow/tasks/histograms.py
mafrahm: changed :+1:

mafrahm:
> Would this also be required for histogramming cutflow variables in the other task chain?

Good point, it's implemented there aswell now

pkausw:
@mafrahm can you fix the linting issue here? I think we could then proceed with approving this PR

mafrahm:
> @mafrahm can you fix the linting issue here? I think we could then proceed with approving this PR

done!

columnflow/tasks/histograms.py
riga: Ok, after looking at the [docs](https://awkward-array.org/doc/main/reference/generated/ak.cartesian.html), "broadcast" still makes sense here.

riga (approved):

Issue #234, opened 2023-04-26
missing information in MLTraining outputs paths


mafrahm:
At the moment, only the version parameter is encoded in the `MLTraining` output path.

We should add the used ml_model and configs to the `store_parts` and probably even the used calibrators/selectors/producers

dsavoiu:
This is partially addressed in PR #237, which adds a config identifier and the `ml_model` to the output path. Adding the the calibrators/selectors/producers would require some refactoring, one reason being that `MLTraining` accepts multiple `configs` and so cannot inherit from the calibrator/selector/producer mixins, as these explicitly require a single `config`.

Also to note is that in PR #237 the `configs` string representation directly joins together the used-supplied strings. To prevent overly long output paths, and for consistency with other places where potentially long lists of strings are encoded in the output, this could be changed to use only the first config string, followed by the total number of configs and a unique hash.

mafrahm:
I just checked the different implementations of `store_parts`. For datasets and processes, we use length + hash as a representation, e.g.
```python
processes_repr = f"{len(self.processes)}_{law.util.create_hash(self.processes)}"
```

But for calibrators and producers, we use the first 5 instances + hash, e.g.
```python
part = "__".join(self.producers[:5])
if len(self.producers) > 5:
    part += f"__{law.util.create_hash(self.producers[5:])}"
```
Where the final string we add to the store parts is `f"prod__{part}"`

For the output parts in the MLTraining this gets a bit more complicated because you can request different calibs/selectors/producers per config (but per default it's the same one for each config).

(For simplicity, I'll only talk about producers in the following)
One way of proceeding could be:
- first checking if the producers are the same for each config
- if that's true: use the representation as we did for other producer tasks (but maybe only the first 1-2 instances before adding a hash)
- if that's false: add the number of producers per config, flatten the producers and then convert to a hash

e.g. for
```python
# three configs, always the same producer
producers = (("prod",), ("prod",), ("prod",))
```
the representation simply becomes `"prod"`, while something like
```python
# three configs with different producers
producers = (("prod_UL16",), ("prod_UL17",), ("prod_UL18",))
```
would become
```python
f"111_{law.util.create_hash(("prod_UL16", "prod_UL17", "prod_UL18"))}"
```

PR #237, opened 2023-05-08
add config and ml model to output path for MLTraining

mafrahm:
As mentioned in #234, the output path of MLTraining was not set correctly. This PR adds the ml model and config into the output path.

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -686,6 +686,21 @@ class MLModelTrainingMixin(MLModelMixinBase):
             configs=list(self.configs),
         )
 
+    def store_parts(self) -> law.util.InsertableDict:
+        parts = super().store_parts()
+        # since MLTraining is no CalibratorsMixin, SelectorMixin, ProducerMixin, ConfigTask,
+        # all these parts are missing in the `store_parts`
+
+        configs_repr = "__".join(self.configs)
+        parts.insert_after("task_family", "configs", configs_repr)
+
+        # TODO: parts for calib, sel, prod
+
+        if self.ml_model_inst:
+            parts.insert_before("version", "ml_model", f"ml__{self.ml_model_inst.cls_name}")
+
+        return parts
+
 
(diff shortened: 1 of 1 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -686,6 +686,21 @@ def __init__(self, *args, **kwargs):
             configs=list(self.configs),
         )
 
+    def store_parts(self) -> law.util.InsertableDict:
+        parts = super().store_parts()
+        # since MLTraining is no CalibratorsMixin, SelectorMixin, ProducerMixin, ConfigTask,
+        # all these parts are missing in the `store_parts`
+
+        configs_repr = "__".join(self.configs)
```
dsavoiu, line 694: In other places we limit such multi-string specifiers by including the number of string and computing a hash. Should we do this here as well? It would make sense if the number of configs is expected to be >3, otherwise the paths would get quite long.

dsavoiu (approved):
Thanks for adding this! In only have some small comments below, which should be kept in mindfor the future. Otherwise I don't see any obvious issues, so I'm fine with merging this as it is.

columnflow/tasks/framework/mixins.py
mafrahm: Good point! I think this is not important at the moment since nobody currently trains on multiple configs, but it should be addressed in another PR regarding #234

columnflow/tasks/framework/mixins.py
```diff
@@ -686,6 +686,21 @@ def __init__(self, *args, **kwargs):
...
+        parts.insert_after("task_family", "configs", configs_repr)
+
+        # TODO: parts for calib, sel, prod
```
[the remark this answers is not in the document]
dsavoiu: Ah, right, thanks for clarifying! Looks like some refactoring will be required to allow `MLTraining` to access this information, but for now these can be left out. I'll make a note of this in #234 to keep track.

PR #238, opened 2023-05-09
Handle negative and missing values in 2d plots

dsavoiu:
This PR handles several edge cases in the plotting of 2D histograms containing negative or missing values (`np.nan`).

In a previous implementation, missing values were replaced by a large negative value, `EMPTY_FLOAT = -99999.0`, and the `cmin` keyword argument passed to `mplhep` was set to `EMPTY_FLOAT + 1`. While this correctly masked missing values and included all "true" negative values, in some cases this led to the lower bound of the colorbar also starting at a large negative value, and all bins being effectively mapped to the color at the upper end of the scale.

To correct this, the implementation proposed here sets the value of bins for missing data to `np.nan`, which is also used internally by `mplhep` to mask values below `cmin`. In addition, an appropriate color bar normalization is chosen depending on the scale and the histogram content, as follows:

* ~~[`Normalize`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Normalize.html) for linear-scale plots of histograms containing **only positive** or **only negative** values~~
* ~~[`TwoSlopeNorm`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.TwoSlopeNorm.html) (centered at zero) for linear-scale plots of histograms containing **both positive and negative** values~~
* ~~[`LogNorm`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.LogNorm.html) for log-scale plots of histograms containing **only non-negative** values~~
* ~~[`SymLogNorm`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.SymLogNorm.html) for log-scale plots of histograms containing **any negative** values~~

A new optional parameter `--colormap` is also added to the `PlotBase2D` class, allowing users to specify an alternative color scheme for 2D plots on the command line. The default color map is [`viridis`](https://matplotlib.org/stable/tutorials/colors/colormaps.html#sequential), ~~except in the case that a histogram contains both negative and positive values, in which case the cyclic/diverging color map [`twilight_r`](https://matplotlib.org/stable/tutorials/colors/colormaps.html#cyclic) centered at zero is used.~~

*Update*: the colorbar normalization choice has been simplified

* [`Normalize`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Normalize.html) for linear-scale plots 
* [`SymLogNorm`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.SymLogNorm.html) for log-scale plots

Three new optional parameters are added to the plot task to specify the valid `z` axis range and how to handle extreme values:

* `--zlim`: CSV parameter with two entries, indicating the range of values covered by the colorbar. Special values like `min`, `max`, `minabs`, `-maxabs` are supported for setting limits based on the data. Default: `min,max`.
* `--extremes`: how to handle values lying outside the plot range indicated by `--zlim`. Can be `color` (use different colors), `hide` (mask extreme values), or `clip` (use the same color as the min/max of the colorbar). Default: `color`
* `--extreme-colors`: CSV parameter with two entries, indicating which colors to use for values below (above) the lower (upper) limit of the colorbar. Only has an effect when `--extremes` is set to `color`. Default: dark and light gray.

When the histogram contains values outside the `z` range, the colorbar is extended with triangle-shaped markers filled with the corresponding colors as a visual indicator of their presence.

### DIFF
--- a/columnflow/plotting/plot_functions_2d.py
+++ b/columnflow/plotting/plot_functions_2d.py
@@ -11,7 +11,6 @@ from collections import OrderedDict
 import law
 
 from columnflow.util import maybe_import
-from columnflow.columnar_util import EMPTY_FLOAT
 from columnflow.plotting.plot_util import (
     remove_residual_axis,
     apply_variable_settings,
@@ -62,8 +61,62 @@ def plot_2d(
     if not zscale:
         zscale = "log" if (variable_insts[0].log_y or variable_insts[1].log_y) else "linear"
 
+    # add all processes into 1 histogram
+    h_sum = sum(list(hists.values())[1:], list(hists.values())[0].copy())
+    if shape_norm:
+        h_sum = h_sum / h_sum.sum().value
+
+    # mask bins without any entries (variance == 0)
+    h_view = h_sum.view()
+    h_view.value[h_view.variance == 0] = np.nan
+
+    h_sum[...] = h_view
+
+    # check histogram value range
+    vmin, vmax = np.nanmin(h_sum.values()), np.nanmax(h_sum.values())
+    vmin, vmax = np.nan_to_num([vmin, vmax], 0)
+    has_pos_neg = np.sign(vmin * vmax) < 0
+
     # default color map
-    default_cmap = "viridis"
+    default_cmap = "twilight_r" if has_pos_neg else "viridis"
+
+    # choose appropriate colorbar normalization
+    # based on scale type and histogram content
+
+    # log scale
+    if zscale == "log" and vmin >= 0:
+        # use standard LogNorm if values are nonnegative
+        cbar_norm = mpl.colors.LogNorm(
+            vmin=vmin,
+            vmax=vmax,
+        )
+
+    # log scale with negative values
+    elif zscale == "log":
+        # use diverging colormap centered at zero
+        # use SymLogNorm to correctly handle both positive and negative values
+        cbar_norm = mpl.colors.SymLogNorm(
+            vmin=vmin,
+            vmax=vmax,
+            linthresh=max(0.05 * min(abs(vmin), abs(vmax)), 1e-6),
+        )
+
+    # linear scale (both positive and negative values)
+    elif has_pos_neg:
+        # use TwoSlopeNorm to scale positive and negative
+        # parts of thecolorbar independently
+        cbar_norm = mpl.colors.TwoSlopeNorm(
+            vmin=vmin,
+            vmax=vmax,
+            vcenter=0.0,
+        )
+
+    # linear scale (purely positive/negative values)
+    else:
+        cbar_norm = mpl.colors.Normalize(
+            vmin=vmin,
+            vmax=vmax,
+        )
 
     # setup style config
     # TODO: some kind of z-label is still missing
@@ -87,10 +140,9 @@ def plot_2d(
             "lumi": config_inst.x.luminosity.get("nominal") / 1000,  # pb -> fb
         },
         "plot2d_cfg": {
-            "norm": mpl.colors.LogNorm() if zscale == "log" else None,
+            "norm": cbar_norm,
             "cmap": colormap or default_cmap,
             # "labels": True,  # this enables displaying numerical values for each bin, but needs some optimization
-            "cmin": 0,  # display zero entries as white points
             "cbar": True,
             "cbarextend": True,
         },
@@ -99,16 +151,6 @@ def plot_2d(
         },
     }
     style_config = law.util.merge_dicts(default_style_config, style_config, deep=True)
-
-    # add all processes into 1 histogram
-    h_sum = sum(list(hists.values())[1:], list(hists.values())[0].copy())
-    if shape_norm:
-        h_sum = h_sum / h_sum.sum().value
-
-    # set bins without any entries (variance == 0) to EMPTY_FLOAT
-    h_view = h_sum.view()
-    h_view.value[h_view.variance == 0] = EMPTY_FLOAT
-    h_sum[...] = h_view
 
     # apply style_config
     ax.set(**style_config["ax_cfg"])
--- a/columnflow/plotting/plot_functions_2d.py
+++ b/columnflow/plotting/plot_functions_2d.py
@@ -38,6 +38,7 @@ def plot_2d(
     density: bool | None = False,
     shape_norm: bool | None = False,
     zscale: str | None = "",
+    colormap: str | None = "",
     skip_legend: bool = False,
     cms_label: str = "wip",
     process_settings: dict | None = None,
@@ -60,6 +61,9 @@ def plot_2d(
     # how to handle yscale information from 2 variable insts?
     if not zscale:
         zscale = "log" if (variable_insts[0].log_y or variable_insts[1].log_y) else "linear"
+
+    # default color map
+    default_cmap = "viridis"
 
     # setup style config
     # TODO: some kind of z-label is still missing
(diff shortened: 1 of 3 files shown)


pkausw:
Hey @dsavoiu , thanks for this very nice PR! This is actually something that recently happened with increasing frequency, so it's excellent that this is potentially fixed now. Just for documentation purposes, can you ellaborate a little on the circumstances that lead to this behavior?

Additionally (and somewhat unrelated to the core of this PR): There is a new issue #241 that also concerns the z-axis of 2D plots. Since you already worked on the range of the z-axis, do you think it might make sense to already introduce a new option as mentioned in the issue with this PR? Though we could of course open a new PR once development picks up of course. Thanks!

pkausw:
Hey @dsavoiu 

> thanks for the feedback and linking to the new issue #241. I also think it would make sense to have a CLI parameter for the plot tasks to explicitly set the vertical scale. In principle this could be added to this PR for the `z` scale, but since it arguably makes sense to have the same functionality also for setting the `y` axis range in 1D plots, it actually might be better to handle that consistently in a separate PR.

You're absolutely right, it would also make sense for the 1D plot class. I guess it's better to work on this in a new PR then.
> 
> As for when setting `cmin` has an effect on the colorbar starting value, I think this is the expected behavior, at least going by the `mplhep` [documentation](https://github.com/scikit-hep/mplhep/blob/57f85c7224dcf49511aac0a2618115d69cf1057b/src/mplhep/plot.py#L614-L615) and [implementation](https://github.com/scikit-hep/mplhep/blob/57f85c7224dcf49511aac0a2618115d69cf1057b/src/mplhep/plot.py#L740-L741). Despite this, I undestood from @mafrahm that the behavior was different for him, although we didn't identify the exact cause.

Thanks for the clarification. Indeed, I was more interested in the underlying cause of the behavior, since also @Bogdan-Wiederspan has investigated and compared this with different people. It's surprising to me that the same command does different things for different people and we should understand this further imho.

columnflow/plotting/plot_functions_2d.py
```diff
@@ -11,7 +11,6 @@
 import law
 
 from columnflow.util import maybe_import
-from columnflow.columnar_util import EMPTY_FLOAT
```
pkausw, line 14: One very stupid question just for my own education since I'm not an expert in the plotting. In many cases, the default value in case e.g. an observable cannot be calculated in a meaningful way is exactly the `EMPTY_FLOAT`. These values are then masked in the plotting. However, I don't see this here anywhere (also not in the `CreateHistogram` task). Since nothing in the previous version looks like it would do this as well, I assume that there are no issues with this new implementation. But where is the masking of e.g. the `EMPTY_FLOAT` values actually happening? Maybe @mafrahm can shed some light on this for me.

pkausw (commented):
Looks all good, I just have a very basic question regarding one part of the implementation

columnflow/plotting/plot_functions_2d.py
dsavoiu: @mafrahm can comment on the rationale of using `EMPTY_FLOAT` here, but to prevent any confusion I think it's important to note that missing values can arise in two very different contexts.

The first is when the value of a variable *in the underlying arrays* is missing or cannot be calculated in a meaningful way (e.g. leading muon p<sub>T</sub> for an event that has no muons). Here, `EMPTY_FLOAT = -99999.0` is stored by convention, since it lies outside the meaningful range of most variables. When filling a histogram, these entries end up in the under/overflow bins. 

The other case concerns empty bins in a histogram. This happens if there *are no events* where the variable being filled has a value falling within the range of the bin, and is *not* related to the occurrence of `EMPTY_FLOAT` in the input array. Indeed, there is no need to use `EMPTY_FLOAT` past the point where histograms are filled, but I'm guessing @mafrahm chose this as a marker for these bins because it's already used in representing missing values in the other case, which may have caused the confusion. Notwithstanding, I think `np.nan` is a better choice as an empty bin marker in histograms because `matplotlib` natively understands this as a missing value and masks the corresponding bins in the plot.

pkausw: I see, thank you very much for the clarification! It's all clear to me know. Personally, I absolutely agree that `np.nan` is a more natural choice, but we'll see what other people say.

dsavoiu:
Thanks for the feedback, @mafrahm! I haven't investigated in detail, but if it is the case that `cmin` only has effect on the colorbar minimum when some bin values are set to `np.nan` seems like a bug in `mplhep`, so we probably shouldn't rely on this behavior. In any case, great that the proposed fix works for your case as well.

Concerning the colors, I agree in retrospect that changing the default colormap based on the histogram content is probably not a good choice, as this behavior come as a surprise for users. There also seem to be issues with tick locationg using `TwoSlopeNorm`, as can also be seen in the plots shared above by @mafrahm. The reason for choosing this norm when negative values are present was to represent the negative side of the axis with a different scaling, so that the magnitude and distribution of these entries could be judged more easily. While appropriate for some use-cases, clearly this is not suitable for the more common case of histograms where only small negative values are expected (e.g. as a result of negative event weights in MC). Therefore, I have reverted back to the more standard `Normalize` instead of `TwoSlopeNorm`.

Nevertheless, the ability to somehow signal the presence of missing values seems like an important functionality for this task. To make this more general, I have added the `--zlim`, `--extremes`, and `--extreme-colors` parameters to the plot task, as mentioned in the updated PR description. This provides a flexible way to specify how out-of-range values are  handled in general, with the special case of negative values being covered by `--zlim 0,max`.

mafrahm (approved):
This update looks very nice, thank you for implementing all these features, Daniel!

Here is an example plot that shows how it looks when entries are outside the `zlim` range
Screenshot 2023-05-15 at 13 51 26

mafrahm:
> Thanks for the feedback, @mafrahm! I haven't investigated in detail, but if it is the case that `cmin` only has effect on the colorbar minimum when some bin values are set to `np.nan` seems like a bug in `mplhep`, so we probably shouldn't rely on this behavior. In any case, great that the proposed fix works for your case as well.

I don't think it's a bug, it just sets the colorbar minimum to something like `np.min(cmin, h.values())`, so in my case, the colorbar minimum was ~0, since this was the lowest bin entry

dsavoiu:
Thanks @mafrahm for the prompt testing, feedback and approval! I would then go ahead with the merge.

> I don't think it's a bug, it just sets the colorbar minimum to something like `np.min(cmin, h.values())`, so in my case, the colorbar minimum was ~0, since this was the lowest bin entry

It could be something like that, although I couldn't find anything similar in the `hist2dplot` function of `mplhep`. (And if `cmin` were `-99999.0` or similar, as was the case previously, wouldn't the above formula return `-99999.0` if `h.values()` did not contain `nan` values and still lead to the distorted plots we saw?) Anyway, as this seems to be solved now, it's probably not that important to check the behavior of `mplhep` in depth, but we should keep a lookout for possible issues arising in the future when `nan` values are involved.

mafrahm:
> Thanks @mafrahm for the prompt testing, feedback and approval! I would then go ahead with the merge.
> 
> > I don't think it's a bug, it just sets the colorbar minimum to something like `np.min(cmin, h.values())`, so in my case, the colorbar minimum was ~0, since this was the lowest bin entry
> 
> It could be something like that, although I couldn't find anything similar in the `hist2dplot` function of `mplhep`. (And if `cmin` were `-99999.0` or similar, as was the case previously, wouldn't the above formula return `-99999.0` if `h.values()` did not contain `nan` values and still lead to the distorted plots we saw?) Anyway, as this seems to be solved now, it's probably not that important to check the behavior of `mplhep` in depth, but we should keep a lookout for possible issues arising in the future when `nan` values are involved.

Sorry, my mistake, I meant `np.max(cmin, h.values())`

Issue #140, opened 2023-01-11
awkward v2 post-transition changes


riga:
After the transition to awkward v2 in #136, a couple small things are things left to do:

- [x] Use coffea's main release once the awkward2_dev branch is merged and released. 🎉 **Done in 851f5d2c2355514a7406bd801f51bc0daab61b32, moved to some advanced RC.**

https://github.com/uhh-cms/columnflow/blob/89f3429bbc4349ee2269fae497f3bf69a0b06ed3/sandboxes/columnar.txt#L3

The coffea draft PR is [here](https://togithub.com/CoffeaTeam/coffea/pull/736).

- [x] Re-implement the `coffea_parquet` IO method in the `ChunkedIOReader`

- [x] Unpin the numpy version in the conda setup once numba is updated. 🎉 **Done in https://github.com/columnflow/columnflow/pull/246.**

https://github.com/uhh-cms/columnflow/blob/89f3429bbc4349ee2269fae497f3bf69a0b06ed3/setup.sh#L494-L495

The numba/numpy bug is going to be fixed by [this PR](http://shorturl.at/gpvCS).

- [ ] Remove workarounds related to segfaults in Pyarrow.

https://github.com/columnflow/columnflow/blob/89f3429bbc4349ee2269fae497f3bf69a0b06ed3/columnflow/columnar_util.py#L955-L958

https://github.com/columnflow/columnflow/blob/89f3429bbc4349ee2269fae497f3bf69a0b06ed3/columnflow/columnar_util.py#L2083-L2087

- [Reported segfault](https://github.com/dask-contrib/dask-awkward/issues/140)
- [PyArrow bug report](https://togithub.com/apache/arrow/issues/20385)
- [PyArrow PR 🎉](https://github.com/apache/arrow/pull/33634)

PR #240, opened 2023-05-09
Check that used/produced columns exist

### LINKED ISSUE #239: Check if `used` and `produced` columns actually exist in array function input/output
When developing new calibrators/selectors/producers (and in general for debugging purposes), it would be useful check explicitly is all columns declared via the `uses`/`produces` keywords actually exist in the input/output event array. Ideally, the check should be configurable on a per-array-function basis.

This is not entirely straightforward to do, since column specifiers can contain wildcards, but it would be sufficient to check that each specifier matched at least one output column, and otherwise issue a warning or an error. Another complication arises due to the special handling of columns prefixed with an `n*`, which indicate the number of list elements per entry in case of jagged data.

dsavoiu:
This PR proposes a way to check for the existence of `uses`/`produces` columns, as described in issue #239.

Each entry in `uses`/`produces` is checked against the input/output array using the `layout.form.select_columns` method of the underlying AwkwardArray, which also takes wildcards into account. If no matching columns are found for one of the specifiers, an `Exception` is raised. 

The `uses` check is done directly before the main array function (calibrator/selector/producer) is called, assuming that the first argument passed is the event array itself. Analogously, the `produces` check is done on the function result (or the first element of the return value in case the function returns a tuple or a list).

The checks are only performed on columns declared directly under `uses`/`produces`, i.e. not including dependent array functions, which perform their own checks once they are actually run.

To deal with the entry-numbering columns `n*`, which are used by `coffea` to reconstruct the structure of jagged columns but are not present as columns in the AwkwardArray itself, the list of filters used when constructing the event chunks with `coffea.nanoevents.NanoEventsFactory.from_root` is extended by an `n`-prefixed version of all routes present in the file (a superset of all possible entry-numbering columns). This eliminates the need to specify these columns manually under `uses`, at the expense of loading potentially unrelated columns starting with `n`, which seems very unlikely to happen to begin with.

An issue seems to remain related to `nGenPart`, which causes a `ValueError` to be raised by `coffea` if it is present in the list of column filters when loading events from nanoAOD.

```
File "modules/columnflow/columnflow/columnar_util.py", line 2677, in read_coffea_root
    chunk = coffea.nanoevents.NanoEventsFactory.from_root(
  
File "data/software/venvs/venv_columnar_dev/lib/python3.9/site-packages/coffea/nanoevents/factory.py", line 137, in from_root
    return cls._from_mapping(
  
File "data/software/venvs/venv_columnar_dev/lib/python3.9/site-packages/coffea/nanoevents/factory.py", line 384, in _from_mapping
    schema = schemaclass(base_form)
  
File "data/software/venvs/venv_columnar_dev/lib/python3.9/site-packages/coffea/nanoevents/schemas/nanoaod.py", line 181, in __init__
    self._form["fields"], self._form["contents"] = self._build_collections(

File "data/software/venvs/venv_columnar_dev/lib/python3.9/site-packages/coffea/nanoevents/schemas/nanoaod.py", line 344, in _build_collections
    output[name] = zip_forms(

File "data/software/venvs/venv_columnar_dev/lib/python3.9/site-packages/coffea/nanoevents/schemas/base.py", line 36, in zip_forms
    raise ValueError
```

This seems to be a bug in `coffea`, likely related to the `distinctChildrenDeep` method of `GenPartArray` (cf. [coffea PR #698](https://github.com/CoffeaTeam/coffea/pull/698)), for which global indices with type `ListOffsetArray64` are produced internally. Following the transition to awkward2, this type is no longer present in the API, so @riga maybe an update is needed for awkward2?

Package versions:
* `awkward`: 2.0.7
* `coffea`: 0.7.20 (git+https://github.com/riga/coffea.git@9ecdee5#egg=coffea)

Closes #239.


columnflow/production/cms/seeds.py
```diff
@@ -18,8 +18,8 @@
 @producer(
     uses={
         # global columns for event seed
-        "run", "luminosityBlock", "event", "nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon",
-        "nElectron", "nTau", "nSV",
+        "run", "luminosityBlock", "event", "nGenJet", "nGenPart", "nPhoton",
+        "nSV",
```
mafrahm, line 22: Is there a reason, why nGetJet, nGenPart and nPhoton are still listed here in the `uses`?

columnflow/columnar_util.py
```diff
@@ -1389,6 +1390,9 @@ def __init__(
         if skip_func:
             self.skip_func = skip_func.__get__(self, self.__class__)
 
+        if self.check_columns_present is None:
+            self.check_columns_present = self._dependency_sets
```
mafrahm, line 1394: I have some producers where I might want to disable the checks. I would just need to change this to an empty set, right?
```
@producer(
    check_columns_present=set(),
)
```

mafrahm (commented):
Looks mostly good to me, I just left some small questions in the inline comments.

columnflow/production/cms/seeds.py
dsavoiu: Yes, this has to do with how the `from_root` NanoAOD loader works in `coffea`. IIUC, whenever a jagged column like `Jet_pt` is passed to the column filter, the `nJet` column is also required and is read out to work out the length of the nested lists. However, the `nJet` column is then removed by `coffea`, so the check introduced by this PR will fail if `nJet` is present under `uses`. (I assume `nJet` is still defined as a behavior, so calling `events.nJet` would still work and return the equivalent of `ak.num('Jet')`, but `nJet` is no longer an actual column in the event array). In contrast, if no columns are passed that begin with `Jet_`, then `nJet` is read as a normal column and is found by the checker.

So, because this producer reads the `Jet`, `Muon`, `Electron`, `Tau` collections explicitly, we need to remove the corresponding `n*` columns or the column presence check will fail. For `GenJet` or `GenPart` or `Photon`, including the `n*` columns is is not problematic, since the corresponding collections are not read.

columnflow/columnar_util.py
```diff
@@ -1600,6 +1614,35 @@ def _get_produced_columns(self, call_cache: set | None = None) -> set[str]:
...
+            "produce" if io_flag == self.IOFlag.PRODUCES else
+            "find"
+        )
+
+        missing = set()
+        for column in columns:
+            found = ak_array.layout.form.select_columns(column).columns()
+            if not found:
+                missing.add(column)
+
```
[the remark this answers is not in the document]
dsavoiu: I think the issue here is that `column` can be a specifier with wildcards like `Jet.*`, but `found` always returns list of actual column names that match the specifier. So doing `columns.difference(found)` would not work in this case.

columnflow/columnar_util.py
dsavoiu: Yes, exactly. Or you could do `check_columns_present=False` instead of `set()` for cleaner semantics. (But not `None`, since that would use the default, which is to check both `uses` and `produces` columns.)

columnflow/production/cms/gen_top_decay.py
```diff
@@ -13,7 +13,7 @@
 
 
 @producer(
-    uses={"nGenPart", "GenPart.genPartIdxMother", "GenPart.pdgId", "GenPart.statusFlags"},
+    uses={"GenPart.genPartIdxMother", "GenPart.pdgId", "GenPart.statusFlags"},
     produces={"gen_top_decay"},
```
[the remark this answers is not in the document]
dsavoiu: This is a good point. I don't have this producer in the analysis workflow that I was using for testing, but it would indeed be important to check that this doesn't break things. Since the producer does not explicitly require `nGenPart` for the computation, I think it should be safe to remove it. If you are using it in your analysis, do you think you could run a quick test to confirm?

columnflow/columnar_util.py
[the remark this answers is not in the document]
mafrahm: Good point, then we can keep the implementation as it is :+1:

columnflow/production/cms/seeds.py
mafrahm: OK, thanks for clarifying

columnflow/production/cms/gen_top_decay.py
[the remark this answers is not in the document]
mafrahm: I did not try this specific producer, but for my signal-specific Gen producer, the `"nGenPart"` column seems to be needed (`AttributeError: no field named 'genPartIdxMotherG'`, even though I added `GenPart.*` to my `uses`).

So to be safe, we might just want to keep `nGenPart` in there and skip the check for this producer?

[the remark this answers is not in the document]
dsavoiu: Ah, I see. I assume you mean the [`gen_hbw_decay_products`](https://github.com/uhh-cms/hh2bbww/blob/3165e824c63dadeaf8bf930b58a99844efbeefc1/hbw/production/gen_hbw_decay.py#L17) producer? In that case the `nGenPart` information is likely needed by the NanoAOD schema for the `distinctParent` accessor to work correctly.

For the `gen_top_decay` producer, the `nGenPart` would also be needed for  `distinctChildrenDeep`, but I find that adding that to the columns when loading a NanoAOD file with `coffea` leads to the error in the PR description. For that reason, `nGenPart` is the only column that is not being added automatically for NanoAOD.

However, I am a bit surprised that your producer runs at all if you *do* add `nGenPart` to the `uses`, as for me the issue occurs even before the producer code is executed. But I see now that there was an update from @riga to the software sandboxes, so maybe the issue was fixed upstream and we can include `nGenPart` back. I'll do some testing.

dsavoiu:
> @dsavoiu I updated some software packages in the central sandboxes:
> 
> ```
> awkward~=2.2
> uproot~=5.0
> git+https://github.com/CoffeaTeam/coffea.git@v2023.5.0.rc0#egg=coffea  # they're using tags now 🎉
> dask-awkward~=2023.5
> ```
> 
> Can you check if the `ListOffsetArray64` error is gone with this?

Hi @riga, that sound promising, thanks! I'll do some checks and let you know.

columnflow/production/cms/gen_top_decay.py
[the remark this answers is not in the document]
riga: > To deal with the entry-numbering columns n*, which are used by coffea to reconstruct the structure of jagged columns but are not present as columns in the AwkwardArray itself, the list of filters used when constructing the event chunks with coffea.nanoevents.NanoEventsFactory.from_root is extended by an n-prefixed version of all routes present in the file (a superset of all possible entry-numbering columns). This eliminates the need to specify these columns manually under uses, at the expense of loading potentially unrelated columns starting with n, which seems very unlikely to happen to begin with.

Maybe we could offload this to the IO? I.e., in case one or more columns under "Jet" are to be loaded AND the input format is ROOT, the "nJet" column is loaded as well. This way, when a parquet file (with already existing starts / stops info) is read, no extra counter column is loaded.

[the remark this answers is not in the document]
riga: You're right of course, I think I confused two PRs :)

columnflow/columnar_util.py
```diff
@@ -1554,7 +1558,13 @@ def get_dependencies(self, include_others: bool = False) -> set[ArrayFunction |
 
         return deps
 
-    def _get_columns(self, io_flag: IOFlag, call_cache: set | None = None, debug=False) -> set[str]:
+    def _get_columns(
+        self,
+        io_flag: IOFlag,
+        call_cache: set | None = None,
+        debug: bool = False,
```
riga, line 1565: Sorry for piggybacking on this PR, but since you touched this line, could you remove the `debug` argument? This looks like some dev remnant that I left there a while ago :)

columnflow/columnar_util.py
```diff
@@ -1600,6 +1614,35 @@ def _get_produced_columns(self, call_cache: set | None = None) -> set[str]:
     def produced_columns(self) -> set[str]:
         return self._get_produced_columns()
 
+    def _check_columns(self, ak_array: ak.Array, io_flag: IOFlag):
```
riga, line 1617: ` -> None`

Just for consistency.

columnflow/columnar_util.py
```diff
@@ -1600,6 +1614,35 @@ def _get_produced_columns(self, call_cache: set | None = None) -> set[str]:
...
+        )
+        action = (
+            "receive" if io_flag == self.IOFlag.USES else
+            "produce" if io_flag == self.IOFlag.PRODUCES else
+            "find"
+        )
+
+        missing = set()
+        for column in columns:
+            found = ak_array.layout.form.select_columns(column).columns()
```
riga, line 1636: Maybe one can use the `has_ak_column` helper here? This would also make sense as `column` could also be a `Route` object.

riga (changes requested):
Looks very good to me! Having this feature included will make things far more streamlined, especially for newcomers that don't need to understand ROOTs constraints for storing jagged data.

Just a few minor things that only take a minute to address :)

And as a crossref: we should add the `n*` columns to the filter names in the coffea_parquet IO method once it's reimplemented (https://github.com/columnflow/columnflow/issues/140, second item). **EDIT**: this was done in the meantime, see [below](https://github.com/columnflow/columnflow/pull/240#issuecomment-1561155569) :)

riga:
@dsavoiu Since the `coffea_parquet` reader method was re-added in the meantime (https://github.com/columnflow/columnflow/pull/247), I took the liberty to paste your changes also there. I hope this was ok for you :)

columnflow/columnar_util.py
dsavoiu: Sure, no problem.

columnflow/columnar_util.py
dsavoiu: Done

columnflow/columnar_util.py
```diff
@@ -1600,6 +1614,35 @@ def _get_produced_columns(self, call_cache: set | None = None) -> set[str]:
...
+        columns = self._get_columns(
+            io_flag=io_flag,
+            # only check own columns
+            no_dependencies=True,
+        )
+        action = (
+            "receive" if io_flag == self.IOFlag.USES else
+            "produce" if io_flag == self.IOFlag.PRODUCES else
+            "find"
+        )
```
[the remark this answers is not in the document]
dsavoiu: Yeah, that makes sense. I'll move it to the `if` clause.

columnflow/columnar_util.py
dsavoiu: I see the point about `column` potentially being a `Route`, but looking at the code again it looks like the only way this can happen is if the columns are given explicitly as routes in the array function decorator, i.e. `uses={Route("foo")}`. As far as I know, in the current code base only strings are used, so do we want to support `Route` objects explicitly?

In any case, I think we need `select_columns` here because that function automatically expands wildcards, while with `has_ak_columns` one would need the actual column names. For `Route` objects, it should be straightforward to just pass `column.string_column` to `select_columns` instead.

dsavoiu:
> @dsavoiu Since the `coffea_parquet` reader method was re-added in the meantime (#247), I took the liberty to paste your changes also there. I hope this was ok for you :)

Absolutely! Thanks for keeping things in sync with the current developments!

columnflow/columnar_util.py
riga: Yup, makes sense :+1:

columnflow/columnar_util.py
```diff
@@ -1553,7 +1559,12 @@ def get_dependencies(self, include_others: bool = False) -> set[ArrayFunction |
 
         return deps
 
-    def _get_columns(self, io_flag: IOFlag, call_cache: set | None = None, debug=False) -> set[str]:
+    def _get_columns(
+        self,
+        io_flag: IOFlag,
+        call_cache: set | None = None,
+        no_dependencies: bool = False,
```
riga, line 1566: There are a other methods that have an argument that control whether certain actions should be taken also for dependencies. They are named a bit inconsistently though (deps, dependencies, recursive, ..). Could you use `dependencies: bool = True` here?

I'll prepare another PR then to adjust the other locations in existing code.

columnflow/columnar_util.py
```diff
@@ -1599,6 +1614,35 @@ def _get_produced_columns(self, call_cache: set | None = None) -> set[str]:
...
+    def _check_columns(self, ak_array: ak.Array, io_flag: IOFlag) -> None:
+        """
+        Check if awkward array contains at least one column matching each
+        entry in 'uses' or 'produces' and raise Exception if none were found.
+        """
+
+        columns = self._get_columns(
+            io_flag=io_flag,
+            # only check own columns
+            no_dependencies=True,
```
riga, line 1626: Would be `dependencies=False` with the adjustment above.

columnflow/columnar_util.py
```diff
@@ -2654,6 +2723,16 @@ def read_coffea_parquet(
...
         ):
             read_dictionary = [Route(s).string_column for s in read_columns]
+
+            # add names prefixed with an 'n' to the list of columns to read
+            # (needed to construct the nested list structure of jagged columns)
+            maybe_jagged_fields = {Route(s)[0] for s in read_columns}
+            read_dictionary.extend(
+                f"n{field}"
+                for field in maybe_jagged_fields
+                if field != "GenPart"
```
riga, line 2733: Is `GenPart` still a special case with the newest software deps?

columnflow/production/cms/seeds.py
```diff
@@ -18,8 +18,8 @@
 @producer(
     uses={
         # global columns for event seed
-        "run", "luminosityBlock", "event", "nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon",
-        "nElectron", "nTau", "nSV",
+        "run", "luminosityBlock", "event",
+        "Photon.pt", "SV.pt",
```
riga, line 22: Are you reading the "pt" fields to trigger that the counter fields are read too?
Couldn't one just keep the "n*" fields here? It might be very costly to e.g. read all GenPart pt's just for the sake of knowing how many particles exist per event.

riga (changes requested):
Thanks for the adjustments!

I went through the PR again and cam up with a couple more comments and questions inline.

columnflow/columnar_util.py
dsavoiu: That makes sense, I can adjust that :+1:

columnflow/columnar_util.py
dsavoiu: OK.

columnflow/columnar_util.py
dsavoiu: Good point. I think the issue is solved now, I just forgot to remove the `if` condition here. I'll look into this.

columnflow/production/cms/seeds.py
dsavoiu: Yes, `pt` is just a proxy in this case so we can get the counts via `ak.num`. But I agree that the overhead might not be justified. This was a bit of a last-minute change since I was seeing issues with `n*`, but I will look into what was causing that and revert this if I can find a solution.

On a related note, an advantage of not relying on the `n*` fields would be that `deterministic_seeds` could then be run during `cf.ProduceColumns` and later, where these fields are not available. Is that something we want to support?

riga: > On a related note, an advantage of not relying on the n* fields would be that deterministic_seeds could then be run during cf.ProduceColumns and later, where these fields are not available. Is that something we want to support?

Good question. I think for that we would need to make sure that the counter columns (or similar ones) are available. Maybe we can split implementation from the `used` declaration, e.g.

```python
@producer(
    uses={"run", "luminosityBlock", "event"},  # commonly-used columns
    produces={"deterministic_seed"},
)
def deterministic_event_seeds_impl(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
    # implementation
    ...


# specialization using counter columns
@producer(
    uses={
        deterministic_event_seeds_impl,
        "nGenJet", "nGenPart", "nJet", "nPhoton", "nMuon",
        "nElectron", "nTau", "nSV",
        # first-object columns for event seed
        "Tau.jetIdx", "Tau.decayMode",
        "Muon.jetIdx", "Muon.nStations",
        "Jet.nConstituents", "Jet.nElectrons", "Jet.nMuons",
    },
    produces={deterministic_event_seeds_impl},
)
def deterministic_event_seeds(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
    return self[deterministic_event_seeds_impl](events, **kwargs)
```

and then re-use the implementation at a later stage in case we need it.

riga: I agree, this is more complicated than I anticipated. Let's keep this for now and put the actual issue at hand on our lists.

columnflow/columnar_util.py
dsavoiu: Done

columnflow/columnar_util.py
dsavoiu: Done

dsavoiu:
Should be ready for merging now, @riga.

PR #242, opened 2023-05-12
Allow producer setup function to return paths to parquet files

mafrahm:
While producers can already set custom requirements, it is currently not possible to include columns from these requirements in the loop of the chunked reader. It is therefore currently not possible to implement a producer that uses the `MLEvaluation` output to create some new columns. This PR implements a first version to make this possible.

A possible implementation for such a producer might look like this:

```python
@producer(uses={"dnn_column"})
def my_producer(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
    # access the 'dnn_column' column, that has been produced in the `MLEvaluation` task
    my_dnn_column = events.dnn_column
    ...

    return events

@my_producer.requires
def my_producer_reqs(self: Producer, reqs: dict) -> None:
    if "ml"  in reqs:
        return

    from columnflow.tasks.ml import MLEvaluation
    # add the requirements to the task (this already works without the PR)
    reqs["ml"] = MLEvaluation.req(self.task, ml_model="my_ml_model")


@my_producer.setup
def my_producer_setup(self: Producer, reqs: dict, inputs: dict) -> Sequence[str]:
    # return the paths of the parquet files you want to load into the chunked reader
    col_file = inputs["ml"]["mlcolumns"].path
    return [col_file]
```

This currently only allows to include 'hard-coded' requirements to the producer. An alternative would be to include the `ml-models` requirements as a parameter (similar to how it's done with `CreateHistograms`). If possible, we might want to implement both.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -1888,7 +1888,7 @@ class TaskArrayFunction(ArrayFunction):
         reqs: dict,
         inputs: dict,
         call_cache: set | None = None,
-    ) -> None:
+    ) -> None | Sequence[str]:
         """
         Recursively runs the :py:meth:`setup_func` of this instance and all dependencies. *reqs*
         corresponds to the requirements created by :py:func:`run_requires`, and *inputs* are their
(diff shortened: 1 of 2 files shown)


riga (changes requested):
Just two minor things.

I would add the open localization issue in the next commit then.

riga (approved):

PR #243, opened 2023-05-16
Uncomment uncertainty combination.

### LINKED ISSUE #229: Error due to `combine_uncs` in yields task
There seems to be an issue related to the `combine_uncs` option when transforming the yields into a string representation

https://github.com/columnflow/columnflow/blob/master/columnflow/tasks/yields.py#L225

riga:
This PR just uncomments a single line in the yields task which instructs the `Number` object to combine all uncertainties.

For this line to work, the sandbox needs to be updated since the `combine_uncs` feature was added to scinum only recently.

E.g.

```bash
source $CF_BASE/sandboxes/venv_columnar_dev.sh reinstall
```

Closes #229.

### DIFF
--- a/columnflow/tasks/yields.py
+++ b/columnflow/tasks/yields.py
@@ -230,7 +230,7 @@ class CreateYieldTable(
 
                     # format yields into strings
                     yields_str[category].append((value / norm_factor).str(
-                        # combine_uncs="all",  # TODO
+                        combine_uncs="all",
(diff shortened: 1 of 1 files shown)


pkausw:
@riga Could you put a notification for the users somehwere in this case then? Since the PR was closed so fast, I'm not sure if people will be aware of the necessity to setup the software stack again

PR #244, opened 2023-05-16
add calib/sel/prod in MLTraining output parts

### LINKED ISSUE #234: missing information in MLTraining outputs paths

mafrahm:
This PR adds Calibrators, Selectors and Producers to the MLTraining output paths, always adding at least the first two producers (etc.) to the output path. If the producers are the same for all configs, the duplicates are removed. If there are different producers per config, the number of producers per config is also added to the string.

closes #234

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -712,7 +712,7 @@ class MLModelTrainingMixin(MLModelMixinBase):
             else:
                 # when functions differ between configs, flatten
                 n_fct_per_config = "".join(str(len(x)) for x in fct_names)
-                fct_names = sum(fct_names[1:], fct_names[0])
+                fct_names = tuple(fct_name for fct_names_cfg in fct_names for fct_name in fct_names_cfg)
 
             part = "_".join(fct_names[:2])
 
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -714,7 +714,7 @@ class MLModelTrainingMixin(MLModelMixinBase):
                 n_fct_per_config = "".join(str(len(x)) for x in fct_names)
                 fct_names = tuple(fct_name for fct_names_cfg in fct_names for fct_name in fct_names_cfg)
 
-            part = "_".join(fct_names[:2])
+            part = "__".join(fct_names[:2])
 
             if len(fct_names) > 2:
                 part += f"_{n_fct_per_config}_{law.util.create_hash(fct_names[2:])}"
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -691,10 +691,35 @@ class MLModelTrainingMixin(MLModelMixinBase):
         # since MLTraining is no CalibratorsMixin, SelectorMixin, ProducerMixin, ConfigTask,
         # all these parts are missing in the `store_parts`
 
-        configs_repr = "__".join(self.configs)
+        configs_repr = "__".join(self.configs[:5])
+
+        if len(self.configs) > 5:
+            configs_repr += f"_{law.util.create_hash(self.configs[5:])}"
+
         parts.insert_after("task_family", "configs", configs_repr)
 
-        # TODO: parts for calib, sel, prod
+        task_array_functions = {
+            "calib": self.calibrators,
+            "sel": tuple((sel,) for sel in self.selectors),
+            "prod": self.producers,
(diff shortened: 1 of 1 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -691,10 +691,35 @@ def store_parts(self) -> law.util.InsertableDict:
...
 
-        # TODO: parts for calib, sel, prod
+        task_array_functions = {
+            "calib": self.calibrators,
+            "sel": tuple((sel,) for sel in self.selectors),
+            "prod": self.producers,
+        }
+
+        for label, fct_names in task_array_functions.items():
+            if all(map(lambda x: x == fct_names[0], fct_names)):
```
dsavoiu, line 708: Not that important, but doing `fct_names[0]` could raise an `IndexError` if empty lists are passed for `self.configs`. Is this caught somewhere?

columnflow/tasks/framework/mixins.py
```diff
@@ -691,10 +691,35 @@ def store_parts(self) -> law.util.InsertableDict:
...
+                # when functions are the same per config, only use them once
+                fct_names = fct_names[0]
+                n_fct_per_config = str(len(fct_names))
+            else:
+                # when functions differ between configs, flatten
+                n_fct_per_config = "".join(str(len(x)) for x in fct_names)
+                fct_names = sum(fct_names[1:], fct_names[0])
```
dsavoiu, line 715: What is the reason for using `sum` in this way here? It seems like it would be adding a tuple to a string, which would raise an exception. Since you only use `fct_names[2:]` below anyway, couldn't this line be dropped completely?

columnflow/tasks/framework/mixins.py
```diff
@@ -691,10 +691,35 @@ def store_parts(self) -> law.util.InsertableDict:
...
+
+            part = "_".join(fct_names[:2])
```
dsavoiu, line 717: For consistency with other mixing, the joiner should probably be a double underscore here.
```suggestion
            part = "__".join(fct_names[:2])
```

dsavoiu (changes requested):
Thanks for following up on this! Looks good overall, but I noticed a few minor things that would need a fix or clarification (see below).

columnflow/tasks/framework/mixins.py
dsavoiu: Ah, sorry, the entries in `fcn_names` are all tuples, so this is essentially just flattening the list of producers across all configs, right? But then it might be better to use a comprehension for better readability.
```suggestion
                fct_names = tuple(fct_name for fct_names_cfg in fct_names for fct_name in fct_names_cfg)
```

columnflow/tasks/framework/mixins.py
```diff
@@ -691,10 +691,35 @@ def store_parts(self) -> law.util.InsertableDict:
...
-        configs_repr = "__".join(self.configs)
+        configs_repr = "__".join(self.configs[:5])
+
+        if len(self.configs) > 5:
+            configs_repr += f"_{law.util.create_hash(self.configs[5:])}"
+
         parts.insert_after("task_family", "configs", configs_repr)
 
-        # TODO: parts for calib, sel, prod
+        task_array_functions = {
```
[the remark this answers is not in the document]
mafrahm: I don't think so, python dicts are automatically ordered as of python 3.7

columnflow/tasks/framework/mixins.py
mafrahm: Yes, you can't run a MLModel without specifying some config (otherwise you get this Exception)
```
Exception: MLModel 'base_test' received configs '()' to define training configs, but did not define any
```

riga (approved):
Looks good to me with the changes addressing the review comments 👍

columnflow/tasks/framework/mixins.py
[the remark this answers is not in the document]
riga: Resolving this for now since the dict is gone entirely :)

columnflow/tasks/framework/mixins.py
riga: It seems that any `[][0]` would be caught here as well now, so resolving :)

dsavoiu (approved):
Looks good to me as well, thanks for addressing the review comments!

PR #245, opened 2023-05-22
Feature/producers groups

mafrahm:
This PR allows setting `producer_groups` and `calibrator_groups` in the config.

### DIFF
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -449,33 +449,46 @@ class ConfigTask(AnalysisTask):
         return kwargs
 
     @classmethod
-    def resolve_config_parameter(
+    def resolve_config_default(
+        cls,
+        config_inst: od.Config,
+        param: str | None = None,
+        default_str: str | None = None,
+    ) -> str | None:
+        # expand default when param is empty
+        if not param or param == law.NO_STR:
+            param = config_inst.x(default_str, None)
+
+        return param
+
+    @classmethod
+    def resolve_config_default_and_groups(
         cls,
         config_inst: od.Config,
         param: tuple[str] | None = None,
-        param_name: str | None = None,
+        default_str: str | None = None,
+        groups_str: str | None = None,
     ) -> tuple[str]:
         """
         Resolves tuple of inputs *param* by using groups defined in the *config_inst*
         that map a string to a tuple of strings. If *param* is empty, the default
         defined in the config is returned.
         """
-        if param and param != law.NO_STR:
-            param = law.util.make_list(param)
+        # expand default when param is empty
+        if not param or param == law.NO_STR:
+            param = config_inst.x(default_str, None)
+            if not param:
+                return ()
 
-            # expand to param group
-            param_group = config_inst.x(f"{param_name}_groups", {})
-            if param_group:
-                outp = []
-                for val in param:
-                    outp.extend(param_group.get(val, [val]))
-            return tuple(outp)
+        # expand to groups
+        if param_groups := config_inst.x(groups_str, {}):
+            values = []
+            for val in law.util.make_list(param):
+                values.extend(param_groups.get(val, [val]))
+            return tuple(values)
 
-        # use default if available and no param is set
-        if outp := config_inst.x(f"default_{param_name}", None):
-            return outp if isinstance(outp, tuple) else (outp,)
-
-        return ()
+        # return the param as is
+        return law.util.make_tuple(param)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -448,6 +448,35 @@ class ConfigTask(AnalysisTask):
 
         return kwargs
 
+    @classmethod
+    def resolve_config_parameter(
+        cls,
+        config_inst: od.Config,
+        param: tuple[str] | None = None,
+        param_name: str | None = None,
+    ) -> tuple[str]:
+        """
+        Resolves tuple of inputs *param* by using groups defined in the *config_inst*
+        that map a string to a tuple of strings. If *param* is empty, the default
+        defined in the config is returned.
+        """
+        if param and param != law.NO_STR:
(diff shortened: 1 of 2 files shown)


columnflow/tasks/framework/base.py
```diff
@@ -448,6 +448,35 @@ def get_array_function_kwargs(cls, task=None, **params):
...
+    def resolve_config_parameter(
+        cls,
+        config_inst: od.Config,
+        param: tuple[str] | None = None,
+        param_name: str | None = None,
+    ) -> tuple[str]:
+        """
+        Resolves tuple of inputs *param* by using groups defined in the *config_inst*
+        that map a string to a tuple of strings. If *param* is empty, the default
+        defined in the config is returned.
```
riga, line 461: In the doc string, could you mention

- the `f"default_{param_name}"` name of the default entry, and
- the `f"{param_name}_groups"` name of the group entry?

columnflow/tasks/framework/base.py
```diff
@@ -448,6 +448,35 @@ def get_array_function_kwargs(cls, task=None, **params):
...
+        if param and param != law.NO_STR:
+            param = law.util.make_list(param)
+
+            # expand to param group
+            param_group = config_inst.x(f"{param_name}_groups", {})
+            if param_group:
+                outp = []
+                for val in param:
+                    outp.extend(param_group.get(val, [val]))
+            return tuple(outp)
```
riga, line 472: Mh, `outp` will not always be defined here. I guess you want to return and empty tuple in this case?

columnflow/tasks/framework/base.py
```diff
@@ -448,6 +448,35 @@ def get_array_function_kwargs(cls, task=None, **params):
...
+
+        # use default if available and no param is set
+        if outp := config_inst.x(f"default_{param_name}", None):
+            return outp if isinstance(outp, tuple) else (outp,)
+
+        return ()
```
riga, line 478: I thought about it again, and maybe we should make the cases which are resolved a bit more expressive. Suggestion:

```python
# expand default when param is empty
if not param or param == law.NO_STR:
    param = config_inst.x(f"default_{param_name}", None)
    if not param:
        return ()

# expand to groups
if param_groups := config_inst.x(f"{param_name}_groups", {}):
    values = []
    for val in law.util.make_list(param):
        values.extend(param_group.get(val, [val]))
    return tuple(values)

# return the param as is
return law.util.make_tuple(param)
```

columnflow/tasks/framework/mixins.py
```diff
@@ -608,20 +597,23 @@ def resolve_producers(
...
 
-        # use config defaults in case each config has one
-        if not producers:
-            defaults = tuple(
-                ProducersMixin.get_default_producers(config_inst)
-                for config_inst in ml_model_inst.config_insts
-            )
-            if all(defaults):
-                producers = defaults
+        producers = producers if producers else ((),)
```
riga, line 600: These two lines should be merged into one.

```python
producers = params.get("producers") or ((),)
```

columnflow/tasks/framework/mixins.py
```diff
@@ -523,20 +509,23 @@ def resolve_calibrators(
...
 
-        # use config defaults in case each config has one
-        if not calibrators:
-            defaults = tuple(
-                CalibratorsMixin.get_default_calibrators(config_inst)
-                for config_inst in ml_model_inst.config_insts
-            )
-            if all(defaults):
-                calibrators = defaults
+        calibrators = calibrators if calibrators else ((),)
```
riga, line 512: These lines should be merged into one.

```python
calibrators = params.get("calibrators") or ((),)
```

riga (changes requested):
A couple inline comments. Plus two additional general things:

- Can you also cover the singular cases (get_default_{calibrator,selector,producer}, etc) by a method that resolves config parameters? The cleanest way would be a second method with a different name (see below), that does not resolve groups.
- The name of the method should be a bit more verbose I think. How about
  - `resolve_config_default_and_groups` for the plural case, and
  - `resolve_config_default` for the singular one?

columnflow/tasks/framework/base.py
mafrahm: Good point, I decided to instead implement these strings as input parameters `default_str` and `groups_str`

columnflow/tasks/framework/base.py
```diff
@@ -448,6 +448,48 @@ def get_array_function_kwargs(cls, task=None, **params):
 
         return kwargs
 
+    @classmethod
+    def resolve_config_default(
+        cls,
+        config_inst: od.Config,
+        param: str | None = None,
+        default_str: str | None = None,
+    ) -> str | None:
```
riga, line 457: Docstring :)
Also, is `param` optional?

columnflow/tasks/framework/base.py
```diff
@@ -448,6 +448,48 @@ def get_array_function_kwargs(cls, task=None, **params):
...
+        config_inst: od.Config,
+        param: tuple[str] | None = None,
+        default_str: str | None = None,
+        groups_str: str | None = None,
+    ) -> tuple[str]:
+        """
+        Resolves tuple of inputs *param* by using groups defined in the *config_inst*
+        that map a string to a tuple of strings. If *param* is empty, the default
+        defined in the config is returned.
+        """
```
riga, line 476: Can you mention the meaning of the two latter parameters here?

columnflow/tasks/framework/mixins.py
```diff
@@ -564,22 +546,21 @@ def resolve_selectors(
...
 
         # broadcast to configs
         n_configs = len(ml_model_inst.config_insts)
         if len(selectors) == 1 and n_configs != 1:
             selectors = tuple(selectors * n_configs)
 
+        # use config defaults
+        selectors = tuple(
+            ConfigTask.resolve_config_default(
+                config_inst, selectors[i], "default_selector",
```
riga, line 559: Just syntax stuff for consistency :)

```python
config_inst,
selectors[i],
"default_selector",
```

riga (changes requested):

riga (approved):
LGTM!

PR #247, opened 2023-05-23
Reimplement coffea_parquet handler.

riga:
With the latest coffea 2023 release candidate, it is possible again to read parquet files through the NanoEventsFactory. Our `ChunkedIOHandler` used to support this method, but support was halted as part of our ak v2 transition in cf.

This PR enables the `coffea_parquet` again and the code is almost an exact copy of the `coffea_root` method, except for some options to be passed to `pa.ParquetFile` instead of `uproot.open`.

This checks the second item in https://github.com/columnflow/columnflow/issues/140.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -2405,7 +2405,7 @@ class ChunkedIOHandler(object):
 
             - "uproot_root"
             - "coffea_root"
-            - "coffea_parquet" (currently unsupported)
+            - "coffea_parquet"
             - "awkward_parquet"
         """
         if source_type is None:
@@ -2438,13 +2438,13 @@ class ChunkedIOHandler(object):
                 cls.close_coffea_root,
                 cls.read_coffea_root,
             )
-        # if source_type == "coffea_parquet":
-        #     return cls.SourceHandler(
-        #         source_type,
-        #         cls.open_coffea_parquet,
-        #         cls.close_coffea_parquet,
-        #         cls.read_coffea_parquet,
-        #     )
+        if source_type == "coffea_parquet":
+            return cls.SourceHandler(
+                source_type,
+                cls.open_coffea_parquet,
+                cls.close_coffea_parquet,
+                cls.read_coffea_parquet,
+            )
         if source_type == "awkward_parquet":
             return cls.SourceHandler(
                 source_type,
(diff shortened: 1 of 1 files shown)


columnflow/columnar_util.py
```diff
@@ -2623,58 +2623,66 @@ def read_coffea_root(
...
+        read_options["runtime_cache"] = None
+        read_options["persistent_cache"] = None
+
+        # inject read_columns
+        if read_columns and (
+            "parquet_options" not in read_options or
+            "read_dictionary" not in read_options["parquet_options"]
+        ):
+            read_dictionary = [Route(s).string_column for s in read_columns]
+            read_options.setdefault("parquet_options", {})["read_dictionary"] = read_dictionary
```
pkausw, line 2675: This is very picky, but `read_dictionary` is not really a dictionary, right? A change of the variable name might be more expressive e.g.

```suggestion
            columns_to_read = [Route(s).string_column for s in read_columns]
            read_options.setdefault("parquet_options", {})["read_dictionary"] = columns_to_read
```

Though I guess one could make the argument that this way, the variable name is the same as in the final `parquet_options`, so I guess we can also just leave it as is

columnflow/columnar_util.py
```diff
@@ -2623,58 +2623,66 @@ def read_coffea_root(
...
+        read_options: dict | None = None,
+        read_columns: set[str | Route] | None = None,
+    ) -> coffea.nanoevents.methods.base.NanoEventsArray:
+        """
+        Given a the location of a parquet file *source_object*, returns an awkward array chunk
+        referred to by *chunk_pos*, assuming nanoAOD structure. *read_options* are passed to
+        ``coffea.nanoevents.NanoEventsFactory.from_parquet``. *read_columns* are converted to
+        strings and, if not already present, added as nested field
+        ``parquet_options.read_dictionary`` to *read_options*.
+        """
```
pkausw, line 2663: (The following comment is triggered by me currently working on other doc strings)

Can you format the doc string such that `sphinx` can automatically compile a nice overview over the parameters and the information about what is returned? So something like

```suggestion
        """
        Function to read parquet files created by :external+coffea:doc:`index` in the nanoAOD format.
        
        Given a the location of a parquet file *source_object*, returns an awkward array chunk
        referred to by *chunk_pos*, assuming nanoAOD structure. *read_options* are passed to
        :external+coffea:py:meth:`coffea.nanoevents.NanoEventsFactory.from_parquet`. *read_columns* are converted to
        strings and, if not already present, added as nested field
        ``parquet_options.read_dictionary`` to *read_options*.
        
        :param cls: The current :py:class:`~.SourceHandler` class
        :type cls: SourceHandler
        :param source_object: path to parquet file containing events to be read
        :type source_object: str
        :param chunk_pos: Position of the current chunk in the ensemble of events
        :type chunk_pos: ChunkPosition
        :param read_options: Options to consider when reading the data from *source_object*, parsed to :external+coffea:py:meth:`coffea.nanoevents.NanoEventsFactory.from_parquet`, defaults to None
        :type read_options: dict | None, optional
        :param read_columns: :py:class:`set` of columns to read from *source_object*. Converted to conform with the format expected by :external+coffea:py:meth:`coffea.nanoevents.NanoEventsFactory.from_parquet`, defaults to None
        :type read_columns: set[str | Route] | None, optional
        :return: coffea :external+coffea:py:class:`~coffea.nanoevents.methods.base.NanoEventsArray` containing events with additional behaviors attached
        :rtype: coffea.nanoevents.methods.base.NanoEventsArray
        """
```

pkausw: Similar is of course true for the other doc strings as well

pkausw (changes requested):
Nice! It's very good that we have this feature again :+1: Just one comment concerning mainly the documentation strings - though we probably have to tackle that in other parts of the module also at some point, so I guess it's not super critical at the moment

columnflow/columnar_util.py
riga: Yeah, since it's done the same in `coffea_root` (filter_name), I thought it would be good to use the exact same terminology as used by the underlying tool (pyarrow) for easier debugging when looking at their docs. I'd tend to continue using this :)

columnflow/columnar_util.py
riga: Mh, would you be ok with moving this to a dedicated PR where we adjust and prettify the doc strings? I like the newer style though (maybe except for the redundant argument type definitions which could be taken from the type hints).

columnflow/columnar_util.py
pkausw: Sounds reasonable :+1:

columnflow/columnar_util.py
pkausw: Jup, I've been working on prettifying the doc strings :D It would be good to talk about this in the next meeting with our collaborators, but I guess for this PR we can also just move ahead. Just thought we could already get started and do a uniform style in the upcoming PRs, but ok :+1:

pkausw (approved):

PR #248, opened 2023-05-23
Catch unkown process ids.

### LINKED ISSUE #235: Catch error when forgetting process_name, but not dataset_name.
When a `dataset_name` is set, but forgotten in `process_names`, then a wrong processID is used in [`normalization_weights_setup()`](https://github.com/columnflow/columnflow/blob/master/columnflow/production/normalization.py)

To be more specific it happens within the for loop:
```
max_id = max(process_inst.id for process_inst in process_insts)
sum_weights_table = sp.sparse.lil_matrix((1, max_id + 1), dtype=np.float32)
for process_id, sum_weights in selection_stats["sum_mc_weight_per_process"].items():
        sum_weights_table[0, int(process_id)] = sum_weights  # error happens here
```

The error is not always triggered for example: 
```
dataset_names = ["graviton_hh_ggf_bbtautau_m1250_madgraph", "graviton_hh_vbf_bbtautau_m400_madgraph"]
process_names= ["graviton_hh_ggf_bbtautau_m1250"]
```
will trigger the error.

The reason for this is, that max_id is set using the `process_names` processID, but the value of processID is coming from the entry in `dataset_name`. Therefore a table is created with a smaller number of columns than necessary. 

I think we should catch this error, since the error message is not corresponding at all with the actual problem.

riga:
This PR adds a check to the config_utils verifying that the linked processes to every dataset are actually known by the config itself. The check is rather trivial and should be used in all analysis configs as a safety measure, which is why I also injected it into the template analysis.

In addition, the PR contains an additional check in the normalization weight producer as mentioned in #235.

Closes #235.

### DIFF
--- a/columnflow/config_util.py
+++ b/columnflow/config_util.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 __all__ = [
     "get_root_processes_from_campaign", "add_shift_aliases", "get_shifts_from_sources",
-    "create_category_id", "add_category", "create_category_combinations",
+    "create_category_id", "add_category", "create_category_combinations", "verify_config_processes",
 ]
 
 import re
@@ -235,3 +235,30 @@ def create_category_combinations(
                     parent_cat.add_category(cat)
 
     return n_created_categories
+
+
+def verify_config_processes(config: od.Config, warn: bool = False) -> None:
+    """
+    Verifies for all datasets contained in a *config* object that the linked process is covered by
+    any process object registered in *config* and raises an exception if not. If *warn* is *True*,
+    a warning is printed instead.
+    """
+    missing_pairs = []
+    for dataset in config.datasets:
+        if not config.has_process(process := dataset.processes.get_first()):
+            missing_pairs.append((dataset, process))
+
+    # nothing to do when nothing is missing
+    if not missing_pairs:
+        return
+
+    # build the message
+    msg = f"found {len(missing_pairs)} dataset(s) whose process is not registered in the '{config.name}' config:"
+    for dataset, process in missing_pairs:
+        msg += f"\n  dataset '{dataset.name}' -> process '{process.name}'"
+
+    # warn or raise
+    if warn:
+        print(f"{law.util.colored('WARNING:', 'red')} {msg}")
+    else:
+        raise Exception(msg)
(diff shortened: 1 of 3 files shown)


columnflow/config_util.py
```diff
@@ -235,3 +235,30 @@ def kwargs_fn(categories):
...
+    # build the message
+    msg = f"found {len(missing_pairs)} dataset(s) whose process is not registered in the '{config.name}' config:"
+    for dataset, process in missing_pairs:
+        msg += f"\n  dataset '{dataset.name}' -> process '{process.name}'"
+
+    # warn or raise
+    if warn:
+        print(f"{law.util.colored('WARNING:', 'red')} {msg}")
+    else:
+        raise Exception(msg)
```
pkausw, line 264: Raising an `Exception` here is very generic. Maybe a `ValueError` might be more specific?

pkausw (changes requested):

pkausw:
Very nice PR :+1: I think at some point, we should talk about a uniform format for the doc strings, but that is another matter I guess

columnflow/config_util.py
```diff
@@ -235,3 +235,30 @@ def kwargs_fn(categories):
...
+def verify_config_processes(config: od.Config, warn: bool = False) -> None:
+    """
+    Verifies for all datasets contained in a *config* object that the linked process is covered by
+    any process object registered in *config* and raises an exception if not. If *warn* is *True*,
+    a warning is printed instead.
+    """
+    missing_pairs = []
+    for dataset in config.datasets:
+        if not config.has_process(process := dataset.processes.get_first()):
+            missing_pairs.append((dataset, process))
```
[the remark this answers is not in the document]
riga: Not right now. In `cmsdb`, we agreed to only link one process per dataset, but actually, this decision should not be reflected here. I'll generalize this to ≥ 1 process :+1:

columnflow/config_util.py
riga: Not sure if `ValueError` fits better. Maybe at some point, we should have a few (not too many) custom exceptions like `CFConfigError` or so :)
But I wouldn't tackle that in this PR.

pkausw (approved):
Bueno :+1:

PR #249, opened 2023-06-07
Limit MET pt in MET calibrator.

riga:
This PR limits the amount of MET (pT) to be passed to the correction set in the `calibration.cms.met` module.

This fix is needed since the correction set evaluation fails if the passed value is above the beam energy. This is not documented (`correction summary` prints `(-inf, inf]` for the range), but it should also be said that the number of simulated events with MET > 6500 GeV (beam energy, or half the center-of-mass energy) is extremely small so that the effect of *not* applying the calibration in these cases is negligible in total.


pkausw (approved):
Looks good to me, thanks!

PR #251, opened 2023-06-16
Updated version of documentation

pkausw:
This PR is intended to port the new version of the documentation style to the master branch.
Unfortunately, the branch was originally intended for simply updating the doc strings in the python code, which is why the PR also involves a lot of updates in multiple files. Apologies for that!

The main difference regarding the style of the documentation is the following

- I started porting the doc strings to the current recommended sphinx style in the rst format. This introduces a dedicated section in the final documentation for Parameters and return values.
- The documentation now collects parameter types and default values automatically from the signature of objects. This required an update of sphinx
- The requirements.txt file for documentation now points directly to the requirements of the individual virtual environments, which should make maintanance easier in the future.
- In the sphinx source files, this setup all works fine if TaskArrayFunctions are included with `.. autoclass:: `

There are ongoing issues that we need to think about in the future, which I'lll document here for future reference
- Due to the on-the-fly definition of TaskArrayFunctions like the centrally-provided calibrators, selectors and producers, sphinx cannot link directly to the code
- Additionally, functions that are wrapped with `init`, `setup` and `require` decorators as well as the call function are displayed with their true internal names, i.e. `init_func`, `setup_func`, `require_func` and `call_func`. This might lead to future confusion since the actual function names are different in the code itself (e.g.`jec_init` is converted to `init_func`). 

Beyond that, this PR should give us a good basis to further develop the documentation :+1:


riga (approved):
LGTM

PR #252, opened 2023-06-28
Fix software priority order

### LINKED ISSUE #250: Fix package priority in software stack
Our software stack consists of multiple layers:

1. micromamba / conda base env: Non-python packages and a handful of strictly needed Python packages (among them is NumPy)
2. cf_prod / cf_dev venv: The base stack of Python packages required to run tasks (law, luigi, pyyaml, six)
3. (**in tasks only**) custom bash envs: Everything a task needs

Naturally, software packages provided by 3 should have precedence over those in 2 and 1 since they're probably newer and usually tailored to fulfill the software requirements defined by tasks.

**However**, there are cases where this priority order is undesired. For instance, when 3 is a CMSSW sandbox, it can come with luigi, law and six preinstalled, but most likely in rather old versions, causing problems when tasks themselves use newer features. For this reason (and since there is rarely any **other** overlap between packages from 1, 2 and 3), the current software setup actually uses the order **3 < 1 < 2**.

As noted in

https://github.com/columnflow/columnflow/blob/master/sandboxes/_setup_cmssw.sh#L295-L296

this can cause issues in the above mentioned *rare* cases (esp. since NumPy comes with 1, as a result of installing gfal2-python bindings).

Therefore, the priority order should strictly be **1 < 3 < 2**.

riga:
This PR fixes the software priority order between the conda/micromamba base env, the main venvs, and task-level venvs, in the exact way proposed in #250.

## Implementation details

For sandboxes based on **venvs**, this is rather simple as each sandbox can control every aspect of its environment and python software stack. Here, the fix simply consists of arranging the PATH and PYTHONPATH variables in a way that reflects the desired order.

For sandboxes based on **cmssw**, however, things are much more complex as overlap between software packages of the outer env and cmssw itself must be resolved. In this PR, I went for completely disabling packages shipped with the conda base env as there is a huge clash between packages related to remote target interactions. As a result, gfal2 is not fully available within cmssw sandboxes. To **still** retrieve inputs from and move outputs to remote target locations, one can use law's target stage-in/-out feature. As an example, this PR changes the `cms.external.CreatePileupWeights` task to use this feature and it works nicely.

Close #250

### DIFF
--- a/columnflow/tasks/cms/external.py
+++ b/columnflow/tasks/cms/external.py
@@ -42,12 +42,24 @@ class CreatePileupWeights(ConfigTask):
     def output(self):
         return self.target(f"weights_from_{self.data_mode}.json")
 
+    def sandbox_stagein(self):
+        # stagein all inputs
+        return True
+
+    def sandbox_stageout(self):
+        # stageout all outputs
+        return True
+
     @law.decorator.log
     @law.decorator.safe_output
     def run(self):
         # prepare the external files and the output weights
         externals = self.requires()
         weights = {}
+
+        # since this tasks uses stage-in into and stage-out from the sandbox,
+        # prepare external files with the staged-in inputs
+        externals.get_files(self.input())
 
         # read the mc profile
         mc_profile = self.read_mc_profile_from_cfg(externals.files.pu.mc_profile)
--- a/columnflow/tasks/external.py
+++ b/columnflow/tasks/external.py
@@ -278,11 +278,11 @@ class BundleExternalFiles(ConfigTask, law.tasks.TransferLocalFile):
 
         return self._file_names
 
-    @property
-    def files(self):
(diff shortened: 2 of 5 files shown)


columnflow/tasks/cms/external.py
```diff
@@ -42,13 +42,25 @@ def requires(self):
     def output(self):
         return self.target(f"weights_from_{self.data_mode}.json")
```
riga, line 44: The lines below show how to use the stagein and stageout feature. One just needs to define these methods which can return masks that, when True, decide that a certain input or output should be staged. This means that the input is moved to a local, temporary location first in the outer environment, and that outputs are first stored locally and then move to the actual remote location after the sandbox terminated (hence, in the outer environment where gfal2 usually exists).

PR #254, opened 2023-07-03
Add selection utils.

riga:
This PR adds two util functions predominantly used in the selection:

- `create_collections_from_masks`: Takes the nested dictionary usually created by SelectionResults (e.g. `{"Jet": {"Jet": indices, "BJet": b_indices, ...}, ...}`) and adds new collections to an event chunk according to the passed structure. As this is done during the reduction process, the ReduceEvents task makes use of this util now.
- `sorted_indices_from_mask`: Takes a boolean mask and an array of equal shape that represents a sorting metric. The return value will be a list of sorted indices corresponding to `True`'s in the boolean mask. This is usually used in SelectionResults to refer to objects to keep in a well-defined order (see above).


columnflow/selection/util.py
```diff
@@ -0,0 +1,68 @@
...
+    return indices[mask[indices]]
+
+
+def create_collections_from_masks(
+    events: ak.Array,
+    object_masks: dict[str, dict[str, ak.Array]] | ak.Array,
+    event_mask: ak.Array | type(Ellipsis) = Ellipsis,
+) -> ak.Array:
+    """
+    """
```
dsavoiu, line 48: Could you add a minimal docstring here?

columnflow/selection/util.py
```diff
@@ -0,0 +1,68 @@
...
+        # -> [[3, 0], ...]
+    """
+    indices = ak.argsort(metric, axis=sort_axis, ascending=ascending)
+    return indices[mask[indices]]
+
+
+def create_collections_from_masks(
+    events: ak.Array,
+    object_masks: dict[str, dict[str, ak.Array]] | ak.Array,
+    event_mask: ak.Array | type(Ellipsis) = Ellipsis,
```
dsavoiu, line 45: It looks like `event_mask` is the mask that has already been applied to `events`, but still needs to be applied to the subfields `object_masks` to make them index-compatible with `events`. I'm wondering if it wouldn't make more sense to require that `events` and `object_masks` be index-compatible already. That would and avoid the indirect dependence on `event_mask` and make the function behavior easier to describe.

columnflow/tasks/reduction.py
```diff
@@ -198,18 +199,7 @@ def run(self):
...
-                    dst_names = list(sel["objects", src_name].fields)
-                    if src_name in dst_names:
-                        # move to the end
-                        dst_names.remove(src_name)
-                        dst_names.append(src_name)
-                    for dst_name in dst_names:
-                        object_mask = sel.objects[src_name, dst_name][event_mask]
-                        dst_collection = events[src_name][object_mask]
-                        events = set_ak_column(events, dst_name, dst_collection)
+                events = create_collections_from_masks(events, sel.objects, event_mask=event_mask)
```
dsavoiu, line 202: Related to the above comment on `event_mask`, it should be possible to pass `sel.objects[event_mask]` here instead of handling the `event_mask` internally in `create_collections_from_mask`.

dsavoiu (changes requested):
These utils are indeed very useful to have as separate functions. I only have a few minor points inline.

columnflow/selection/util.py
riga: Added :+1:

columnflow/selection/util.py
riga: You're right of course. There should be no need to apply an event-level mask and filtering objects is also nothing I would expect from a function called `create_collections_from_masks`.

I removed the argument and moved the filtering to `ReduceEvents` where this method is used.

columnflow/tasks/reduction.py
riga: Yup. The comment I made above applies here as well.

columnflow/tasks/reduction.py
```diff
@@ -200,18 +201,13 @@ def run(self):
...
-                        object_mask = sel.objects[src_name, dst_name][event_mask]
-                        dst_collection = events[src_name][object_mask]
-                        events = set_ak_column(events, dst_name, dst_collection)
+                # apply the event mask
+                obj_masks = sel.objects
+                for src_name in obj_masks.fields:
+                    for dst_name in obj_masks[src_name].fields:
+                        obj_mask = obj_masks[src_name, dst_name][event_mask]
+                        obj_masks = set_ak_column(obj_masks, f"{src_name}.{dst_name}", obj_mask)
+                events = create_collections_from_masks(events, obj_masks)
```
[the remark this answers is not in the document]
riga: True ... :)

dsavoiu (approved):
Thanks for the follow-up @riga! Looks like this is now ready to merge as well.

PR #257, opened 2023-07-03
Allow custom cms label postfixes in plots.

riga:
This PR updates the variable plot task so that it is possible to pass any custom CMS label postfix that wasn't registered previously.

E.g. `law run cf.PlotVariables1D --cms-label "My custom postfix"` is supported now.


mafrahm (approved):
LGTM

PR #258, opened 2023-07-03
Propagate local and global shift correctly to CSPs

riga:
A while ago when we added the concept of local and global shifts, we decided to only pass the `local_shift_inst` to CSPs (calibrators, selectors, producers) and simply name it `shift_inst` in their instance attributes.

However, after working with some less standard CSPs for the first AGC implementation, I ran into situations where it comes in handy to know the difference between global and local shifts.

In this PR, the above simplification is removed and both shifts are propagated to CSPs using their actual names. We need to document this properly in a future sprint to finalize the first draft of docs.

### DIFF
--- a/columnflow/production/cms/btag.py
+++ b/columnflow/production/cms/btag.py
@@ -108,9 +108,10 @@ def btag_weights(
         # save the new column
         return set_ak_column(events, column_name, weight, value_type=np.float32)
 
-    # when the uncertainty is a known jec shift, obtain the propagated effect and do not produce
-    # additional systematics
-    if self.shift_inst.is_nominal:
+    # when the globally requested uncertainty is a known jec shift, obtain the propagated effect and
+    # do not produce additional systematics
+    shift_inst = self.global_shift_inst
+    if shift_inst.is_nominal:
         # nominal weight and those of all method intrinsic uncertainties
         events = add_weight("central", None, "btag_weight")
         for syst_name, col_name in self.btag_uncs.items():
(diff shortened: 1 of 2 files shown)


dsavoiu (approved):
I agree that it's best to keep the nomenclature consistent between `Task` and CSP attributes, so I'm approving the changes.

But I'm not fully clear on whether this is just a name change of the CSP attribute name from `shift_inst` to `local_shift_inst`, of if there is some other change in the logic? From the PR description I would have expected `global_shift_inst` to not be accessible from the CSP body, but looking at the base task implementation it seems like this was already the case.

And actually, since the CSP methods all have access to `self.task` and therefore also `self.task.local_shift_inst` etc., could you remind me what the reason was for passing the shift instances to the CSP as instance attributes via `kwargs`?

riga:
The CSPs don't always have access to the task, for instance before any task is instantiated, their parameters are modified and resolved by means of classmethods in the task hierarchy. At this point, the task *classes* already need to talk to some CSPs, followed by the actual task instantiation.

dsavoiu:
> The CSPs don't always have access to the task, for instance before any task is instantiated, their parameters are modified and resolved by means of classmethods in the task hierarchy. At this point, the task _classes_ already need to talk to some CSPs, followed by the actual task instantiation.

Ah, right! I was focused on the CSP body itself, where `self.task` should be instantiated by the time it runs, but of course the `init` and `setup` methods may also need to know the shifts. Thanks for the reminder ;)

PR #259, opened 2023-07-03
Allow customizing default columnar sandbox

riga:
All our standard tasks that require the *usual* software packages for columnar processing currently use `bash::$CF_BASE/sandboxes/venv_columnar.sh` (with an optional dev version).

It is already possible to change this default using re-assignments on task-class level from within a personal analysis, but in case one just wants to simply exchange this sandbox globally with a custom one, this might be rather inconvenient.

This PRs adds a new config `[analysis] default_columnar_sandbox` and updates all columnar tasks to use this config value instead.


law.cfg
```diff
@@ -29,6 +29,9 @@ inference_modules: columnflow.inference
 # namespace of all columnflow tasks
 cf_task_namespace: cf
 
+# default sandbox for main tasks with standard packages for columnar processing
+default_columnar_sandbox: bash::$CF_BASE/sandboxes/venv_columnar.sh
```
dsavoiu, line 33: When are environment variables in the config actually resolved? If the `$CF_BASE` is expanded directly after the config is parsed but before being passed to the task class, it may interfere with the hash calculation. Haven't tested this, but if that's the case then a simple solution would be to simply escape the `$`.

dsavoiu (commented):
Having this in the `law` config is definitely useful for analyses wanting to completely replace the default columnar sandbox.

I just have one comment related to the environment variable substitution below.

law.cfg
riga: It's only expanded via `law.config.get_expanded()`. In the task definitions, I made sure to use `law.config.get()` instead.

dsavoiu: Ah, OK! Wasn't aware there were two `get` methods. Then I guess there is no issue :)

dsavoiu (approved):

PR #260, opened 2023-07-03
Skip calibration outputs and requirements dynamically.

riga:
Calibrators and producers, similar to selectors, can produce columns very dynamically, i.e., under certain circumstances, some columns might be added or not.

Using the fact that we can predict this dynamic behavior, this PR adds some features to make the task structure more dynamic as well.

In case a calibrator does not produce any column at all:

- `CalibratorEvents` has no outputs and is considered complete instantly. So far, the task ran, required IO and produced an empty array.
- `SelectEvents` and `ReduceEvents` do not require `CalibrateEvents`.

In case a producer does not produce any column at all:

- `ProduceColumns` has no outputs and is considered complete instantly. So far, the task ran, required IO and produced an empty array.
- `CreateHistograms`, `UniteColumns`, `PrepareMLEvents` and `MLEvaluation` do not require `ProduceColumns`.

The changes required for this are minimal :)

### DIFF
--- a/columnflow/tasks/calibration.py
+++ b/columnflow/tasks/calibration.py
@@ -56,7 +56,7 @@ class CalibrateEvents(
     def output(self):
         outputs = {}
 
-        # only declare the output in case the producer actually creates columns
+        # only declare the output in case the calibrator actually creates columns
         if self.calibrator_inst.produced_columns:
             outputs["columns"] = self.target(f"calib_{self.branch}.parquet")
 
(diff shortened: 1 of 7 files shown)


riga:
Actually, the same should be done for the `ProduceColumns` task, i.e., in case it produces no columns at all, downstream tasks should not require it.

Adding draft status for now.

dsavoiu (approved):
Very nice! Not scheduling/running these tasks when if do not produce any columns could potentially save a lot of overhead. I don't see any major issues, so I think this can be merged.

riga:
Thank you!

PR #261, opened 2023-07-04
Rename cf_prod venv to cf, improve dev_sandbox

riga:
This PR contains two changes:

- The main `cf_prod` sandbox is renamed to `cf`. This is more consistent to how other sandboxes are handled and named (e.g. `venv_columnar` and `venv_columnar_dev`).
- The `dev_sandbox` helper is a bit smarter now and can also drop the "_dev" postfix and in case the current env is not a dev env. The behavior can be controlled with arguments. From the docstring:

```
If *sandbox* does not contain the "_dev" postfix and both
:py:attr:`env_is_dev` and *add* are *True*, the postfix is appended.

If *sandbox* does (!) contain the "_dev" postfix,
:py:attr:`env_is_dev` is *False* and *remove* is *True*,
the postfix is removed.

In any other case, *sandbox* is returned unchanged.
```


columnflow/util.py
```diff
@@ -316,11 +316,21 @@ def after_call(state):
     return before_call, call, after_call
 
 
-def dev_sandbox(sandbox: str) -> str:
+def dev_sandbox(sandbox: str, add: bool = True, remove: bool = True) -> str:
     """
-    Takes a sandbox key *sandbox* and injects the subtring "_dev" right before the file extension
-    (if any) in case the current environment is used for development (see :py:attr:`env_is_dev`) and
-    the corresponding sandbox exists. Otherwise *sandbox* is returned unchanged.
+    Takes a sandbox key *sandbox* and adds or removes the subtring "_dev" right before the file
```
dsavoiu, line 321: Minor typo:
```suggestion
    Takes a sandbox key *sandbox* and adds or removes the substring "_dev" right before the file
```

dsavoiu (changes requested):
Completely in favor of renaming `cf_prod` to `cf`! I think the extra functionality in `dev_sandbox` will also be useful (I remember running into some corner cases where remote jobs were looking for a nonexistent dev sandbox, so removing the suffix dynamically could actually solve that problem).

Just one minor thing inline below, but otherwise good to go.

PR #262, opened 2023-07-04
Add skip_fn to create_category_combinations.

riga:
This PR adds an additional argument to `skip_fn` to the `create_category_combinations` helper to decide whether a category combination should be created in the first place, allowing to limit the amount of combinations to those that are actually needed.

Just like `kwargs_fn`, `skip_fn` receives the categories for the current combination (before it's added of course) in a dictionary. The combination is skipped in case it returns *True*.

@dsavoiu What auxiliary info did you add per sub category in your personal branch? Feel free to just commit to *this* branch to include that info :)

---

**Note**: *kwargs_fn* and *skip_fn* receive the exact same argument to make decisions, but *name_fn* got a different object (group names mapped to category names instead of category **instances**, forwarded as keyword arguments instead of a single dict). I changed this for consistency to use the same object.

Would you agree to this change? This might require a small change to your analyses.


columnflow/config_util.py
```diff
@@ -202,15 +205,16 @@ def kwargs_fn(categories):
             for root_cats in itertools.product(*_categories):
                 # build the name
                 root_cats = dict(zip(_group_names, root_cats))
-                cat_name = name_fn(**{
-                    group_name: cat.name
-                    for group_name, cat in root_cats.items()
-                })
+                cat_name = name_fn(root_cats)
```
riga, line 208: This is the additional change.

mafrahm (approved):
Looks good to me and I fully agree with the change regarding the `name_fn`.

Do you already have an exemplary use case of the `skip_fn`? Just for me to take a short look into that

riga:
Thanks for the review @mafrahm . I think @dsavoiu had the use case that he wasn't interested in all intermediate categories, so if four sets of leaf categories should be combined, `skip_fn` would be simply

```python
def skip_fn(categories: dict[str, od.Category]) -> bool:
    return len(categories) < 4
```

columnflow/config_util.py
```diff
@@ -202,15 +205,16 @@ def kwargs_fn(categories):
...
 
                 # skip when already existing
                 if skip_existing and config.has_category(cat_name, deep=True):
                     continue
 
+                # skip when manually triggered
+                if callable(skip_fn) and skip_fn(root_cats, cat_name):
```
dsavoiu, line 215: Since `name_fn` now received only `root_cats`, is there a reason for the additional `cat_name` being passed to `skip_fn` here?

dsavoiu (commented):
This looks good to me as well, and I also fully agree with making the signature of `name_fn` identical to `kwargs_fn`. 

I just have a question inline about the arguments passed to `skip_fn`.

columnflow/config_util.py
riga: Right ... even in the example above I skipped it ^^. Let's remove it.

PR #263, opened 2023-07-04
Refactor call caches in TaskArrayFunctions.

riga:
This PR refactors the way that TaskArrayFunction's perform caching of results of their `__call__` method.

The previous implementation simply did a call counting in in case a function was called before, the first n (needed to be configured) input arguments were returned.

Now, an actual result cache is maintained and cleared properly. The clearing is triggered by the outermost triggered task array function which triggers the clearing of its dependencies at the right moment. Also, the cache itself is fully threadsafe (values are cached per thread id and cache interactions are guarded by RLock's).

A couple other functions used a keyword argument `call_cache` which is also part of a caching mechanism which, however, has nothing to do with the actual result caching of the `call` method. I therefore renamed the argument everywhere to just `_cache` (`_` denoting that it should not be set manually).


dsavoiu:
Thanks for the implementation and sorry for the long wait! I was meaning to do some more rigorous testing for this PR since I was running into some issues with missing columns and/or columns of mismatched length that were not present in the old implementation, but didn't manage to find the time.

I think having a proper cache is a much better approach than just returning the first `n` arguments. However, I think there is a conceptual issue with the proposed implementation. To illustrate this, consider the following situation with two producers for columns `foo` and `bar`, one of which depends on the other, and a `default` producer that depends on both:

```python
@producer
def produce_foo(self, events, **kwargs):
    return set_ak_column(events, "foo", ...)

@producer
def produce_bar(self, events, **kwargs):
    events = self[produce_foo](events, **kwargs)  # call #1
    return set_ak_column(events, "bar", ...)

@producer(...)
def default(self, events, **kwargs):
    events = self[produce_bar](events, **kwargs)
    events = self[produce_foo](events, **kwargs)  # call #2
    ...
```

If one were to do `events.bar` in the `default` producer, the current implementation would raise an `AttributeError`, since the `produce_foo` call <span>#</span>2 would return the cached result from call <span>#</span>1, at which time the `bar` column had not yet been produced. Clearly this is not desirable, and `bar` should always be present in the array after `produce_bar` is called, regardless of what other producers it depends on.

In the old implementation this was avoided by directly returning `args[0]` when the producer had been called before, which would preserve the `bar` column in the example above. However, that approach also has its limitations, in that if a `TaskArrayFunction` returns more than one argument (say, `events, selection_results`, as with exposed `Selector`s), there are only returned on the first call, while any subsequent call returns only the event array. This is also not desirable because that would make the return value of `self[...](events, ...)` depend on the dependency structure in a non-transparent way.

Right now I see two possible solutions to have both some type of caching implemented, and ensure that no columns are dropped by returning old cached results:
* where there is a cache hit, run `update_ak_array(cached_result, args[0])` to ensure that columns added since the last cached call are kept; or
* don't cache the event array at all (instead return `args[0]` as before), but ensure that any subsequent arguments `args[1:]` are cached.

Both solutions require special treatment of `args[0]`, assuming it is always the event `ak.Array`. We already assume this in other places, so I think that's fine, but it might break things with simple selectors that only return an event mask. 
Other than that, the first solution could be more robust against edge cases that I did not consider here, but the `update_ak_array` method would probably introduce an overhead that might not be worth it. For that reason I'm mildly in favor of the second solution. What to you think, @riga?

riga:
Thanks for the deep dive! And I fully agree, the previous solution is perhaps a bit clumsy, and the one contained in this PR is definitely dropping columns.

Regarding your **second suggested solution**, I think not even `args[1:]` should be cached, since we don't have any control over what objects are actually passed there and they could very well have a specific state, leading to the same difficulty raised by your comment above. Hence, the second solution would boil down to the previous implementation (but correct me if I'm wrong). \*

Regarding your **first suggested solution**, I think there are edge cases which can lead to further complications. For instance, when a column `X` was removed between the instants where `cached_result` was initially cached and the array function was called a second time (triggering `update_ak_array(cached_result, args[0])`), the resulting events array will again contain column `X`.

Therefore, and in addition to the potential overhead you mentioned, I would tend towards solution 2 as well. I think the way we want the caching to work is "don't do anything with my inputs in case the task array function was called before", but of course, this entails the question of what is returned in this case.

In case we go with solution 2 (with \* to be discussed), maybe we can make the previous approach less *clumsy*. Instead of controlling the return value with the number *n_return* of input arguments, we could make this configurable with a function, e.g.

```python
@selector(
    cached_return_value=(lambda self, events, **kwargs: events, SelectionResult()),
    ...  # uses, produces, etc
)
def default(self: Selector, events: ak.Array, **kwargs) -> [ak.Array, SelectionResult]:
    ...  # implementation
```

We can provide sensible defaults for `cached_return_value` (which are probably different for producers+calibrators, selectors, and categorizers) which should apply to most user cases.

dsavoiu:
Thanks for the reply! (And again sorry for the wait.) I think we're closing in on a solution, but I there are still a few things to iron out.

Concerning the **first suggested solution**, I hadn't considered the implication of removing columns (we don't rely on this in our workflows AFAIK), but I agree we probably don't want to risk running into such edge cases. In that sense I fully agree that in general it's a good design principle here to _not touch the events array at all_ if there is a cache hit, so I think we can probably disregard this solution.

Reading your comments to the **second suggested solution**, I realize now that I had made a mistake and confused between the additional arguments supplied to the function (`args[1:]`) and its return values. What I actually meant to suggest was that, on a cache hit, the `events`/`args[0]` should be passed through without modification, but any other return values should be cached on the first call and returned from the cache on every subsequent call.

This should fix surprising failures with cached results like in the following situation:
```python
@selector
def my_selector(events, **kwargs):
    # ... implementation: `events` is modified, `result` is created
    return events, result

events, result = self[my_selector](events, **kwargs)
events, result = self[my_selector](events, **kwargs)  # fails, since it only returns `events`
```

Apart from that, there is also the question of how generic the caching implementation should be. Do we want to single out `args[0]` as the one argument to be passed through as the first return value and return the remaining return values from the cache? Or do we want to also support situations with *N* arguments and *M* return values (with a number of arguments and return values possibly overlapping)? It seems there could be use cases for these situations, which would depend on the particular selector instance. 

In light of that, I like the idea of a configurable `cached_return_value`! Assuming `my_selector` as above, and that we want to pass through `events` but return `result` from the cache, I'm guessing a sensible default implementation would look like this: 
```
def cached_return_value(self, events, **kwargs):
    cached_return_values = self._get_cached_result()
    return events, *cached_return_values[1:]
```

Does that sound like what you had in mind, @riga?

riga:
Yup, that's exactly how I pictured it, too :+1:

This way, using the example you added above,

```python
def cached_return_value(self, events, **kwargs):
    cached_return_values = self._get_cached_result()
    return events, *cached_return_values[1:]
```

the decision to use `cached_return_values[1:]` from the cache, and in particular not `*cached_return_values[0]` would be an active decision that can be made on a per-CSP basis.

Then, for this PR, the remaining task would boil down to finding the best defaults for the four types of task array functions. After checking the typical return values, I think this is even the sample implementation for all of them:

```python
return events if len(cached_return_values) < 2 else (events, *cached_return_values[1:])
```

For calibrators and producers, this mostly returns the events (`args[0]`) as is. Selectors usually return the `SelectionResult`, and categorizers a single per-event mask, so the second case should apply here.

dsavoiu (approved):
This looks great, thanks a lot! I tested this for a concrete analysis use case and it seems to work as expected. Just one minor typo I noticed in the docstring, but otherwise this looks ready to merge.

PR #265, opened 2023-07-10
Read inputs for lambda-valued variable expressions.

dsavoiu:
This PR adds a way for variables to declare input columns in case their `expression` is not a string-valued column specifier but a callable/lambda which calculates the variable value on the fly from the event array.

In such cases, the list of columns that are actually needed by the expression must be specified in a different way. The approach chosen here requires that these are given via an optional `inputs` key in the `aux` field of the variable, for instance:
```python
config.add_variable(
    name="ht",
    expression=lambda events: ak.sum(events.Jet.pt, axis=1),
    binning=(40, 0.0, 800.0),
    unit="GeV",
    x_title="$H_{T}$",
    aux={
        "inputs": {"Jet.pt"},
    },
)
```

This is intended as more of a workaround, so we can also discuss a more definitive fix. An alternative approach would be to wrap the bare `lambda` with something that allows specifying input columns, as it's being done for the CSP decorators via the `uses` keyword. However, it might take longer to understand what a better solution could be, so for the time being we should consider merging this workaround to provide early support for analyses relying on lambda-valued variable expressions.

### DIFF
--- a/columnflow/tasks/histograms.py
+++ b/columnflow/tasks/histograms.py
@@ -116,20 +116,17 @@ class CreateHistograms(
         # define columns that need to be read
         read_columns = {"category_ids", "process_id"} | set(aliases.values())
         read_columns |= {
-            Route(variable_inst.expression)
-            for variable_inst in (
-                self.config_inst.get_variable(var_name)
-                for var_name in law.util.flatten(self.variable_tuples.values())
-            )
-            if isinstance(variable_inst.expression, str)
-        } | {
-            # for variable_inst with custom expressions, read columns declared via aux key
             Route(inp)
             for variable_inst in (
                 self.config_inst.get_variable(var_name)
                 for var_name in law.util.flatten(self.variable_tuples.values())
             )
-            for inp in variable_inst.x("inputs", [])
+            for inp in (
+                [variable_inst.expression]
+                if isinstance(variable_inst.expression, str)
+                # for variable_inst with custom expressions, read columns declared via aux key
+                else variable_inst.x("inputs", [])
+            )
         }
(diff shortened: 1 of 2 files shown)


columnflow/tasks/histograms.py
```diff
@@ -117,12 +117,19 @@ def run(self):
...
             )
+            if isinstance(variable_inst.expression, str)
+        } | {
+            # for variable_inst with custom expressions, read columns declared via aux key
+            Route(inp)
+            for variable_inst in (
+                self.config_inst.get_variable(var_name)
+                for var_name in law.util.flatten(self.variable_tuples.values())
+            )
+            for inp in variable_inst.x("inputs", [])
```
riga, line 132: I'm not opposed to the double-comprehension, but it could be already worth to write a single loop over `variable_inst`'s and then fill `read_columns` in the body.

riga (changes requested):
Looks good! Just a single syntax thing :)

columnflow/tasks/histograms.py
dsavoiu: True! Running over the `variable_inst`s twice is not that nice. In the end I refactored this as a single comprehension with two for loops, but if you prefer I can also convert it to a traditional for loop.

dsavoiu:
Thanks for the review @riga! I addressed your comment above, please check if you agree with the changes.

I also realized I had forgotten to adapt the logic in the `cf.CreateCutflowHistograms` task, so I added that as well.

riga (approved):
Thank you! This looks good and will be very handy for defining variables! Let's merge.

PR #266, opened 2023-07-11
allow setting default csp via function

mafrahm:
This PR allows defining your default_{calibrator,selector,producer,ml_model} as a function that takes all task parameters as inputs.

This PR also allows setting multiple CSPs as your default_csp. For tasks that can only use one CSP, the first one is used.

Example 1: choosing the ml_model based on some default that is set in the inference_model
```python
from columnflow.inference import InferenceModel

def default_ml_model(container, **task_params):
    """ Function that chooses the default_ml_model based on the inference_model if given """
    default_ml_model = None

    # check if task is using an inference model
    if "inference_model" in task_params.keys():
        inference_model = task_params.get("inference_model", None)

        # if inference model is not set, assume it's the container default
        if inference_model in (None, "NO_STR"):
            inference_model = container.x.default_inference_model

        # get the default_ml_model from the inference_model_inst
        inference_model_inst = InferenceModel._subclasses[inference_model]
        default_ml_model = getattr(inference_model_inst, "ml_model_name", default_ml_model)

        return default_ml_model

    return default_ml_model

config_inst.x.default_ml_model = default_ml_model
```

Example 2: extend your default_producers based on which ML Model is used (using Example 1)
```python
def default_producers(container, **task_params):
    """ Default producers chosen based on the Inference model and the ML Model """

    # per default, use the ml_inputs and event_weights
    default_producers = ["ml_inputs", "event_weights"]

    # check if a ml_model has been set
    ml_model = task_params.get("mlmodel", None) or task_params.get("mlmodels", None)

    # only consider 1 ml_model
    if isinstance(ml_model, (list, tuple)):
        ml_model = ml_model[0]

    # try and get the default ml model if not set
    if ml_model in (None, "NO_STR"):
        ml_model = default_ml_model(container, **task_params)

    # if a ML model is set, add some producer that depends on the ML Model (e.g. for categorising based on ML outputs)
    if ml_model not in (None, "NO_STR"):
        default_producers.insert(0, f"ml_{ml_model}")

    return default_producers

config_inst.x.default_producer = default_producer
```

### DIFF
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -270,7 +270,7 @@ class AnalysisTask(BaseTask, law.SandboxTask):
     def resolve_config_default(
         cls,
         task_params: dict[str, Any],
-        param: str,
+        param: str | tuple[str] | None,
         container: str | od.AuxDataMixin = "config_inst",
         default_str: str | None = None,
         multiple: bool = False,
@@ -372,7 +372,8 @@ class AnalysisTask(BaseTask, law.SandboxTask):
             return param
 
         # expand default when param is empty
-        if param in (None, law.NO_STR):
+        print("d", default_str, param)
+        if param in (None, law.NO_STR, tuple()):
             param = container.x(default_str, None) if default_str else None
 
             # allow default to be a function, taking task parameters as input
@@ -391,7 +392,7 @@ class AnalysisTask(BaseTask, law.SandboxTask):
     def resolve_config_default_and_groups(
         cls,
         task_params: dict[str, Any],
-        param: tuple[str],
+        param: str | tuple[str] | None,
         container: str | od.AuxDataMixin = "config_inst",
         default_str: str | None = None,
         groups_str: str | None = None,
@@ -437,7 +438,8 @@ class AnalysisTask(BaseTask, law.SandboxTask):
             return param
 
         # expand default when param is empty
-        if param in (None, law.NO_STR):
+        print("d+g", default_str, param)
+        if param in (None, law.NO_STR, tuple()):
             param = container.x(default_str, None) if default_str else None
 
             # allow default to be a function, taking task parameters as input
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -12,7 +12,7 @@ import importlib
 import itertools
 import inspect
 import functools
-from typing import Sequence
+from typing import Sequence, Callable, Any
 
 import luigi
 import law
@@ -440,25 +440,43 @@ class ConfigTask(AnalysisTask):
     @classmethod
     def resolve_config_default(
         cls,
-        config_inst: od.Config,
+        task_params: dict[str, Any],
         param: str,
+        container: str | od.AuxDataMixin = "config_inst",
         default_str: str | None = None,
     ) -> str | None:
         """
         Applies the default value identified by *default_str* in the *config_inst*,
         if no *param* is given.
         """
+
+        # get the container inst
+        if isinstance(container, str):
+            container = task_params.get(container, None)
+
+        if not container:
+            return param
+
         # expand default when param is empty
         if not param or param == law.NO_STR:
-            param = config_inst.x(default_str, None) if default_str else None
+            param = container.x(default_str, None) if default_str else None
+
+            # allow default to be a function, taking task parameters as input
+            if isinstance(param, Callable):
+                param = param(**task_params)
+
+        # if multiple params are returned, use the first
+        if param and isinstance(param, (list, tuple)):
+            param = param[0]
 
         return param
 
     @classmethod
     def resolve_config_default_and_groups(
         cls,
-        config_inst: od.Config,
+        task_params: dict[str, Any],
         param: tuple[str],
+        container: str | od.AuxDataMixin = "config_inst",
         default_str: str | None = None,
         groups_str: str | None = None,
     ) -> tuple[str]:
@@ -467,14 +485,26 @@ class ConfigTask(AnalysisTask):
         defined in the *config_inst* that map a string to a tuple of strings.
         If *param* is empty, the default identified by *default_str* in the config is returned.
         """
+        # get the container inst
+        if isinstance(container, str):
+            container = task_params.get(container, None)
+
+        if not container:
+            return param
+
         # expand default when param is empty
         if not param or param == law.NO_STR:
-            param = config_inst.x(default_str, None) if default_str else None
+            param = container.x(default_str, None) if default_str else None
+
+            # allow default to be a function, taking task parameters as input
+            if isinstance(param, Callable):
+                param = param(**task_params)
+
             if not param:
                 return ()
 
         # expand to groups
-        if default_str and (param_groups := config_inst.x(groups_str, {})):
+        if default_str and (param_groups := container.x(groups_str, {})):
             values = []
             for val in law.util.make_list(param):
                 values.extend(param_groups.get(val, [val]))
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -372,7 +372,6 @@ class AnalysisTask(BaseTask, law.SandboxTask):
             return param
 
         # expand default when param is empty
-        print("d", default_str, param)
         if param in (None, law.NO_STR, tuple()):
             param = container.x(default_str, None) if default_str else None
 
@@ -438,7 +437,6 @@ class AnalysisTask(BaseTask, law.SandboxTask):
             return param
 
         # expand default when param is empty
-        print("d+g", default_str, param)
         if param in (None, law.NO_STR, tuple()):
             param = container.x(default_str, None) if default_str else None
 
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -463,7 +463,7 @@ class ConfigTask(AnalysisTask):
 
             # allow default to be a function, taking task parameters as input
             if isinstance(param, Callable):
-                param = param(**task_params)
+                param = param(container, **task_params)
 
         # if multiple params are returned, use the first
         if param and isinstance(param, (list, tuple)):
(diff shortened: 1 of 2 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -807,7 +809,7 @@ def resolve_param_values(cls, params: dict[str, Any]) -> dict[str, Any]:
 
             # apply ml_model_groups and default_ml_model from the config
             params["ml_models"] = cls.resolve_config_default_and_groups(
-                params["config_inst"],
+                params,
                 params.get("ml_models"),
```
riga, line 813: Can you add `container=config_inst` here as well?

columnflow/tasks/framework/base.py
```diff
@@ -440,25 +440,43 @@ def get_array_function_kwargs(cls, task=None, **params):
...
-            param = config_inst.x(default_str, None) if default_str else None
+            param = container.x(default_str, None) if default_str else None
+
+            # allow default to be a function, taking task parameters as input
+            if isinstance(param, Callable):
+                param = param(container, **task_params)
+
+        # if multiple params are returned, use the first
+        if param and isinstance(param, (list, tuple)):
+            param = param[0]
```
riga, line 470: We have cases (CalibratorsMixin, ProducersMixing (both plural)) where we might want to define a list/tuple with more than one default. Do you think it would make sense here to add a flag `multiple=False/True` or alike, to signalize that not just the first value should be returned?

Also, if multiple values are expected (`multiple=True`) and the callable just returns a single value (e.g. because the same function was used for "default_producer" and "default_producers"), one could use this info and return a 1-tuple.

columnflow/tasks/framework/mixins.py
```diff
@@ -48,13 +48,13 @@ def resolve_param_values(cls, params):
...
 
         # add the default calibrator when empty
-        if "config_inst" in params:
-            params["calibrator"] = cls.resolve_config_default(
-                params["config_inst"],
-                params.get("calibrator"),
-                default_str="default_calibrator",
-            )
-            params["calibrator_inst"] = cls.get_calibrator_inst(params["calibrator"], params)
+        params["calibrator"] = cls.resolve_config_default(
```
riga, line 51: IIRC, there were edge cases when "config_inst" was not part of params. In this case, the old behavior was that `params["calibrator"]` was not even set, and `cls.get_calibrator_inst` below wasn't invoked.

If these edge cases still exist, you would now assign `params["calibrator"] = None` and then run `cls.get_calibrator_inst` with None, whereas a sequence is expected there.

To be safe, could you do sth like

```python
if config_inst := params.get("config_inst"):
    params["calibrator"] = cls.resolve_config_default(
        ...,
        container=config_inst,
        ...
    )
    params["calibrator_inst"] = cls.get_calibrator_inst(params["calibrator"], params)
```

?

columnflow/tasks/framework/mixins.py
```diff
@@ -119,14 +119,13 @@ def get_calibrator_insts(cls, calibrators, kwargs=None):
...
-                groups_str="calibrator_groups",
-            )
-            params["calibrator_insts"] = cls.get_calibrator_insts(params["calibrators"], params)
+        params["calibrators"] = cls.resolve_config_default_and_groups(
+            params,
+            params.get("calibrators"),
+            default_str="default_calibrator",
+            groups_str="calibrator_groups",
+        )
+        params["calibrator_insts"] = cls.get_calibrator_insts(params["calibrators"], params)
```
riga, line 128: Same as above.

riga (changes requested):
Looks good and this definitely adds tons of flexibility (we need to talk about docs very soon ^^).

Just minor things inline.

columnflow/tasks/framework/mixins.py
mafrahm: sure, I can do that. I assume this is the case for selectors and producers aswell?

riga: I think so :+1:

columnflow/tasks/framework/base.py
```diff
@@ -266,6 +266,197 @@ def has_obj(name):
...
+            if isinstance(param, Callable):
+                param = param(container, **task_params)
+
+        # return either a tuple or only the first param, based on the *multiple* parameter
+        if multiple:
+            return law.util.make_tuple(param, cast=True)
+        elif param and isinstance(param, (list, tuple)):
+            return param[0]
+
+        return param
```
riga, line 388: Could you write

```python
if not param:
    return () if multiple else None

param = law.util.make_tuple(param)
return param if multiple else param[0]
```

to add proper `None`/`()` handling?

columnflow/tasks/framework/base.py
```diff
@@ -266,6 +266,197 @@ def has_obj(name):
...
+                # some inference model implementation
+                ...
+
+            params = {"config_inst": config_inst, "ml_model": None, "inference_model": "my_inference_model"}
+            resolve_param_values(params)  # sets params["ml_model"] to "my_ml_model"
+
+            params = {"config_inst": config_inst, "ml_model": "some_ml_model", "inference_model": "my_inference_model"}
+            resolve_param_values(params)  # sets params["ml_model"] to "some_ml_model"
+        """
+
```
riga, line 366: Can we add an early return here for the case that the param is indeed set to make things faster?

```python
if param not in (None, law.NO_STR, ()):
    return param
```

This would also be a bit more expressive as in "in case the parameter has a value, this function actually does not do anything".

columnflow/tasks/framework/base.py
```diff
@@ -266,6 +266,197 @@ def has_obj(name):
...
+    ) -> tuple[str]:
+        """
+        Resolves tuple of inputs *param* by using groups identified via the *groups_str* from
+        an :py:class:`order.AuxDataMixin` *container* that maps a string to a tuple of strings.
+        If *param* is empty, the default identified by *default_str* from the *container* is returned.
+        If *default_str* points to a function, it should take *container* and *task_params* as an input.
+
+        Example:
+
+        .. code-block:: python
```
riga, line 407: There needs to be an empty line after code-block::.

columnflow/tasks/framework/base.py
```diff
@@ -266,6 +266,197 @@ def has_obj(name):
...
+            def resolve_param_values(params):
```
riga, line 408: Can you simplify this example? I tried to recap what the method does, but the example code is a bit entangled. Like

```python
config_inst = ...

params = {"producer": None}

ConfigTask.resolve_config_default_and_groups(
    params,
    params.get("producer"),
    container=config_inst,
    default_str="default_producer",
    groups_str="producer_groups",
)
# -> ("features_1", "features_2", "features_3")
```

columnflow/tasks/framework/base.py
```diff
@@ -266,6 +266,197 @@ def has_obj(name):
...
+                "producer": None,
+            }
+            resolve_param_values(params)  # sets params["producers"] to ("features_1", "features_2", "features_3")
+        """
+        # get the container inst
+        if isinstance(container, str):
+            container = task_params.get(container, None)
+
+        if not container:
+            return param
```
riga, line 437:
```python
if not container:
    return law.util.make_tuple(param)
```

columnflow/tasks/framework/base.py
```diff
@@ -266,6 +266,197 @@ def has_obj(name):
...
+
+        # expand to groups
+        if default_str and (param_groups := container.x(groups_str, {})):
+            values = []
+            for val in law.util.make_list(param):
+                values.extend(param_groups.get(val, [val]))
+            return tuple(values)
+
+        # return the param as is
+        return law.util.make_tuple(param)
```
riga, line 458: Mh, the group expansion only works for one level of nesting.

```python
if not param:  # moved one indentation level to the left
    return ()

# expand groups recursively
if default_str and (param_groups := container.x(groups_str, {})):
    values = []
    lookup = law.util.make_list(param)
    handled_groups = set()
    while lookup:
        value = lookup.pop(0)
        if value in param_groups:
            if value in handled_groups:
                raise Exception(f"definition of '{group_str}' contains self-references")
            lookup = law.util.make_list(param_groups[value]) + lookup
            handled_groups.add(value)
        else:
            values.append(value)
    param = values

return law.util.make_tuple(param)
````

**Review by riga — CHANGES_REQUESTED** [review]
After going through it again, I think we can do some code simplifications and make the group resolution recursive. (all inline)

I fully agree with lifting the class methods up to `AnalysisTask` :+1:

File: columnflow/tasks/framework/base.py (hunk 6 above)
Comment (mafrahm) [reply columnflow/tasks/framework/base.py:366]: sure, but then we'd ignore the `multiple` parameter to cast the param to the correct type

File: columnflow/tasks/framework/base.py (hunk 10 above)
Comment (mafrahm) [reply columnflow/tasks/framework/base.py:458]: Not sure if we'll ever need to recursively define groups of producers, but it's implemented and works :)

File: columnflow/tasks/framework/base.py [hunk 11]
```diff
@@ -266,6 +266,208 @@ def has_obj(name):
...
+                param = param(container, **task_params)
+
+        # return either a tuple or only the first param, based on the *multiple* parameter
+        if not param:
+            return () if multiple else None
+
+        param = law.util.make_tuple(param)
+        return param if multiple else param[0]
+
+        return param
```
riga, line 391:
```suggestion
        if param in (None, law.NO_STR, tuple()):
            param = None

            # get the container inst
            if isinstance(container, str):
                container = task_params.get(container)

            # expand default when container is set
            if container and default_str:
                param = container.x(default_str, None)

                # allow default to be a function, taking task parameters as input
                if isinstance(param, Callable):
                    param = param(container, **task_params)

        # when still empty, return an empty value
        if not param:
            return () if multiple else None
        
        # return either a tuple or the first param, based on the *multiple*
        param = law.util.make_tuple(param)
        return param if multiple else param[0]
```

riga (changes requested):

columnflow/tasks/framework/base.py
riga: You're right. I added a suggestion below.

riga (approved):
LGTM!

PR #268, opened 2023-07-16
Prefer calibrator(s), selector and producer(s) parameters from cli

riga:
This PR adds small adjustments to the CSP mixins in order for them to prefer their main parameters, i.e., `--calibrator(s)`, `--selector` and `--producer(s)` from the CLI set on task-level over values passed programmatically via `.req()`.

For example, when running

```bash
law run cf.CreateHistograms --producers weights --ml-model some_model ...
```

`CreateHistograms` will tell the ML tasks to also require `--producers weights`. However, when the weights are only needed for histogramming, this might not be strictly necessary. In this case, one would want the `MLEvaluation` task to actually have a different, i.e. no `--producers`. This can be done by adding a task-level parameter

```bash
--cf.MLEvaluation-producers ''
```

To correctly distinguish between empty strings and parameters whose default should be resolved, there is a `RESOLVE_DEFAULT` placeholder now (equals to `"DEFAULT"` for usage on the command line). When a parameter receives this value, it's dynamic default is used instead.

### DIFF
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -21,9 +21,13 @@ import order as od
 from columnflow.util import DotDict
 
 
+# default analysis and config related objects
 default_analysis = law.config.get_expanded("analysis", "default_analysis")
 default_config = law.config.get_expanded("analysis", "default_config")
 default_dataset = law.config.get_expanded("analysis", "default_dataset")
+
+# placeholder to denote a default value that is resolved dynamically
+RESOLVE_DEFAULT = "DEFAULT"
 
 
 class Requirements(DotDict):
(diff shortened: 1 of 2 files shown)


riga:
@mafrahm I wanted to add you for the review of this, but this feature is needed asap for the AGC. I'm merging now, but since I know you were interested in this change, please have a look if you can.

columnflow/tasks/framework/base.py
```diff
@@ -115,23 +119,23 @@ def req_params(cls, inst, **kwargs):
...
             "retries", "acceptance", "tolerance", "parallel_jobs", "shuffle_jobs", "htcondor_cpus",
             "htcondor_gpus", "htcondor_pool",
         }
         kwargs["_prefer_cli"] = _prefer_cli
 
-        # when cls accepts a version, but non was actively requested, use the version map to assign it
+        # when cls accepts a version but none was requested use the version map to assign it
         version_map = None
         if isinstance(getattr(cls, "version", None), luigi.Parameter) and "version" not in kwargs:
             version_map = cls.get_version_map(inst)
```
mafrahm, line 132: Unrelated to the PR, but since I'm seeing this right now:, a few days ago, I tried running some tasks using the `config_inst.x.versions` aux instead of adding a version manually, but when running a task without specifying a version, the task would crash.

mafrahm (commented):
Looks good to me!

columnflow/tasks/framework/base.py
riga: The version map lookup only comes into play when a `req()` is called, and for that the triggered task still needs a version to be set.

PR #269, opened 2023-07-17
Add DeferredColumn.

riga:
This PR extends `ArrayFunction` to accept so-called `DeferredColumn`'s in the sets of used and produced columns. During the initialization of functions, these two sets are traversed and the actual requirements are determined. At this point, and before types such as IOFlagged objects, other functions, routes or strings are handled, `DeferredColumn`'s are evaluated (using their `__call__` method):

- When the return value is falsy, the object is skipped.
- When the return value is a set (i.e., nothing that the Route init would understand), the requirements are (left-)extended by the contained objects.
- When the return value is anything else, the usual handling is continued with that value.

The change is rather minimal but it allows for having dynamic adjustments of the content of `uses` and `produces`. One example is a campaign-dependent filter, that can simply subclass the basic `ArrayFunction.DeferredColumn` class

```python
class IF_NANO_V11(ArrayFunction.DeferredColumn):

    def __call__(self, func: ArrayFunction) -> Any:
        if func.config_inst.campaign.x.version == 11:
            return super().__call__(func)

        return None
```

or using a decorator to do the same (note the arguments in `super()` which are required when calling super() outside the usual class definition)

```python
@deferred_column
def IF_NANO_V11(self, func: ArrayFunction) -> Any:
    if func.config_inst.campaign.x.version == 11:
        return super(IF_NANO_V11, self).__call__(func)

    return None
```

With that, a CSP can define dynamic requirements right in the decorator:

```
@producer(
    uses={
        "common", "columns",
        IF_NANO_V11("only", "existing", "in_v11"),
    },
)
def ...
```

The init of `DeferredColumn` accepts multiple arguments, or a single set.


dsavoiu (approved):
This looks really useful and nice! I only have a minor suggestion regarding a docstring, but otherwise this is good to merge.

riga:
Great, thanks for the review!

PR #270, opened 2023-07-17
Add process scale attribute to inference model

riga:
This PR adds an additional `scale` field to process objects in the inference model.
The datacard writer uses this value to scale shapes and integrals with that factor.

This can be helpful in cases where artificial effects of different signal or background contributions should be estimated.


mafrahm (approved):
LGTM

PR #275, opened 2023-07-21
Pilot parameter to MergeHistograms and MergeShiftedHistograms

mafrahm:
The "pilot" parameter is also added to "prefer_cli", allowing to set the pilot e.g. when creating datacards:

```law run cf.CreateDatacards --version v1 --workflow htcondor --workers 4 --cf.MergeShiftedHistograms-pilot```

With this minor change, the time spent to build the dependency tree when creating datacards is reduced to a few seconds and the number of workflows that needs to be submitted is reduced significantly.
(If you prefer to submit a workflow for each systematic uncertainty, you can use the `--cf.MergeHistograms-pilot` instead)


riga:
I performed a test on this branch with the pilot parameter set and no requirements being previously request. As you already guessed, all workflows up to the heavy processing in ProduceColumns, MLTraining, MLEvaluation, etc. are bypassed, which leads to 1 htcondor job eventually run all upstreams task.

I don't think we should go with this approach, but we should follow-up on accelerating the initial task graph building.

PR #277, opened 2023-07-31
Fix ratio plotting for MC stack with empty bins

### LINKED ISSUE #276: Plotting data-mc ratio with empty MC bins leads to error
Trying to plot data-mc plots with empty MC bins and non-empty data bins leads to a negative `yerr`:
```
ValueError: 'yerr' must not contain negative values
```

mafrahm:
This PR sets all negative `yerr` values to 0 in the `plot_errorbars` function.

Closes #276

### DIFF
--- a/columnflow/plotting/plot_all.py
+++ b/columnflow/plotting/plot_all.py
@@ -20,7 +20,12 @@ mplhep = maybe_import("mplhep")
 od = maybe_import("order")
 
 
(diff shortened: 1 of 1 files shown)


dsavoiu (approved):
Thanks for the hotfix @mafrahm. This should indeed solve issue #276 (empty MC bins, corresponding data bins not empty), so I think we can go ahead and merge to restore plotting functionality for this edge case.

PR #280, opened 2023-08-09
use true number of pileup interactions in pileup producer

mafrahm:
It seems that in the `pu_weight` producer, the wrong column (`PV.npvs`) has been used to determine the PU event weights.

The mc_profile that is used to determine the PU weights [1] only contains values up to a number of Pileup of 73, which leads to the PU weights shown in [2].

In the PU producer, we are using these weights to determine the PU event weights based on the number of reconstructed primary vertices (`PV.npvs`) [3], however, this column takes values above 73, which leads to both large event weights of159 and event weights of 0.
Based on the information found in [4], I think that the correct column to determine the event weights would be `Pileup.nTrueInt`. I also took a look into the NanoAODs and it seems that for the UL samples, the max. value of this column seems to be around ~< 73, which would be in agreement with the mc_profile we have seen in [1].

[1] https://github.com/columnflow/columnflow/blob/3a69281abeefc614986b89f0a7601e7a45df5888/columnflow/tasks/cms/external.py#L65
(mc profile from here: https://raw.githubusercontent.com/cms-sw/cmssw/435f0b04c0e318c1036a6b95eb169181bbbe8344/SimGeneral/MixingModule/python/mix_2017_25ns_UltraLegacy_PoissonOOTPU_cfi.py)
[2] Small section of Number of Pileup vs. nominal PU weight:
```
 (66, 1.7605077201696504),
 (67, 2.4973721771534723),
 (68, 3.604846391289538),
 (69, 3.510370430997905),
 (70, 5.546157726604674),
 (71, 16.473486457289216),
 (72, 30.20828844763221),
 (73, 159.5512994439475),
 (74, 0.0),
 (75, 0.0),
 (76, 0.0),
 ```
 [3] https://github.com/columnflow/columnflow/blob/3a69281abeefc614986b89f0a7601e7a45df5888/columnflow/production/cms/pileup.py#L34
 [4] https://twiki.cern.ch/twiki/bin/viewauth/CMS/PileupJSONFileforData

### DIFF
--- a/columnflow/production/cms/pileup.py
+++ b/columnflow/production/cms/pileup.py
@@ -31,7 +31,7 @@ def pu_weight(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
     functions below.
     """
     # compute the indices for looking up weights
-    indices = events.Pileup.nTrueInt.to_numpy() - 1
+    indices = events.Pileup.nTrueInt.to_numpy().astype("int32") - 1
     max_bin = len(self.pu_weights) - 1
     indices[indices > max_bin] = max_bin
 
--- a/columnflow/production/cms/pileup.py
+++ b/columnflow/production/cms/pileup.py
@@ -19,7 +19,7 @@ set_ak_column_f32 = functools.partial(set_ak_column, value_type=np.float32)
 
 
 @producer(
-    uses={"PV.npvs"},
+    uses={"Pileup.nTrueInt"},
     produces={"pu_weight", "pu_weight_minbias_xs_up", "pu_weight_minbias_xs_down"},
     # only run on mc
     mc_only=True,
@@ -31,7 +31,7 @@ def pu_weight(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
     functions below.
     """
     # compute the indices for looking up weights
-    indices = events.PV.npvs.to_numpy() - 1
+    indices = events.Pileup.nTrueInt.to_numpy() - 1
     max_bin = len(self.pu_weights) - 1
     indices[indices > max_bin] = max_bin
 


dsavoiu (approved):
Hi @mafrahm, good catch! To my knowledge, the variable most commonly used for pileup reweighting is indeed `nTrueInt` (also known as `nPUMean` or just _μ_ in some contexts), which is the mean value of the Poisson distribution used to sample the actual number of pileup interactions mixed into the event.

A reweighting based on `PV.npvs` (i.e. the number of primary vertices) is also possible, since this variable roughly scales with pileup, but one disadvantage of that is that it relies on tracking information, so it's not a measure of pileup alone. If one want to use that, the PU profile histograms contained in the external files used by the `cf.CreatePileupWeights` task of course have to be computed using the appropriate variable.

Apparently `pu_weight` producer makes the assumption that `PV.npvs` was used for the profiles, but the PU profile files seem to contain histograms based on `Pileup.nTrueInt`, leading to the issue you are seeing.

The choice of variable depends on the analysis. However, I think in all our analyses the reweighting is done based on `Pileup.nTrueInt`, so I agree that this PR should be merged ASAP.

We should also check with @riga on the off-chance that the hh2bbtautau uses a different reweighting scheme. In that case we would need to avoid hard-coding the variable name in the producer, but that is best handled in a new PR and shouldn't prevent this from being merged.

mafrahm:
HI @dsavoiu, thank you for the quick review!
I took a quick look into the config form the hh2bbtautau analysis and the MC PU profile seems to be the same as I linked in my PR ([1]), so I think we're save to merge.

[1] https://github.com/uhh-cms/hh2bbtautau/blob/7666ed0426c87baa8d143ec26a216c3cadde513b/hbt/config/configs_run2ul.py#L612C33-L612C197

dsavoiu:
OK, thanks for checking! Then I also think we should go ahead with the merge, and if needed we can follow up with @riga once he's back.

riga:
This looks good, thanks for the fix!

On the hh2bbtautau side we only have to adjust the renormalization, so we're good with the upstream change.

PR #281, opened 2023-08-11
Fix/nan pdf weights

mafrahm:
This PR includes a fix in the PDF weight producer for datasets that include zero-entries in the nominal LHEPdfWeight by setting the produced pdf_weight to 0 for such cases.

We should also discuss how to handle cases where the stddev becomes too large. I implemented a warning when the variations are above 50%.


columnflow/production/cms/pdf.py
```diff
@@ -74,4 +74,24 @@ def pdf_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
     events = set_ak_column_f32(events, "pdf_weight_up", 1 + stddev)
     events = set_ak_column_f32(events, "pdf_weight_down", 1 - stddev)
 
+    if ak.any(stddev > 0.5):
+        # catch events with large pdf variations
+        # TODO: how should we handle large PDF weight uncertainties (especially above 100%)?
+        logger.warning(
+            f"In dataset {self.dataset_inst.name}, there are {ak.sum(stddev > 0.5)} " +
+            "entries with pdf uncertainty above 50%",
+        )
```
pkausw, line 83: I like this summary very much, but it might be worthwhile to know the fraction of occurances w.r.t. total amount of events, like

 ```suggestion
        occurances = ak.sum(stddev>0.5)
        frac = occurances/ak.count(stddev, axis=0) * 100
        logger.warning(
            f"In dataset {self.dataset_inst.name}, there are {occurances} ({frac:.2f}%) " +
            "entries with pdf uncertainty above 50%",
        )
```

**Review by pkausw — CHANGES_REQUESTED** [review]
LGTM, just one very minor comment

File: columnflow/production/cms/pdf.py (hunk 1 above)
Comment (pkausw) [reply columnflow/production/cms/pdf.py:83]: Regarding what we should do in case the std dev is very large or the nominal weight is 0: As far as I'm aware, there isn't one global way to go about this, because it shouldn't happen in theory. Since the producer is part of the cf backend, we could think about exposing options about what to do to the user, e.g.

- throw a fatal error and displace the sample (or even file?) and the number of occurances
- just display the warning
- discard the events that are problematic

This is obviously up for debate.

File: columnflow/production/cms/pdf.py [hunk 2]
```diff
@@ -74,4 +74,26 @@ def pdf_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
     events = set_ak_column_f32(events, "pdf_weight_up", 1 + stddev)
     events = set_ak_column_f32(events, "pdf_weight_down", 1 - stddev)
 
+    if ak.any(stddev > 0.5):
```
riga, line 77: Could you store `stddev > 0.5` as e.g. `outlier_mask = stddev > 0.5` and use it in L80?

columnflow/production/cms/pdf.py
```diff
@@ -74,4 +74,26 @@ def pdf_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
...
+        # catch events with large pdf variations
+        # TODO: how should we handle large PDF weight uncertainties (especially above 100%)?
+        occurances = ak.sum(stddev > 0.5)
+        frac = occurances / ak.count(stddev, axis=0) * 100
+        logger.warning(
+            f"In dataset {self.dataset_inst.name}, there are {occurances} ({frac:.2f}%) " +
+            "entries with pdf uncertainty above 50%",
+        )
+
+    if ak.any(pdf_weight_nominal == 0):
```
riga, line 87: Same as above, (e.g.) `zero_mask = pdf_weight_nominal == 0`.

From the PR title I understand that nan's are an issue at some point. Where do they appear? I'm wondering if setting everything (nominal + variations) to zero helps.

columnflow/production/cms/pdf.py
```diff
@@ -74,4 +74,26 @@ def pdf_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
...
+        # set all pdf weights to 0 when the nominal pdf weight is 0
+        invalid_pdf_weight = (pdf_weight_nominal == 0)
+        logger.warning(
+            f"In dataset {self.dataset_inst.name}, {ak.sum(invalid_pdf_weight)} nominal " +
```
riga, line 91: Trailing +.

riga (changes requested):

columnflow/production/cms/pdf.py
mafrahm: They appear when the 0th pdf weight entry (pdf_weight_nominal) is 0, on which we normalize here: https://github.com/columnflow/columnflow/blob/17d2d12b699cd94d0e554692f3b3fc3e803b7d79/columnflow/production/cms/pdf.py#L66

riga (approved):
LGTM!

mafrahm:
I also just added three parameters, `outlier_mode` and `outlier_message_type` to customize how to handle outliers above a threshold `outlier_threshold`

columnflow/production/cms/pdf.py
```diff
@@ -30,24 +30,49 @@
...
+        - ``"remove"``: pdf weight nominal/up/down are all set to 0, essentially removing the event
+
+    The *outlier_message_type* allows to define the type of message that is given for datasets with outliers:
+
+        - ``"ignore"``: no message is given
+        - ``"info"``: a `logger.info` message is given
+        - ``"debug"``: a `logger.debug` message is given
+        - ``"warning"``: a `logger.warning` message is given
+        - ``"raise"``: an Exception is raised
+
```
riga, line 60: I think outlier mode and message type have some overlap in their meaning, e.g. `raise` should be a mode.

How about

```
The *outlier_mode* defines the procedure of how to handle events with a pdf uncertainty above the *outlier_threshold*. Supported modes are:

    - ``"ignore"``: events are kept unmodified, no message is printed
    - ``"warning"``: events are kept unmodified, a warning is printed
    - ``"remove"``: pdf weight nominal/up/down are all set to 0, a warning is printed  
    - ``"raise"``: an exception is raised
```

This seems a bit more streamlined to me.

riga (changes requested):

columnflow/production/cms/pdf.py
mafrahm: In my opinion, these are two separate things we might want to customize, one parameter for verbosity and one for functionality (how to handle the outlier events). With your suggestion, there would be no option to "remove" the outlier events but also print no message.

pkausw: Just to restart the conversation in this thread: I agree with @mafrahm, the handling of the events and the logging/information for the users can (and maybe should) be to separate things. Maybe just use one `outlier_log_mode` and an `outlier_action` as input parameters? So something like this:

```
The *outlier_action* defines the procedure of how to handle events with a pdf uncertainty above the *outlier_threshold*. Supported modes are:

    - ``"ignore"``: events are kept unmodified
    - ``"remove"``: pdf weight nominal/up/down are all set to 0, a warning is printed  
    - ``"raise"``: an exception is raised
    
 Additionally, the verbosity of the procedure can be set with *outlier_log_mode*, which offers the following options:

    - ``"info"``: a `logger.info` message is given
    - ``"debug"``: a `logger.debug` message is given
    - ``"warning"``: a `logger.warning` message is given

```

This way we could have silent handling of outlier weights, or a more verbose documentation of such incidents in different logger streams. What do you think @mafrahm @riga ?

mafrahm: Hi @pkausw, I agree with your proposal, with two exceptions:
- `outlier_action = remove` should not print a warning per default but based on the *outlier_log_mode*
- the `outlier_log_mode` should offer a mode to completely ignore the message (e.g. None or False)

pkausw: sounds good to me :+1:

riga: To me as well :+1:

columnflow/production/cms/pdf.py
```diff
@@ -30,24 +30,55 @@
...
        - https://arxiv.org/pdf/1510.03865.pdf
     """
+
+    if not outlier_action:
+        outlier_action = "ignore"
+    if not outlier_log_mode:
+        outlier_log_mode = "none"
+
+    assert outlier_action in ("ignore", "remove", "raise")
+    assert outlier_log_mode in ("none", "info", "debug", "warning")
```
riga, line 73: Just some last thing: we should remove all `assert`'s from our code base (as done 2a76125).

```suggestion
    if outlier_action not in (known_actions := ("ignore", "remove", "raise")):
        raise ValueError(
            f"unknown outlier_action '{outlier_action}', known values are {','.join(known)}",
        )
    if outlier_log_mode not in (known_log_modes := ("none", "info", "debug", "warning")):
        raise ValueError(
            f"unknown outlier_log_mode '{outlier_log_mode}', known values are "
            f"{','.join(known_log_modes)}",
        )
```

riga (changes requested):
Just one last suggestion to get rid of the `assert`'s, then we could merge.

riga (approved):

columnflow/production/cms/pdf.py
```diff
@@ -30,24 +30,62 @@
...
+
     Resources:
 
        - https://arxiv.org/pdf/1510.03865.pdf
     """
+
+    if not outlier_action:
+        outlier_action = "ignore"
+    if not outlier_log_mode:
+        outlier_log_mode = "none"
```
pkausw, line 70: Is this necessary? The defaults are already set in the function, right?

pkausw (commented):
Generally looks good to me, thanks! Just one super minor comment, whether we change this now or not isn't critical imho. If you disagree, feel free to just mark this as done and go ahead with the merge :+1:

columnflow/production/cms/pdf.py
mafrahm: Not really, I thought it might make sense to just allow something like False, None and empty strings, but you're probably right and we should just stick with the given (and documented) options

pkausw (approved):
LGTM, thanks! :+1:

pkausw (approved):

PR #282, opened 2023-08-18
Make category id columns configurable.

riga:
This is a rather short PR that makes the columns used to extract category ids in the `CreateHistograms` task configurable. Also, multiple columns are accepted now that are concatenated along the category id axis.

We could also make this a parameter or config value, but for now I think it's ok to keep it as a class member.


mafrahm (approved):
LGTM

PR #285, opened 2023-08-18
Add Categorizer object.

riga:
This PR adds a dedicated `Categorizer` object that is meant to produce masks for defining categories. In particular, the changes are:

- Added the new `Categorizer` as a `TaskArrayFunction` subclass.
- Added the `exposed` attribute to all CSPs and added exceptions to the mixins in case they unexposed CSPs are requested from the CLI. This was only done for selectors before but is a good generic approach.
- Categorizers are supposed to return `(events, mask)`.
- All existing core selectors now return `(events, results)`.
- In the `category_ids` producer the new categorizer object is used instead of selectors.
- Template analysis:
  - Use categorizes in category definitions.
  - Updated the example selectors to always return `(events, results)`.
- Updated the `create_analysis.sh` script for easier local debugging.

### DIFF
--- a/columnflow/production/categories.py
+++ b/columnflow/production/categories.py
@@ -10,7 +10,7 @@ from collections import defaultdict
 
 import law
 
-from columnflow.selection import Selector
+from columnflow.categorization import Categorizer
 from columnflow.production import Producer, producer
 from columnflow.util import maybe_import
 from columnflow.columnar_util import set_ak_column
@@ -36,17 +36,12 @@ def category_ids(
 
     for cat_inst in self.config_inst.get_leaf_categories():
         # start with a true mask
-        cat_mask = np.ones(len(events)) > 0
+        cat_mask = np.ones(len(events), dtype=bool)
 
         # loop through selectors
-        for selector in self.category_to_selectors[cat_inst]:
-            # run the selector for events that still match the mask, then AND concat
-            _cat_mask = self[selector](events[cat_mask], **kwargs)
-            cat_mask[cat_mask] &= np.asarray(_cat_mask == 1)
-
-            # stop if no events are left
-            if not ak.any(cat_mask):
-                break
+        for categorizer in self.categorizer_map[cat_inst]:
+            events, mask = self[categorizer](events, **kwargs)
+            cat_mask = cat_mask & mask
 
         # covert to nullable array with the category ids or none, then apply ak.singletons
         ids = ak.where(cat_mask, np.float64(cat_inst.id), np.float64(np.nan))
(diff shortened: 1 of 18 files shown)


columnflow/production/categories.py
```diff
@@ -65,33 +60,32 @@ def category_ids(
...
-            if selector.exposed:
-                logger.warning(
-                    f"selection of category {cat_inst.name} seems to refer to an exposed selector "
-                    "whose return value is most likely incompatible with category masks",
+            # the categorizer must be exposed
+            if not categorizer.exposed:
+                raise RuntimeError(
+                    f"cannot use unexposed categorizer '{categorizer}' to evaluate category "
+                    f"{cat_inst}",
                 )
```
pkausw, line 85: Just for clarification, this means that we also plan to support hidden categorizers, but what's the use case for that? Sth like a "super" categorizer that consists of smaller ones? Not opposed to that idea, but since they only provide boolean masks and can't easily provide additional information (like selectors can), I don't think this is strictly necessary, right?

columnflow/selection/cms/json_filter.py
```diff
@@ -43,9 +43,9 @@ def get_lumi_file_default(self, external_files: DotDict) -> str:
 def json_filter(
     self: Selector,
     events: ak.Array,
-    data_only=True,
+    data_only: bool = True,
     **kwargs,
-) -> ak.Array:
+) -> tuple[ak.Array, SelectionResult]:
```
pkausw, line 48: Can you also add the explanation in the `:return:` command in the doc string?

columnflow/selection/cms/met_filters.py
```diff
@@ -83,7 +83,7 @@ def met_filters(
         # append to the result
         result = (result & vals)
 
-    return result
+    return result, SelectionResult()
```
pkausw, line 86: shouldn't this be the ak array itself? Result looks to be a boolean mask (which granted, is an ak Array itself, but it's fundamentally different than in other selectors, right?

```suggestion
    return result, SelectionResult(
        {
            "steps": {
                "met_filters": result,
            }
        }
    )
```

columnflow/selection/cms/met_filters.py
```diff
@@ -40,7 +40,7 @@ def met_filters(
     self: Selector,
     events: ak.Array,
     **kwargs,
-) -> ak.Array:
+) -> tuple[ak.Array, SelectionResult]:
```
pkausw, line 43: Similar comment as before, can you update the doc string description?

columnflow/selection/matching.py
```diff
@@ -165,4 +165,4 @@ def jet_lepton_delta_r_cleaning(
     clean_jet_indices = self[delta_r_jet_lepton](events, "Jet", ["Muon", "Electron"], threshold=threshold)
 
     # TODO: should not return a new object collection but an array with masks
-    return SelectionResult(objects={"Jet": clean_jet_indices})
+    return events, SelectionResult(objects={"Jet": clean_jet_indices})
```
pkausw, line 168: Update doc string?

pkausw (changes requested):
Nice PR, thanks @riga ! Just some minor comments from my side

columnflow/production/categories.py
riga: In general, I think it's good to have the option to mark a selector / producer / calibrator as "not exposed", i.e., an error will show up when someone tries to use them from the command line.

My thought behind adding the same for categorizers was a) consistency, and b) although they're not used from the command line, but by `od.Category` objects, it could be useful to mark certain categorizers as "do not use in categories, but call from another function". This could be the case when a categorizer has additional arguments, that cannot be automatically known by `od.Category`'s.

columnflow/selection/cms/json_filter.py
riga: Done 👍

columnflow/selection/cms/met_filters.py
riga: Yup, you're a right. I did the same also for the json filter selector above.

columnflow/selection/matching.py
riga: Added 👍

columnflow/production/categories.py
pkausw: OK sounds good to me :+1:

pkausw (approved):

riga:
Thanks!

Since this one contains core changes, maybe we still wait for input from @mafrahm .

mafrahm (approved):
I only took a rather short look into the changes, but from my side you can merge

PR #286, opened 2023-08-18
Add checks for overlapping input columns.

### LINKED ISSUE #283: Raise awareness for priority of producers
We've had an issue with our producer setup in our ``HH->bbWW`` analysis, which could happen in other analyses as well, if one isn't aware of the priority of producers. In our case we wanted to create datacards for different categories based on the output scores of the machine learning process classification. For this task more than one producer was called, that produced category_ids. The latter producer overwrites the column for the category_ids, so that in our case the categorization, that was needed for the inference-task, wasn't there anymore, which lead to datacards with no events.
So if one uses different producers for a task (e.g. CreateHistograms), that produce a column for a variable with the same name, one might run into some issues if the order of the given producers for the task is not right. In this case the latter producer gets prioritized and overwrites the column, that is being produced by both producers. One needs to be aware of this when one sets the producers for the given task.

Solution approach:

Raise a warning if one uses more than one producer, that produce a column for the same variable. The warning should include, that the user should be aware, that the latter producer gets prioritized and overwrites the column output of the previous producer. One might add different modes, where one could ignore this warning, if the user is aware of this producer behaviour.

riga:
This PR adds checks for overlapping input columns in several tasks that read more than one input file (as suggested by @jolange).

The check can be enabled / disabled by adding the respective task family to the `check_overlapping_inputs` config value, similar to the `check_finite_output` mechanism. When one or more overlapping columns are detected across at least two arrays, a `ValueError` is raised. In some cases a warning could be a better choice but this should be done in a future PR.

The checks in all tasks work similar in that overlapping columns are detected across all input files, i.e., original events and addition columns. Two exceptions are `SelectEvents` and `ReduceEvents` which do not consider the original events but only additional columns as the latter are meant to overwrite columns existing in nano input files.

Closes #283.

### DIFF
--- a/analysis_templates/cms_minimal/law.cfg
+++ b/analysis_templates/cms_minimal/law.cfg
@@ -52,10 +52,12 @@ chunked_io_pool_size: 2
 chunked_io_debug: False
 
 # csv list of task families that inherit from ChunkedReaderMixin and whose output arrays should be
-# checked for non-finite values before saving them to disk (right now, supported tasks are
-# cf.CalibrateEvents, cf.SelectEvents, cf.ProduceColumns, cf.PrepareMLEvents, cf.MLEvaluation,
-# cf.UniteColumns)
+# checked (raising an exception) for non-finite values before saving them to disk
 check_finite_output: cf.CalibrateEvents, cf.SelectEvents, cf.ProduceColumns
+
+# csv list of task families that inherit from ChunkedReaderMixin and whose input columns should be
+# checked (raising an exception) for overlaps between fields when created a merged input array
+check_overlapping_inputs: None
 
 # whether to log runtimes of array functions by default
 log_array_function_runtime: False
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import gc
 import time
 import itertools
+from collections import Counter
 from typing import Sequence, Any
 
 import luigi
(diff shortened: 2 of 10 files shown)


analysis_templates/cms_minimal/law.cfg
```diff
@@ -52,11 +52,13 @@ chunked_io_pool_size: 2
 chunked_io_debug: False
 
 # csv list of task families that inherit from ChunkedReaderMixin and whose output arrays should be
-# checked for non-finite values before saving them to disk (right now, supported tasks are
-# cf.CalibrateEvents, cf.SelectEvents, cf.ProduceColumns, cf.PrepareMLEvents, cf.MLEvaluation,
-# cf.UniteColumns)
+# checked (raising an exception) for non-finite values before saving them to disk
```
pkausw, line 55: Since you changed the comment, does this mean that the finite value check is now available in all tasks?

columnflow/tasks/framework/mixins.py
```diff
@@ -1390,3 +1395,28 @@ def raise_if_not_finite(cls, ak_array: ak.Array) -> None:
...
+
+        # determine overlapping routes
+        counts = Counter(sum(map(get_ak_routes, ak_arrays), []))
+        overlapping_routes = [r for r, c in counts.items() if c > 1]
+
+        # raise
+        if overlapping_routes:
+            raise ValueError(
+                f"found {len(overlapping_routes)} overlapping columns across {len(ak_arrays)} "
+                f"arrays: {','.join(overlapping_routes)}",
```
pkausw, line 1421: This isn't really `arrays`, but more `fields` or `routes` or `columns` right? So it might be more appropriate to write sth like

```suggestion
                f"columns: {','.join(overlapping_routes)}",
```

law.cfg
```diff
@@ -47,11 +47,13 @@ chunked_io_pool_size: 2
 chunked_io_debug: False
 
 # csv list of task families that inherit from ChunkedReaderMixin and whose output arrays should be
-# checked for non-finite values before saving them to disk (right now, supported tasks are
-# cf.CalibrateEvents, cf.SelectEvents, cf.ProduceColumns, cf.PrepareMLEvents, cf.MLEvaluation,
-# cf.UniteColumns)
+# checked (raising an exception) for non-finite values before saving them to disk
```
pkausw, line 50: Similar comment as in `cms_minimal` law config

pkausw (changes requested):
Nice PR, just some small questions/suggestions :+1:

analysis_templates/cms_minimal/law.cfg
riga: Yes, all tasks that produce columns have this check enabled now, so I wanted to remove the list in order for analyses, that started from the example at some point in the past, not to have an outdated list of tasks in their main cfg.

law.cfg
riga: Here we should indeed keep list of tasks. Added back :+1:

columnflow/tasks/framework/mixins.py
riga: Yup! Added in the same commit as the change below.

pkausw (approved):
LGTM :+1:

PR #287, opened 2023-08-19
Cache bundle repo requirement in HTCondorWorkflow.

### LINKED ISSUE #279: Modifying local repo files interferes with remote jobs
We are seeing an issue in our `mttbar` analysis setup, with jobs submitted to the _HTCondor_ batch system beginning to fail whenever any file in the local repository is modified. I'm describing our workflow below to illustrate the problem

We first submit a task to the batch system with `--workflow htcondor` (we also set `htcondor_share_software = True` in the law config, which may be relevant here). The `cf.BundleRepo` task is then scheduled, runs, and uploads an archive of the repository to the configured storage, and the actual taks scheduling and job submision then goes on as expected:
```
$ law run cf.ProduceColumns ... --workflow htcondor

INFO: luigi-interface - [pid 40097] Worker Worker(salt=8640385522, workers=1, host=naf-cms16.desy.de, username=<redacted>, pid=40097) running   cf.BundleRepo(include_files=, custom_checksum=NO_STR, source_path=NO_STR, analysis=mtt.config.analysis_mtt.analysis_mtt, replicas=5)

uploaded mttbar.c0d3955abbc7876c70bd12f438321ae57832def3.0.tgz
[...other replicas...]

going to submit 1 htcondor job(s), run2_2017_nano_v9_limited, st_tchannel_tbar_powheg
submitted 1/1 job(s)
submitted 1 job(s), run2_2017_nano_v9_limited, st_tchannel_tbar_powheg
18:59:23: all: 1, pending: 0 (+0), running: 1 (+1), finished: 0 (+0), retry: 0 (+0), failed: 0 (+0), run2_2017_nano_v9_limited, st_tchannel_tbar_powheg
```

If no further action is taken, the jobs finish successfully. However, if at any point a file in the analysis repository is modified, the jobs start failing. This can be any file, in particular also a file that is not tracked by git. The task status message reads:
```
19:00:24: all: 1, pending: 0 (+0), running: 0 (-1), finished: 0 (+0), retry: 1 (+1), failed: 0 (+0), run2_2017_nano_v9_limited, st_tchannel_tbar_powheg
1 failed job(s) in task cf.ProduceColumns_mtt_config_analy__1__87146dcf1d:
    job: 1, branch(es): [0], id: 12781840.0, status: retry, code: 20, error: undefined, job script error: bootstrap file failed,
    log: <path_redacted>/data/mtt_store/analysis_mtt/cf.ProduceColumns/run2_2017_nano_v9_limited/st_tchannel_tbar_powheg/nominal/calib__skip_jecunc/sel__default/prod__default/v2/stdall_0To1.txt
```

And the log file contains the following error message:
```
-- setup -------------------------------------------------------------------------------------------

-- bootstrap file
run bootstrap file 'remote_bootstrap_3c46256fbf.sh'
found existing software at <path_redacted>/data/software
law_wlcg_get_file: could not determine file to load from '<path_redacted>/data/mtt_store/analysis_mtt/cf.BundleRepo'
with file name 'mttbar.90a067047cf104212683d528ae8cfae7f7f3e4d3.[^\.]+.tgz'
bootstrap file failed with code 1, stop job
...

-- end ---------------------------------------------------------------------------------------------
...
job exit code : 20
```

Note the mismatched repo hash at submission time (`c0d3955`) and execution time (`90a0670`).

Is this expected behavior? It seems to me like this would negatively affect users developing code while running jobs out of the same repository, which should be a common enough use case.

I would also argue that some sort of persistence mechanism is needed so the bootstrap script running on the remote does not rely on the current state of the repository (even with `htcondor_share_software = True`) to calculate the hash for the repo image, but I am not familiar enough with the bootstrap script/bundling mechanism to understand how this could be implemented. Maybe @riga can help here?

As a partial fix, one could at least exclude untracked files from the checksum calculation. I'm actually a bit puzzled about the reasoning to include them in the first place, since as far as I can tell, only tracked files are actually added to the repo image, so the hash would change even if the content does not. This seems counterintuitive to me, but it's entirely possible that I'm missing something.

riga:
This PR adds a small adjustment to the base `HTCondorWorkflow` to avoid race conditions regarding the checksum calculation of the `BundleRepo` task, as reported in #279.

The issue appears when, during the processing of tasks required to submit jobs via htcondor,  the `BundleRepo` requirement is run way before the actual `HTCondorWorkflow` is triggered. So far, the latter created an instance of `BundleRepo` every time the workflow requirements were created, i.e., during dependency tree building and then again within the workflow run method itself to access the repo checksum to be sent with jobs, 

https://github.com/columnflow/columnflow/blob/a812d032db9e753de886e92e86d34baed7eec341/columnflow/tasks/framework/remote.py#L449

When a change to the repository happens between these to instants, the checksum will be different, and a wrong (i.e., not actually provided by `BundleRepo`) checksum will be sent with jobs, which fail to find the bundle.

The fix simply consists of a cached `BundleRepo` instance that is used in both cases.

Fixes #279.

### DIFF
--- a/columnflow/tasks/framework/remote.py
+++ b/columnflow/tasks/framework/remote.py
@@ -343,11 +343,18 @@ class HTCondorWorkflow(AnalysisTask, law.htcondor.HTCondorWorkflow):
         BundleCMSSWSandbox=BundleCMSSWSandbox,
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # cached BundleRepo requirement to avoid race conditions during checksum calculation
+        self.bundle_repo_req = self.reqs.BundleRepo.req(self)
+
     def htcondor_workflow_requires(self):
         reqs = law.htcondor.HTCondorWorkflow.htcondor_workflow_requires(self)
 
-        # add the repository bundle
-        reqs["repo"] = self.reqs.BundleRepo.req(self)
+        # add the repository bundle and trigger the checksum calculation
+        reqs["repo"] = self.bundle_repo_req
+        self.bundle_repo_req.checksum
 
         # main software stack
         if not self.htcondor_share_software:


pkausw (approved):
Good to go :+1:

PR #288, opened 2023-08-19
Fix venv stacking by ignoring existing deps.

### LINKED ISSUE #278: Empty `cf` sandbox
After moving to a more recent version of Columnflow in our analysis setup, we began seeing an issue with the `cf` "production" sandbox, in that it does not seem to contain any of the packages specified in `requirements_prod.txt` (actually, it is completely empty aside from a minimal venv directory structure). This leads to failures of remote jobs, which rely on the production `cf` sandbox to run.

I suspect the reason for the empty sandbox has to do with the fact that the `requirements_prod.txt` is a subset of the packages installed in the `cf_dev` sandbox, which also contains development tools like IPython, and which is sourced by the analysis setup script. When triggering a build of `cf` from the sourced setup, nothing will be installed in `cf`, since all requirements are already found in the already-sourced `cf_dev` sandbox.

If that is indeed the case, an easy fix would be to run `pip --ignore-installed -r requirements_prod.txt` for `cf`, so the packages potentially found in `cf_dev` are not taken into account and therefore installed explicitly. However, this might also lead to packages being unnecessarily reinstalled. I'm wondering why we did not see this before, since in principle the problem should have been there from the beginning.

riga:
This PR fixes in issue of seemingly empty sandboxes due to interplays between nested venvs as reported in #278.

The reason for this is exactly as mentioned in the linked issue, as is the solution provided by this PR.

In short, there are cases where venvs are dynamically installed while being located in an existing venv, potentially introducing non-reproducible interplays between them. Ignoring existing dependencies via `--ignore-installed / -I` is doing the trick, however, leading to potentially longer installation times as a side-effect. For this reason, this PR contains optimizations that make the setup process faster than before:

- All venv related `pip install` commands are now run with `-I`.
- When setting up a venv, one can choose between either the `--copies` (copy all files from outside the venv into the venv, even multiple times if necessary, e.g. `python`, `python3`, `python3.9` will all be copies of the outer python and of each other) or the `--symlinks` mode (allow symlinks to files pointing outside the venv *and* between files within the venv, e.g. `python -> python3`, etc). **Neither mode** is optimal for us w.r.t. relocatability and  bundle size. Therefore, the venv setup now uses `--symlinks` mode, followed by a manual replacement of symlinks with their referred file in case it's located outside the venv.
- The `pip install` commands, previously performed sequentially, are now handled in a single line, e.g. `pip install -U -I pip setuptools -r .../sandboxes/cf.txt -r ...`. This way, a consistent dependency tree is constructed once and no package is installed twice (or more).
- The `requirements_{prod,dev}.txt` files are moved into `sandboxes/{cf,dev}.txt` to be fully consistent with other sandboxes. There was no need to have requirement files at root level in the first place.
- Renamed relocateable -> relocatable (typo).

Fixes #278.


setup.py
```diff
@@ -59,7 +59,7 @@
     long_description=long_description,
     long_description_content_type="text/markdown",
     install_requires=install_requires,
-    python_requires=">=3.7, <4",
+    python_requires=">=3.7, <=3.10",
```
pkausw, line 62: Shouldn't this be

```suggestion
    python_requires=">=3.7, <=3.9",
```

?

pkausw (changes requested):
Nice PR, one miniscule comment :+1:

setup.py
riga: True.

Issue #155, opened 2023-02-06
Further distinguish cms- and experiment agnostic code.


riga:
#145 added a first simple separation CMS-related and generic analysis code in terms of tasks and producers/calibrators.

After some deeper dive, there are still some parts that exhibit some CMS dependence:

- sandbox bundle definition in config,
- sandbox bundling prior to job submission (as part of HTCondorWorkflow requirements), and
- some variables queried during the initial setup.

These remaining items should be resolved as well, by e.g. moving them into `cms` scopes or by making underlying mechanism more generic and treating the current behavior as a specialization.

PR #298, opened 2023-08-31
Support for crab job submission.

### LINKED ISSUE #296: Support for crab jobs
With law now being able to submit jobs to crab ([example](https://github.com/riga/law_example_CMSCrabWorkflows)), we should bring this feature to columnflow to be able to get more jobs in parallel.

This requires a new type of workflow, `CrabWorkflow`, that should be available to all current remote workflows. However, some restructuring might be needed to reflect that this is a CMS-only option.

riga:
This PR adds support for crab jobs and is therefore rather extensive.

The necessary steps were:

- Adding `CrabWorkflow`, inheriting from `law.cms.CrabWorkflow`
- Adding a new setup function to the `remote_boostrap.sh` script.
- Added new variables that are queried during the setup (crab related storage element and base directory)
  - Here I also updated the way that "common variables" are queried. So far, every project had to ask for the same variables with **hopefully** the same defaults, and in case there was an update in cf, all projects had to adjust their setup. There is a common function `cf_setup_interactive_common_variables` now which is defined once in cf and can be called downstream.
  - There is a variable now called `CF_FLAVOR` which is empty by default. However, when set to `cms`, CMS-specific features (such as the crab submission) are enabled. We should use the same mechanism to further separate experiment (mostly CMS) specifics from the CF core (see #155.)
- The way that cmssw sandboxes are setup in remote jobs had to be slightly tweaked to find the correct `scramv1` executable in crab jobs. 
- There is a common `RemoteWorkflowMixin` now that handles the lookup of software dependencies of remote jobs.
- `myproxy` was added to the conda stack.

Who'd be in for the review @mafrahm @dsavoiu @pkausw ?

Closes #296.

### DIFF
--- a/columnflow/tasks/cms/base.py
+++ b/columnflow/tasks/cms/base.py
@@ -1,0 +1,110 @@
+# coding: utf-8
+
+"""
+CMS related base tasks.
+"""
+
+import os
+
+import law
+
+from columnflow.tasks.framework.base import Requirements, AnalysisTask
+from columnflow.tasks.framework.remote import (
+    RemoteWorkflowMixin, BundleRepo, BundleSoftware, BundleBashSandbox, BundleCMSSWSandbox,
+)
+
+
+class CrabWorkflow(AnalysisTask, law.cms.CrabWorkflow, RemoteWorkflowMixin):
+
+    # example for a parameter whose value is propagated to the crab job configuration
+    crab_memory = law.BytesParameter(
+        default=law.NO_FLOAT,
+        unit="MB",
+        significant=False,
+        description="requested memory in MB; empty value leads to crab's default setting; "
+        "empty default",
+    )
+
+    exclude_params_branch = {"crab_memory"}
+
+    # mapping of environment variables to render variables that are forwarded
+    crab_forward_env_variables = {
+        "CF_CERN_USER": "cf_cern_user",
+        "CF_STORE_NAME": "cf_store_name",
+    }
+
+    # upstream requirements
+    reqs = Requirements(
+        BundleRepo=BundleRepo,
+        BundleSoftware=BundleSoftware,
+        BundleBashSandbox=BundleBashSandbox,
+        BundleCMSSWSandbox=BundleCMSSWSandbox,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # cached BundleRepo requirement to avoid race conditions during checksum calculation
+        self.bundle_repo_req = self.reqs.BundleRepo.req(self)
+
+    def crab_stageout_location(self):
+        # the storage site and base directory on it for crab specific outputs
+        return (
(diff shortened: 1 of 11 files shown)


columnflow/tasks/framework/remote.py
```diff
@@ -268,11 +269,135 @@ def run(self):
...
+
+        :param reqs: Dictionary of workflow requirements to be extended.
+        """
+        # add the repository bundle and trigger the checksum calculation
+        if getattr(self, "bundle_repo_req", None) is not None:
+            reqs["repo"] = self.bundle_repo_req
+        elif "BundleRepo" in self.reqs:
+            reqs["repo"] = self.reqs.BundleRepo.req(self)
+        if "rep" in reqs:
+            self.bundle_repo_req.checksum
```
mafrahm, line 295: should this be "repo" instead of "rep"?

columnflow/tasks/framework/remote.py
```diff
@@ -513,12 +567,17 @@ def htcondor_use_local_scheduler(self):
...
+        info = self.common_destination_info(info)
+        return info
+
 
 _default_slurm_flavor = law.config.get_expanded("analysis", "slurm_flavor", "maxwell")
 _default_slurm_partition = law.config.get_expanded("analysis", "slurm_partition", "cms-uhh")
 
 
-class SlurmWorkflow(AnalysisTask, law.slurm.SlurmWorkflow):
+class SlurmWorkflow(AnalysisTask, law.slurm.SlurmWorkflow, RemoteWorkflowMixin):
```
mafrahm, line 580: Does it make sense to also use the `RemoteWorkflowMixin.add_bundle_requirements` function in the `slurm_workflow_requires`?

columnflow/tasks/framework/remote.py
riga: Fixes :+1:

columnflow/tasks/framework/remote.py
riga: Slurm is designed in a way that jobs are actually executed inside a directory local to the submission node, with all software and env variables available on the remote side. Thus, there was no mechanism in place yet for the bundling of repos / sandboxes, and I left this part to a future PR in case this becomes necessary for Slurm jobs.

columnflow/tasks/cms/base.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def crab_workflow_requires(self):
+        reqs = law.cms.CrabWorkflow.crab_workflow_requires(self)
+
+        # add requirements dealing with software bundling
+        self.add_bundle_requirements(reqs)
+
+        return reqs
+
+    def crab_job_config(self, config, submit_jobs):
```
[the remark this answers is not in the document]
riga: Yep, I added the missing type hints :+1:

riga:
@mafrahm I think if there is nothing else on top, we would be ready to merge :)

mafrahm (approved):

PR #300, opened 2023-09-12
disable call cache for Categorizers

mafrahm:
I came across the issue that my categorization fails as soon as some Categorizer is called multiple times. This is due to the implementation of the call caching (see [here](https://github.com/columnflow/columnflow/blob/d59cc33de9db650a0411fac2d1ef954e73299ccf/columnflow/columnar_util.py#L2178)), which works such that it only returns the first argument (the `events` ak.Array) when a TaskArrayFunction has been called before. This, however, does not work with the Categorizer, since two return values are expected and with the current implementation, the second output (the mask) is not cached.

This PR proposes the simple temporary solution of disabling the call caching for Categorizers until we have implemented some more advanced call caching.

### DIFF
--- a/columnflow/categorization/__init__.py
+++ b/columnflow/categorization/__init__.py
@@ -52,6 +52,9 @@ class Categorizer(TaskArrayFunction):
             # get the categorizer name
             cls_name = cls_dict.pop("cls_name", func.__name__)
 
+            # disable call caching since the current implementation does not work with multiple returns
+            cls_dict["call_force"] = True
+
             # create the subclass
             subclass = cls.derive(cls_name, bases=bases, cls_dict=cls_dict, module=module)
 


pkausw:
Hi @mafrahm , thanks! Quick and probably stupid question, can't you just increase the number of `n_return` [here](https://github.com/columnflow/columnflow/blob/d59cc33de9db650a0411fac2d1ef954e73299ccf/columnflow/columnar_util.py#L2178)?

mafrahm:
Hi @pkausw, no, the `n_return` parameter does not solve our problem since we only try to return multiple input arguments, while the mask we're interested in is only given as an output value.

dsavoiu (approved):
Thanks for the hotfix. I'm OK with merging this, although there will be a performance penalty from disabling caching altogether. I expect it will be small in most cases (provided the number of nested categories is not that large), but I guess we will have to live with that before a more permanent fix is in place.

Just to note, there is a PR underway (#263) where such a fix is being discussed, but it might take a while before we converge there. So let's merge this for now and make a note to revert this changeset once #263 is merged.

PR #302, opened 2023-09-26
Configure filling of empty bins in inference model.

riga:
So far, the datacard writer had an internal setting that controlled whether empty shape bins would be filled with a small value (usually 1e-5).

This PR allows the InferenceModel to configure this behavior on a per-category level.

### DIFF
--- a/columnflow/inference/__init__.py
+++ b/columnflow/inference/__init__.py
@@ -251,7 +251,7 @@ class InferenceModel(Derivable):
         config_data_datasets: Sequence[str] | None = None,
         data_from_processes: Sequence[str] | None = None,
         mc_stats: float | tuple | None = None,
-        fill_empty_bins: float = 1e-5,
+        empty_bin_value: float = 1e-5,
     ) -> DotDict:
         """
         Returns a dictionary representing a category (interchangeably called bin or channel in other
@@ -265,7 +265,7 @@ class InferenceModel(Derivable):
               when *config_data_datasets* is not defined, make of a fake data contribution.
             - *mc_stats*: Either *None* to disable MC stat uncertainties, or a float or tuple of
               floats to control the options of MC stat options.
-            - *fill_empty_bins*: When bins are empty, they are filled with this value.
+            - *empty_bin_value*: When bins are no content, they are filled with this value.
         """
         return DotDict([
             ("name", str(name)),
@@ -274,7 +274,7 @@ class InferenceModel(Derivable):
             ("config_data_datasets", list(map(str, config_data_datasets or []))),
             ("data_from_processes", list(map(str, data_from_processes or []))),
             ("mc_stats", mc_stats),
-            ("fill_empty_bins", fill_empty_bins),
+            ("empty_bin_value", empty_bin_value),
             ("processes", []),
         ])
 
--- a/columnflow/inference/__init__.py
+++ b/columnflow/inference/__init__.py
@@ -251,6 +251,7 @@ class InferenceModel(Derivable):
         config_data_datasets: Sequence[str] | None = None,
         data_from_processes: Sequence[str] | None = None,
         mc_stats: float | tuple | None = None,
+        fill_empty_bins: float = 1e-5,
     ) -> DotDict:
         """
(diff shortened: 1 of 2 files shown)


mafrahm (approved):
The implementation looks good to me, thank you for taking care of that!

I feel like it might be a good idea to keep 0 as the default `fill_empty_bins` value to not introduce unexpected behaviour. But I also don't really have a feeling how this affects the final results. What do you think?

riga:
At least from the CMS side there is usually the recommendation to fill empty bins with epsilon's (usually 1e-5), so I think it would be good to keep it as a default.

columnflow/inference/__init__.py
```diff
@@ -264,6 +265,7 @@ def category_spec(
               when *config_data_datasets* is not defined, make of a fake data contribution.
             - *mc_stats*: Either *None* to disable MC stat uncertainties, or a float or tuple of
               floats to control the options of MC stat options.
+            - *fill_empty_bins*: When bins are empty, they are filled with this value.
```
pkausw, line 268: It would be nice to change the doc string so sphinx can automatically pick it up correctly. This shouldn't be too much work either I think, would you mind changing it?

columnflow/inference/cms/datacard.py
```diff
@@ -315,7 +315,6 @@ def write(
     def write_shapes(
         self,
         shapes_path: str,
-        fill_empty_bins: float = 1e-5,
```
pkausw, line 318: Same comment as before, would be nice if we could update the doc string already now

columnflow/inference/__init__.py
```diff
@@ -251,6 +251,7 @@ def category_spec(
         config_data_datasets: Sequence[str] | None = None,
         data_from_processes: Sequence[str] | None = None,
         mc_stats: float | tuple | None = None,
+        fill_empty_bins: float = 1e-5,
```
pkausw, line 254: `fill_empty_bins` sounds like something active, i.e. a function. Maybe `fill_emtpy_bins_with` would be more descriptive

columnflow/inference/cms/datacard.py
```diff
@@ -352,15 +351,19 @@ def write_shapes(
...
+            cat_obj = self.inference_model_inst.get_category(cat_name)
+
+            # helper to fill empty bins in-place
+            fill_empty = lambda h: None
+            if cat_obj.fill_empty_bins:
+                def fill_empty(h):
+                    value = h.view().value
+                    mask = value <= 0
+                    value[mask] = cat_obj.fill_empty_bins
+                    h.view().variance[mask] = cat_obj.fill_empty_bins
```
pkausw, line 365: See other comment about `fill_empty_bins`, might be better to call it `fill_empty_bins_with` or `default_fill_value` or something similar

pkausw (changes requested):

pkausw:
Thanks for the nice PR @riga ! Just some small comments from my side.

Also, I agree that 1e-5 is a good default value. The current statistical inference tools in CMS (i.e. combine) can't handle bins with 0 values well - if the sum of the process(es) in a bin is 0, it will lead to instabilities in the calculation of the test statistics. So 1e-5 is a safer option :+1:

columnflow/inference/__init__.py
riga: Fully agree, but it's actually quite a lot of work (the two files have 1900 lines of code with 40% being docs) that should perhaps be done in a separate PR to have *this*PR only adding the new logic.

I'd open the second PR in a bit.

columnflow/inference/cms/datacard.py
riga: And same comment as above :) Having this in a dedicated PR could be cleaner.

columnflow/inference/__init__.py
riga: True, I will rename that :+1:

columnflow/inference/cms/datacard.py
riga: Yup, renamed :+1:

pkausw (approved):
LGTM

PR #305, opened 2023-10-09
Centralize types.

riga:
This is a small PR that adds `columnflow.types` as a place where types are imported centrally and from which our code base should, in turn, get those types.

There are usually 4 or 5 packages where most types are located, such as `types`, `typing`, `type_extensions`, or `collections.abc`, and it might also be python version dependent where a specific type is eventually coming from (e.g. a type that has been previously only available in `typing_extensions`, a third-party package, might be available in the built-in `typing` in a newer python version). With this new central module, we can make these version-dependent imports in a single place instead of cluttering a bunch of files throughout the repo.

The numerous other code changes in this PR are only related to changed imports.

### DIFF
--- a/columnflow/calibration/__init__.py
+++ b/columnflow/calibration/__init__.py
@@ -7,8 +7,8 @@ Object and event calibration tools.
 from __future__ import annotations
 
 import inspect
-from typing import Callable, Sequence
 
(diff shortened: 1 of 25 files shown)


mafrahm (approved):
LGTM

PR #310, opened 2023-10-12
Allow writing root files in UniteColumns.

### LINKED ISSUE #306: Add "--file-format {parquet,root}" option to UniteColumns
The `UniteColumns` task should have the option to select a different file format for the created output file. Right now, we always create parquet files, but we should look into uproot to see whether we can easily convert to root.

@gsaha009

riga:
As documented in #306, the `UniteColumns` task should have the option to create root files to be able to share "snapshots" of events and columns with groups that have a root-based workflow.

This PR adds an option `--file-type {parquet,root}` to the `UniteColumns` task which allows switching between the formats. `parquet` still remains the default. The core functionality for exporting arrays to root trees is offloaded to two new columnar utils:

- `fill_ak_optionals`: Fills optional values (`None`'s) in an array and effectively removes the `ak.OptionType` flag from the inner primitive type (e.g. `?float32` → `float32`). This is required optional types cannot be converted to branches in root trees.
- `sorted_ak_to_root`: Takes an awkward array, removes option types using `fill_ak_optionals ` and converts the array into flat branches (using `flatten_ak_array`) and saves them in a tree using `uproot`.

@gsaha009

Closes #306.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -10,11 +10,12 @@ __all__ = [
     "mandatory_coffea_columns", "EMPTY_INT", "EMPTY_FLOAT",
     "Route", "RouteFilter", "ArrayFunction", "TaskArrayFunction", "ChunkedIOHandler",
     "eval_item", "get_ak_routes", "has_ak_column", "set_ak_column", "remove_ak_column",
-    "add_ak_alias", "add_ak_aliases", "update_ak_array", "flatten_ak_array", "sort_ak_fields",
-    "sorted_ak_to_parquet", "attach_behavior", "layout_ak_array", "flat_np_view",
-    "deferred_column", "optional_column",
+    "add_ak_alias", "add_ak_aliases", "update_ak_array", "fill_ak_optionals", "flatten_ak_array",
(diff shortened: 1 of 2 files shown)


columnflow/columnar_util.py
```diff
@@ -925,6 +926,45 @@ def _create_matcher(routes):
...
+    null_value: int | float = EMPTY_INT,
+) -> ak.Array:
+    """
+    Fills missing (None) values in a non-nested *ak_array* with *null_value*, preserving the
+    original primitive type but removing the option type. Please note that even if *ak_array* does
+    not contain any None values, but its type contains a option type flag, *null_value* is actually
+    not used yet still required by :py:func:`awkward.fill_none` which is used internally.
+
+    :param ak_array: The input array.
+    :param null: The value that is used to fill potential None values.
```
pkausw, line 940:
```suggestion
    :param null_value: The value that is used to fill potential None values.
```

pkausw (changes requested):
LGTM, thanks! Just one typo in the docs (which are very beautiful btw :) )

pkausw (approved):

pkausw (approved):

PR #312, opened 2023-10-12
Use ZSTD(1) as a default compression for all parquet files.

### LINKED ISSUE #309: Optimize compression for parquet files
Recently, users tried to process very large NanoAOD sample, e.g. the `TTToHadron` sample which has a size of ~1TB. In this case, the user only had a limited amount of disk space available (also ~1TB), and this space completely ran out during the CalibrateEvents task.

Further studies have shown that the original NanoAOD .root files with the full set of information is of order ~700MB/file. The corresponding parquet files from the CalibrateEvents task are of order ~500MB/file, although they only contain a fraction of the information (~128 columns with `float32` or `uint64`). Essentially, we almost double the size of a NanoAOD sample in the CalibrateEvents task. Maybe we can further optimize this with different compression rates?

riga:
As discussed in #309, we should enforce a better default compression when writing parquet files. Instead of the default "SNAPPY" algorithm, this PR uses ZSTD(1).

This is done by having a central method `AnalysisTask.get_parquet_writer_opts()`, which is used throughout all relevant tasks now and whose return value is passed as `writer_opts` to `law.pyarrow.merge_parquet_task`.

Closes #309.

### DIFF
--- a/columnflow/tasks/calibration.py
+++ b/columnflow/tasks/calibration.py
@@ -147,7 +147,9 @@ class CalibrateEvents(
         # merge output files
         with output["columns"].localize("w") as outp:
             sorted_chunks = [output_chunks[key] for key in sorted(output_chunks)]
-            law.pyarrow.merge_parquet_task(self, sorted_chunks, outp, local=True)
+            law.pyarrow.merge_parquet_task(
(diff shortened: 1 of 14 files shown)


mafrahm (approved):
Looks good to me!

PR #313, opened 2023-10-12
Enhance cmssw sandboxes

riga:
This PR enhances the way how CMSSW sandboxes are built and bundled.

For #306 I was testing different approaches for converting arrays to root files and one of them was `ak.to_rdataframe` which required ROOT to be installed. This would have only been possible in a cmssw sandbox, but with additional software installed through a venv on top. This was not possible before as there was no custom hook to install something between `cmsenv` and `scram b` during the cmssw setup.

This is now possible with this PR. As a showcase I added a `cmssw_columnar` sandbox which makes use of this hook to install our `columnar` requirements on top of cmssw.

Besides the features above, I streamlined the sandbox file resolution in the sandbox tasks in `tasks/framework/remote.py`.

### DIFF
--- a/columnflow/tasks/framework/remote.py
+++ b/columnflow/tasks/framework/remote.py
@@ -93,11 +93,31 @@ class BundleSoftware(AnalysisTask, law.tasks.TransferLocalFile):
         self.transfer(bundle)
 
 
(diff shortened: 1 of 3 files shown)


mafrahm (approved):
LGTM

PR #314, opened 2023-10-13
allow merging of MultiSettingsParameters and implement tests

mafrahm:
This PR implements some tests for the `SettingsParameter` and `MultiSettingsParameter`.
It also fixes one issue, where dicts would be overwritten if you provide the same key twice, e.g.

```python
MultiSettingsParameter().parse("tt,A=2:st,A=2:tt,B=True") 
#  before: {"tt": {"B": True}, "st": {"A": 2.0}}
#  now: {"tt": {"A": 2.0, "B": True}, "st": {"A": 2.0}}
```

### DIFF
--- a/tests/run_tests
+++ b/tests/run_tests
@@ -26,6 +26,11 @@ action() {
     ret="$?"
     [ "${gret}" = "0" ] && gret="${ret}"
 
+    # test_task_parameters
+    bash "${this_dir}/run_test" test_task_parameters
+    ret="$?"
+    [ "${gret}" = "0" ] && gret="${ret}"
+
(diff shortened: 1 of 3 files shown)


tests/run_tests
```diff
@@ -26,6 +26,11 @@ action() {
     ret="$?"
     [ "${gret}" = "0" ] && gret="${ret}"
 
+    # test_task_parameters
```
riga, line 29:
```suggestion
    # test_task_parameters
    echo
```

riga (approved):

riga:
That looks good!

PR #332, opened 2023-10-25
fix build docu for master branch

nprouvost:
Correcting the path to the standard columnflow package requirement file in the docs/requirements.txt file, such that the readthedocs build does not fail anymore.


riga (approved):
Great, thank you!

PR #333, opened 2023-10-26
Update columnar software in venvs.

riga:
This PR updates the python packages in the `venv_columnar*` sandboxes.

Most notably, switching to pyarrow 13 is successful now, most likely due to updates in awkward itself.


pkausw (changes requested):
Thanks for the update! Very nice, but I think we also need to increase the version of the docs requirements, since these also depend on the `columnar` requirements

pkausw (approved):

pkausw:
LGTM, thanks @riga ! I made one additional minor change in `setup.py` -- the python version requirement was not compatible with the currently used python version `3.9.1`, so I increased this. Would be good if you could maybe also have a look though :D

PR #339, opened 2023-10-30
Use dict_encoding for selection results.

### LINKED ISSUE #338: Use dict encoding for mask columns
In #329, we disabled dict encoding for all our parquet outputs. However, it might be nice to enable the dict encoding for columns that only include masks since they might profit quite a lot from the dict encoding.

riga:
This PR adds a switch to `get_parquet_writer_opts` to follow different encoding strategies in case values to be written are predominantly masks. So far, only dict encoding is used, but we might want to investigate this further in the future (e.g. different compression methods and levels).

This behavior is enabled now in `SelectEvents` for writing selection masks, reducing file sizes by approximately 20%.

Closes #338.


mafrahm (approved):
Looks good to me!

PR #342, opened 2023-11-06
Ensure Python 3.7 compatibility.

riga:
This PR re-ensures the compatibility with Python 3.7 (as recently discussed). In essence, the only non-compatible syntax was the use of walrus operators `:=` which is changed back by this PR.

### DIFF
--- a/columnflow/production/cms/pdf.py
+++ b/columnflow/production/cms/pdf.py
@@ -61,14 +61,19 @@ def pdf_weights(
 
     Resources:
 
-       - https://arxiv.org/pdf/1510.03865.pdf
+        - https://arxiv.org/pdf/1510.03865.pdf
     """
 
-    if outlier_action not in (known_actions := ("ignore", "remove", "raise")):
+    known_actions = ("ignore", "remove", "raise")
+    if outlier_action not in known_actions:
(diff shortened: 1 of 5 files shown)


columnflow/production/cms/pdf.py
```diff
@@ -61,14 +61,19 @@ def pdf_weights(
...
-    if outlier_action not in (known_actions := ("ignore", "remove", "raise")):
+    known_actions = ("ignore", "remove", "raise")
+    if outlier_action not in known_actions:
         raise ValueError(
             f"unknown outlier_action '{outlier_action}', known values are {','.join(known_actions)}",
         )
-    if outlier_log_mode not in (known_log_modes := ("none", "info", "debug", "warning")):
+        raise ValueError(
+            f"unknown outlier_action '{outlier_action}', known values are {','.join(known_actions)}",
+        )
```
mafrahm, line 74: Here is some code duplication

mafrahm (approved):
Looks good, there is just one minor inline comment

columnflow/production/cms/pdf.py
riga: Fixed :+1:

PR #343, opened 2023-11-06
Set proper LAW_CONFIG_FILE in remote jobs.

riga:
This PR updates the behavior of remote jobs in their handling of the LAW_CONFIG_FILE.

So far, remote jobs disregarded the LAW_CONFIG_FILE set in the local submission environment. This can obviously lead to different settings being used in remote jobs than expected. This is fixed now.

In addition, I streamlined the handling of files to be sent (or set) in remote jobs (law config, kerberos token, voms proxy token, wlcg tools, ...).


mafrahm (approved):
LGTM

PR #345, opened 2023-11-08
Refactor chunked parquet reading

riga:
This PR adds an additional thread-safe method for the loading chunks of parquet files to our custom `DaskArrayReader` that is used by the more general `ChunkedIOHandler`.

### Current implementation

So far, we determined the partitioning of parquet files, which was always matching our default chunk size since all parquet files we handle are usually produced within columnflow itself with that chunk size. As a result, it was always possible to load a full partition and pass it to the event chunk loop without any adjustments or sharing of chunk overlaps between threads (although this was implemented).

### Issue

However, it appears this has changed in a previous ak/dak release (*or* maybe this is due to our recent update to the default parquet compression and encoding). As a result, there is only a single partition of data in parquet files, and hence the `DaskArrayReader` is forced to load the complete file into memory at once, and then artificially divide it into chunks. This forces the reader the bypass the very idea of multi-threaded IO.

### Solution

The good news is that dask awkward arrays (dak) now support slicing (they were previously raising NotImplementedError's). This PR preserves the old partition-style materialization strategy (i.e. when to call `compute()` on what), but adds a new one that relies on slicing and that is the default now. With that we leave the IO optimization to the upstream dak package, which very likely has more means to do so in the first place.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -2369,7 +2369,18 @@ class DaskArrayReader(object):
     to be read more than once, another instance of this class should be used.
     """
 
-    def __init__(self, path: str, open_options: dict | None = None):
+    class MaterializationStrategy(enum.Flag):
+        """
+        Flag to define which materialization strategy to follow.
+        """
+        SLICES = enum.auto()
+        PARTITIONS = enum.auto()
+
+    def __init__(
+        self: DaskArrayReader,
+        path: str, open_options: dict | None = None,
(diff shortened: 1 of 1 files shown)


mafrahm (approved):
Looks good to me! Did you test the improved performance?

riga:
Yes, but so far only on a relatively small file in `ProduceColumns` but it was about 10% faster. Our `SelectEvents` step is fully CPU limited so a performance check there was not sensible.

PR #346, opened 2023-11-08
running Producer in cf.PrepareMLEvents

mafrahm:
This PR allows setting a Producer in a MLModel to run during the `cf.PrepareMLEvents` task.

### DIFF
--- a/columnflow/ml/__init__.py
+++ b/columnflow/ml/__init__.py
@@ -116,9 +116,12 @@ class MLModel(Derivable):
     # falls back to cls_name if None
     store_name: str | None = None
 
+    # default setting wether to run the preparation_producer in MLEvaluation
+    preparation_producer_in_ml_evaluation: bool = True
+
     # names of attributes that are automatically extracted from init kwargs and
     # fall back to classmembers in case they are missing
-    init_attributes: list[str] = ["single_config", "folds", "store_name"]
+    init_attributes: list[str] = ["single_config", "folds", "store_name", "preparation_producer_in_ml_evaluation"]
 
     def __init__(
         self: MLModel,
--- a/columnflow/ml/__init__.py
+++ b/columnflow/ml/__init__.py
@@ -42,7 +42,7 @@ class MLModel(Derivable):
     (:py:meth:`requires`), diverging training and evaluation phase spaces
     (:py:meth:`training_configs`, :py:meth:`training_calibrators`, :py:meth:`training_selector`,
     :py:meth:`training_producers`), or how hyper-paramaters are string encoded for output
-    declarations (:py:meth:`parameter_pairs`). The optional `py:meth:`preparation_producer`` allows
+    declarations (:py:meth:`parameter_pairs`). The optional py:meth:`preparation_producer` allows
     setting a producer that is run during the initial preparation of ML columns.
 
     .. py:classattribute:: single_config
@@ -116,7 +116,7 @@ class MLModel(Derivable):
     # falls back to cls_name if None
     store_name: str | None = None
 
-    # default setting wether to run the preparation_producer in MLEvaluation
+    # flag denoting whether the preparation_producer is invoked before evaluate()
     preparation_producer_in_ml_evaluation: bool = True
 
     # names of attributes that are automatically extracted from init kwargs and
@@ -350,12 +350,15 @@ class MLModel(Derivable):
         return list(requested_producers)
 
     def preparation_producer(
-        self,
+        self: MLModel,
         config_inst: od.Config,
     ) -> str | None:
         """
-        This method allows setting a producer that can be called as part of the preparation of the
-        ML input columns.
+        This method allows setting a producer that can be called as part of the preparation
+        of the ML input columns given a *config_inst*.
+
+        :param config_inst: :py:class:`~order.Config` object for which the producer should run.
+        :return: Name of a :py:class:`Producer` class or *None*.
         """
         return None
 
--- a/columnflow/ml/__init__.py
+++ b/columnflow/ml/__init__.py
@@ -7,7 +7,7 @@ Definition of basic objects for describing and creating ML models.
 from __future__ import annotations
 
 import abc
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 
 import law
 import order as od
@@ -42,7 +42,8 @@ class MLModel(Derivable):
     (:py:meth:`requires`), diverging training and evaluation phase spaces
     (:py:meth:`training_configs`, :py:meth:`training_calibrators`, :py:meth:`training_selector`,
     :py:meth:`training_producers`), or how hyper-paramaters are string encoded for output
-    declarations (:py:meth:`parameter_pairs`).
+    declarations (:py:meth:`parameter_pairs`). The optional `py:meth:`increment_stats`` hook
+    allows gathering stats from the prepared input columns.
 
     .. py:classattribute:: single_config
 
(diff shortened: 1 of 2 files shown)


riga:
I resolved conflicts resulting from updated type hints.

columnflow/ml/__init__.py
```diff
@@ -115,9 +116,12 @@ class MLModel(Derivable):
     # falls back to cls_name if None
     store_name: str | None = None
 
+    # default setting wether to run the preparation_producer in MLEvaluation
```
riga, line 119: Small typo, `whether`.
Also, maybe this sentence could be

```
# flag denoting whether the preparation_producer is invoked before evaluate()
```

to make it more clear when the producer is called exactly.

columnflow/ml/__init__.py
```diff
@@ -345,6 +349,16 @@ def training_producers(
         """
         return list(requested_producers)
 
+    def preparation_producer(
+        self,
```
riga, line 353: `self: MLModel,`

columnflow/ml/__init__.py
```diff
@@ -345,6 +349,16 @@ def training_producers(
...
+        config_inst: od.Config,
+    ) -> str | None:
+        """
+        This method allows setting a producer that can be called as part of the preparation of the
+        ML input columns.
```
riga, line 358:
```
This method allows setting a producer that can be called as part of the preparation of the ML input columns given a *config_inst*.

:param config_inst: :py:class:`~order.Config` object for which the producer should run.
:return: Name of a :py:class:`Producer` class or *None*.
```

columnflow/tasks/ml.py
```diff
@@ -64,12 +70,29 @@ def __init__(self, *args, **kwargs):
                 f"{self.__class__.__name__}",
             )
 
+    @property
+    def preparation_producer_inst(self):
+        producer = self.ml_model_inst.preparation_producer(self.config_inst)
+        if producer and self._preparation_producer_inst is None:
```
riga, line 76: In case the `preparation_producer` method is doing something complicated, I would evaluate the method (first line) only when the `_preparation_producer_inst` is not yet cached.

Also, with the comment above, check for `law.no_value` instead of *None*.

columnflow/tasks/ml.py
```diff
@@ -51,6 +54,9 @@ class PrepareMLEvents(
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        # cache for producer inst
+        self._preparation_producer_inst = None
```
riga, line 58: Maybe use `law.no_value` here instead of *None* to mark this attribute as "not cached so far", since *None* is a viable value in case no preparation producer is configured.

columnflow/tasks/ml.py
```diff
@@ -537,58 +734,78 @@ def run(self):
...
+
+                # add aliases
+                events = add_ak_aliases(
+                    events,
+                    aliases,
+                    remove_src=True,
+                    missing_strategy=self.missing_column_alias_strategy,
+                )
+
+                # asdasd
```
riga, line 777: :)

columnflow/tasks/ml.py
```diff
@@ -537,58 +734,78 @@ def run(self):
         read_columns = {Route("deterministic_seed")}
         read_columns |= set(map(Route, aliases.values()))
         read_columns |= set.union(*self.ml_model_inst.used_columns.values())
+        if self.preparation_producer_inst:
+            read_columns |= self.preparation_producer_inst.used_columns
+        read_columns = {Route(c) for c in read_columns}
```
riga, line 739: All objects in read_columns should be Route's already, so no need to re-cast.

columnflow/tasks/ml.py
```diff
@@ -445,9 +603,32 @@ class MLEvaluation(
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        # cache for producer inst
+        self._preparation_producer_inst = None
```
riga, line 607: Same as above, maybe use `law.no_value`.

columnflow/tasks/ml.py
```diff
@@ -445,9 +603,32 @@ class MLEvaluation(
...
+
         # set the sandbox
         self.sandbox = self.ml_model_inst.sandbox(self)
 
+    @property
+    def preparation_producer_inst(self):
+        producer = None
+        if self.ml_model_inst.preparation_producer_in_ml_evaluation:
+            producer = self.ml_model_inst.preparation_producer(self.config_inst)
+        if producer and self._preparation_producer_inst is None:
```
riga, line 617: Same as above, restructure such that nothing is executed twice when already cached.

riga (changes requested):
Nice, this looks really good and will surely make the ML pipeline more flexible.

I think we can proceed with this rather fast, I just had minor comments (inline).

columnflow/tasks/ml.py
```diff
@@ -64,12 +70,33 @@ def __init__(self, *args, **kwargs):
...
+
+        producer = self.ml_model_inst.preparation_producer(self.config_inst)
+        if producer:
+            self._preparation_producer_inst = ProducerMixin.get_producer_inst(producer, {"task": self})
+
+            # overwrite the sandbox when set
+            if self._preparation_producer_inst.sandbox:
+                self.sandbox = self._preparation_producer_inst.sandbox
+
+        return self._preparation_producer_inst
```
riga, line 87: Mh, this potentially returns `law.no_value`. I think in case `preparation_producer` returns None, the `self._preparation_producer_inst` should also return None, which is does not right now.

columnflow/tasks/ml.py
```diff
@@ -96,10 +125,14 @@ def requires(self):
...
-        return {"mlevents": law.SiblingFileCollection([
-            self.target(f"mlevents_fold{f}of{k}_{self.branch}.parquet")
-            for f in range(k)
-        ])}
+        outputs = {
+            "mlevents": law.SiblingFileCollection([
+                self.target(f"mlevents_fold{f}of{k}_{self.branch}.parquet")
+                for f in range(k)
+            ]),
+            "stats": self.target(f"stats_{self.branch}.parquet"),
```
riga, line 133: Can there be a "stats" output if there is no preparation producer? If not, there should be no "stats" entry at all I assume.

columnflow/tasks/ml.py
```diff
@@ -445,9 +607,38 @@ class MLEvaluation(
...
+            if (
+                self._preparation_producer_inst.sandbox and
+                self.sandbox != self._preparation_producer_inst.sandbox
+            ):
+                raise Exception(
+                    f"Task {self.__class__.__name__} got different sandboxes from the MLModel ({self.sandbox}) "
+                    f"than from the preparation_producer ({self._preparation_producer_inst.sandbox})",
+                )
+
+        return self._preparation_producer_inst
```
riga, line 640: Same comment as above.

riga (changes requested):

columnflow/tasks/ml.py
mafrahm: Currently, there is one default field that is produced independently of the preparation_producer: https://github.com/columnflow/columnflow/blob/5c5dbf2edf76787968384a0fd65559d448a0103f/columnflow/tasks/ml.py#L251-L253
In that way, we always have some stats file that can be merged. If we change that, we also have to make sure that MergeMLStats will not be requested by MLTraining

riga: I think this is actually a field one always wants / needs to have, so I'm good with that :)

riga (approved):

PR #348, opened 2023-11-09
Introducing ColumnCollection's to define columns to keep.

riga:
This PR introduces `ColumnCollection`'s which are flags to denote certain types of - as the names suggests - collections of columns.

Background: we got a couple questions recently on why certain columns were not existing in processed arrays at e.g. `ProduceColumns` or `CreateHistograms` level. The reason always was that the columns to keep were not properly defined in the `cfg.x.keep_columns` map. However, we already discussed that this is not very transparent anyway.

I think it's correct that these kind of "columns to keep" need to be defined *somewhere*, in most use cases it would be sufficient to have a placeholder in the `keep_columns` map that means something like "just keep all columns produced by the calibrators / selector / producers / ...". `ColumnCollection`s are exactly that.

So far, there are just a handful of flags (e.g. `ColumnCollection.ALL_FROM_SELECTOR`) but this could even be extended in the future.

For the particular use case of determining the columns to keep at various places where some columns are dropped (ReduceEvents, UniteColumns, MergeSelectionMasks), I added `find_keep_columns` to our mixin classes which makes a lot of sense inheritance-wise (e.g. all tasks having the `SelectorMixin` should be able to access columns produced by a selector and thus, should be able to resolve the `ColumnCollection.ALL_FROM_SELECTOR` flag).

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -7,7 +7,7 @@ Helpers and utilities for working with columnar libraries.
 from __future__ import annotations
 
 __all__ = [
-    "mandatory_coffea_columns", "EMPTY_INT", "EMPTY_FLOAT",
+    "mandatory_coffea_columns", "ColumnCollection", "EMPTY_INT", "EMPTY_FLOAT",
     "Route", "RouteFilter", "ArrayFunction", "TaskArrayFunction", "ChunkedIOHandler",
     "eval_item", "get_ak_routes", "has_ak_column", "set_ak_column", "remove_ak_column",
     "add_ak_alias", "add_ak_aliases", "update_ak_array", "flatten_ak_array", "sort_ak_fields",
@@ -529,6 +529,18 @@ class Route(od.TagMixin):
                         res = ak.fill_none(res, null_value)
 
         return res
+
+
+class ColumnCollection(enum.Flag):
+    """
+    Enumeration containing flags that describe arbitrary collection of columns.
+    """
+
+    ALL_FROM_CALIBRATOR = enum.auto()
+    ALL_FROM_CALIBRATORS = enum.auto()
+    ALL_FROM_SELECTOR = enum.auto()
+    ALL_FROM_PRODUCER = enum.auto()
+    ALL_FROM_PRODUCERS = enum.auto()
 
 
 def get_ak_routes(
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -18,6 +18,7 @@ import law
 import order as od
 
 from columnflow.types import Sequence, Callable, Any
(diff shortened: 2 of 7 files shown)


columnflow/tasks/framework/base.py
```diff
@@ -710,6 +711,19 @@ def store_parts(self):
...
+    def find_keep_columns(self: ConfigTask, collection: ColumnCollection) -> set[Route]:
+        """
+        Returns a set of :py:class:`Route` objects describing columns that should be kept given a
+        type of column *collection*.
+
+        :param collection: The collection to return.
+        :return: A set of :py:class:`Route` objects.
+        """
+        if collection == ColumnCollection.MANDATORY_COFFEA:
+            return set(Route(c) for c in mandatory_coffea_columns)
```
mafrahm, line 723: I think you never initialized the `MANDATORY_COFFEA` class attribute

columnflow/columnar_util.py
```diff
@@ -531,6 +531,18 @@ def apply(
...
+    """
+    Enumeration containing flags that describe arbitrary collections of columns.
+    """
+
+    ALL_FROM_CALIBRATOR = enum.auto()
+    ALL_FROM_CALIBRATORS = enum.auto()
+    ALL_FROM_SELECTOR = enum.auto()
+    ALL_FROM_PRODUCER = enum.auto()
+    ALL_FROM_PRODUCERS = enum.auto()
+
```
mafrahm, line 544: you could also add one `ALL_FROM_ML_EVALUATION` ColumnCollection

mafrahm (commented):
Looks good to me, just two minor inline comments

columnflow/tasks/framework/base.py
riga: True. Added :+1:

columnflow/columnar_util.py
riga: Good point. Added :+1:

columnflow/tasks/framework/mixins.py
```diff
@@ -502,6 +530,15 @@ def store_parts(self):
...
         return parts
 
+    def find_keep_columns(self: ConfigTask, collection: ColumnCollection) -> set[Route]:
+        if collection == ColumnCollection.ALL_FROM_PRODUCERS:
+            return set.union(*(
+                producer_inst.produced_columns
+                for producer_inst in self.producer_insts
+            ))
+
+        return super().find_keep_columns(collection)
```
mafrahm, line 540: I think we also need this method for the `MLModelMixin` and `MLModelsMixin`

riga: Added :+1:

riga:
@mafrahm I want to add another commit soon to correctly combine sets with those found in a super function.

mafrahm (approved):

PR #354, opened 2023-11-15
Add unit tests for SelectionResult object

### LINKED ISSUE #351: Improve adding of `object` fields in SelectionResult objects
While writing unit tests for the `SelectionResult` class, I came upon this feature of the current implementation. Consider these two object fields:

```python
result1 = {
  "objects": {
      "Jet": {
          "Jet": ak.Array([[0, 1], [2, 3], [0], [0]]),
      },
  },
}

result2 = {
  "objects": {
    "Muon": {
      "Muon": ak.Array([[0, 1], [2, 3], [0], [0]]),
      },
     "Jet": {
       "Jet": ak.Array([[0, 3], [2, 3], [0, 1], [0]]),
       "selectedJet": ak.Array([[0, 1], [2, 3], [0], [0]])
     } 
  },
}
```

After adding two SelectionResult instances derived from the dictionaries above (called `bar`, I see this

```python
In [54]: bar = selection_result1 + selection_result2
In [55]: bar.objects
Out[55]: 
DotDict([('Jet',
          DotDict([('Jet',
                    <Array [[0, 3], [2, 3], [0, 1], [0]] type='4 * var * int64'>),
                   ('selectedJet',
                    <Array [[0, 1], [2, 3], [0], [0]] type='4 * var * int64'>)])),
         ('Muon',
          DotDict([('Muon',
                    <Array [[0, 1], [2, 3], [0], [0]] type='4 * var * int64'>)]))])
 ```

As you can see, the `Jet/Jet` section is simply updated (and thus overwritten) in the `add` operation. Imho, this is not the expected result - if two selections both produce the same collection (in this case `Jet/Jet`), the logical thing would be to select objects that survive both selections (i.e. do an AND between the indices). 

Of course, we might also want to avoid this behavior altogether, since defining the same collection of objects in different selectors might cause confusion. On the other hand, this would limit the usage of the selectors - maybe a user **wants** to update a given collection of objects based on some other criteria.

Imho, there are two way we could go about this:

- Supress the definition of the same field in different SelectionResults, i.e. raise some kind of Error
- update the add operation to do a proper AND of the indices.

What do you think about this?

pkausw:
This PR introduces unit tests for the SelectionResult class. Additionally, it also fixes the following issues:

- defining the same set of objects in `objects` is now prohibited, as suggested in #351 
- previously, converting a SelectionResult with an additional nested top-level field did not work correctly. Now fixed
- converting an empty SelectionResult generated an error, fixed now by simply returning an empty ak.Array

There are two tests for writing and reading SelectionResults to/from parquet files with awkward and dask awkward. This might be superseded by a future unit test that checks this in a more general approach.

Closes #351


pkausw:
@riga , any updates on this PR?

pkausw:
This is an extremely minor PR, which has been open for a _very_ long time now. Is there anything that needs to be changed? Otherwise, I would vote for merging it soon @riga

PR #364, opened 2023-11-24
implement custom_style_config plotting parameter

mafrahm:
This PR implements the `custom_style_config` parameter, which is typically just a string pointing to a key of the `config.x.custom_style_config_groups` auxiliary. The usecase is to overwrite the `style_config` that is handed to all our plotting functions.

This parameter allows lots of customizability from the user side, but many of the options are better handled by other parameters.

Some examples for the config:
```python
config.x.default_custom_style_config = {
    "legend_cfg": {"title": "my custom title", "ncols": 2},
}
config.x.custom_style_config_groups = {
    "example": {
        "legend_cfg": {"title": "my custom legend title", "ncols": 2},
        "ax_cfg": {"ylabel": "my ylabel", "xlim": (0, 100)},
        "rax_cfg": {"ylabel": "some other ylabel"},
        "annotate_cfg": {"text": "category label usually here"},
    },
}
```

### DIFF
--- a/columnflow/tasks/framework/plotting.py
+++ b/columnflow/tasks/framework/plotting.py
@@ -10,7 +10,7 @@ import law
 import luigi
 
 from columnflow.types import Any, Callable
-from columnflow.tasks.framework.base import ConfigTask
+from columnflow.tasks.framework.base import ConfigTask, RESOLVE_DEFAULT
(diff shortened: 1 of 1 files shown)


pkausw (approved):
Nice PR, adds a lot more flexibilty :+1: LGTM

PR #365, opened 2023-11-27
Feature/get events from categories

### LINKED ISSUE #352: Add helper to select events from one category
It would be nice to have some helper function that automatically selects events given a config category based on an existing `category_ids`column.

```
masked_events = columnflow.columnar_util.select_category(events, config.get_category("1e"))
```

This should automatically select all events that contain a category id of one of the "1e" leaf categories.

mafrahm:
Closes #352

### DIFF
--- a/columnflow/config_util.py
+++ b/columnflow/config_util.py
@@ -41,12 +41,12 @@ def get_events_from_categories(
     Helper function that returns all events from an awkward array *events* that are categorized
     into one of the leafs of one of the *categories*.
 
-    :param events: awkward array. Requires the 'category_ids' field to be present.
-    :param categories: sequence of category instances. Can also be a sequence of strings when passing a
+    :param events: Awkward array. Requires the 'category_ids' field to be present.
+    :param categories: Sequence of category instances. Can also be a sequence of strings when passing a
         *config_inst*.
-    :param config_inst: optional config instance to load category instances.
+    :param config_inst: Optional config instance to load category instances.
     :raises ValueError: If "category_ids" is not present in the *events* fields.
-    :return: awkward array of all events that are categorized into one of the leafs of one of the
+    :return: Awkward array of all events that are categorized into one of the leafs of one of the
         *categories*
     """
     if "category_ids" not in events.fields:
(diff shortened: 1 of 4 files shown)


columnflow/config_util.py
```diff
@@ -25,8 +25,50 @@
...
+
+
+def get_events_from_categories(
+    events: ak.Array,
+    categories: Sequence[str | od.Category],
+    config_inst: od.Config | None = None,
+) -> ak.Array:
+    """
+    Helper function that returns all events from an awkward array *events* that are categorized
+    into one the leafs of one of the *categories*.
```
pkausw, line 42:
```suggestion
    into one of the leafs of one of the *categories*.
```

tests/test_config_util.py
```diff
@@ -0,0 +1,66 @@
...
+        leaf_22 = main_2.add_category("leaf_22", id=22)
+        leaf_22.add_category("leaf_221", id=221)
+
+        # awkward array with category_ids (only leaf ids)
+        self.events = ak.Array({
+            "category_ids": [[1], [21, 221], [21], [221]],
+            "dummy_field": [1, 2, 3, 4],
+        })
+
+    def test_get_events_from_categories(self):
```
pkausw, line 40: You could probably split this test into each of the different blocks (i.e. smaller unit tests), but this is optional imho

tests/test_config_util.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # check that leaf category is working
+        events = get_events_from_categories(self.events, ["main_1", "main_2"], self.config_inst)
+        self.assertTrue(ak.all(events.dummy_field == ak.Array([1, 2, 3, 4])))
+
+        # check that directly passing category inst is working
+        events = get_events_from_categories(self.events, self.config_inst.get_category("main_1"))
+        self.assertTrue(ak.all(ak.any(events.category_ids == 1, axis=1)))
+        self.assertTrue(ak.all(events.dummy_field == ak.Array([1])))
+
```
pkausw, line 63: The new helper also `raises` an Exception at some point, would also be good to test this explicitely

pkausw (changes requested):
Nice PR @mafrahm ! I think it would be nice to further test things that are not supposed to work, both to make sure that unexpected behaviors are less probable and also for more clarity of how this module is supposed to work

riga (approved):
LGTM!

PR #372, opened 2023-12-18
Documentation about the usage of Machine Learning

Bogdan-Wiederspan:
This PR covers:
- A code example with a working workflow from setup to training and evaluation
- A documentation that explains how to...
  - prepare ML events, sandbox, datasets and configs
  - init the model with your config
  - use different producers, calibrators etc. depending on training and evaluation mode
  - start the training of the model
  - use already trained models to evaluate new data
  - best practices


docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+In this case we want to preserve the input and target features, as well as some weights:
+```{literalinclude} ./ml_code.py
+:language: python
+:pyobject: MLModel.uses
+```
+
+## output:
+In the {py:meth}`~columnflow.ml.MLModel.output` function, you define your local target directory that your current model instance will have access to.
+
+Since machine learning in columnflow uses k-fold by default, it is a good idea to have a separate directory for each fold, and this should be reflected in the {py:meth}`~columnflow.ml.MLModel.output` path.
```
pkausw, line 104:
```suggestion
Since machine learning in columnflow uses k-fold cross validation by default, it is a good idea to have a separate directory for each fold, and this should be reflected in the {py:meth}`~columnflow.ml.MLModel.output` path.
```

Also: You write this as if this is a well-established fact, but I don't think so. While most people probably should use k-fold cross validation, not a lot of people do. So this should definitely be something that is stated pretty early on on this page

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+It is of good practice to use `task.target` to be that your "machine-learning-task" related files are saved in the same directory.
```
pkausw, line 105: >  ... to use `task.target` to be that your "machine-learning-task" related ...

What do you mean by this?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+:start-at: def open_model
+:end-at: return loaded_model
+```
+
+## train:
+In the {py:meth}`~columnflow.ml.MLModel.train` function, you implement the initialization of your models and the training loop.
+The `task` corresponding to the models training is {py:class}`~columnflow.tasks.ml.MLTraining`.
+By default {py:meth}`~columnflow.ml.MLModel.train` has access to the location of the models `inputs` and `outputs`.
+
+In columnflow, k-fold is enabled by default.
```
pkausw, line 131:
```suggestion
In columnflow, k-fold cross validation is enabled by default.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+The `Self` argument in {py:meth}`~columnflow.ml.MLModel.train` referes to the instance of the fold.
```
pkausw, line 132:
```suggestion
The `self` argument in {py:meth}`~columnflow.ml.MLModel.train` referes to the instance of the fold.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+Using `Self`, you have also access to the entire `analysis_inst`ance the `config_inst`ance of the current fold, and to all the derived parameters of your model.
```
pkausw, line 133:
```suggestion
Using `self`, you have also access to the entire `analysis_inst`ance the `config_inst`ance of the current fold, and to all the derived parameters of your model.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+
+With this information, you can call and prepare the columns to be used by the model for training.
+In the following exmaple a very simple dummy trainings loop is performed using Keras fit function.
```
pkausw, line 136:
```suggestion
In the following example a very simple dummy trainings loop is performed using the Keras fit function.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def train
+:end-at: return
+```
```
pkausw, line 141: Side note: you should also explain `self.build_model()`, `self.open_input_files()` and `self.prepare_events()` somewhere. API is also fine of course

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+### Good practice for training:
+It is considered to be a good practice to define helper functions for model building and column preprocessing and call them in {py:meth}`~columnflow.ml.MLModel.train`.
+The reasoning behind this design decision is that in {py:meth}`~columnflow.ml.MLModel.train`, you want to focus more on the definition of the actual trainings loop.
+Another reason is that especially the preparation of events, can take a large amount of code lines, making it messy to debug.
```
pkausw, line 145:
```suggestion
Another reason is that especially the preparation of events can take a large amount of code lines, making it messy to debug.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+In the following example it is shown how to define these helper functions.
+First of all one needs a function to handle the opening and combination of all parquet files.
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def open_input_files
+:end-at: return all_events
+```
+
+In `prepare_events` we use the combined awkward and filter out all columns we are not interested in during the training, to stay lightweight.
```
pkausw, line 154:
```suggestion
In `prepare_events` we use the combined awkward array and filter out all columns we are not interested in during the training, to stay lightweight.
```

pkausw: arguably, the most lightweight-way to load the columns would probably be in your `open_input_files` function

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def build_model
+:end-at: return model
+```
+With this little example one can see why it is of good practice to separate this process into small chunks.
+
+## evaluate:
+Within {py:meth}`~columnflow.ml.MLModel.train` one defined the trainings process of the ML model, while in {py:meth}`~columnflow.ml.MLModel.evaluate` its evaluation is defined.
+The corresponding `task` is {py:class}`~columnflow.tasks.ml.MLEvaluation` and depends on {py:class}`~columnflow.tasks.ml.MLTraining` and will therefore trigger a training if no training was performed before.
```
pkausw, line 172:
```suggestion
The corresponding `task` is {py:class}`~columnflow.tasks.ml.MLEvaluation`, which depends on {py:class}`~columnflow.tasks.ml.MLTraining` and will therefore trigger a training if no training was performed before.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+
+For each fold of the k-folds a neural network model is trained and can be accessed from `models`.
```
pkausw, line 174:
```suggestion
For each fold of the k-folds a neural network model is trained and can be accessed by `models`.
```

?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+The actual loading, of the trained model stored in the list, is defined in {py:meth}`~columnflow.ml.MLModel.open_model` function.
+
+The awkward array `events` is loaded in chunks and contains the merged events of all folds.
+To filter out the test set, create a mask with `fold_indices`.
+
+If you want to preserve columns and write them out into a parquet file, append the columns to `events` using `set_ak_column` and return `events`.
```
pkausw, line 180:
```suggestion
If you want to preserve columns and write them out into a parquet file, append the columns to `events` using {py:func}`~columnflow.columnar_util.set_ak_column` and return `events`.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+All columns not present in {py:meth}`~columnflow.ml.MLModel.produces` are then filtered out.
+
+In the following example, the models prediction as well as the number of muons and electrons are saved, all the other columns in events are thrown away, since they are not present in `use`:
```
pkausw, line 183: > ...thrown away, since they are not present in `use`.

Don't you mean `produce`?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+    --calibrators calibrator_name \
+    --selector selector_name \
+    --dataset datasets_1,dataset_2,dataset_3,...
+
+law run cf.MLEvaluation \
+    --version v1 \
+    --ml-model test_model \
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators skip_jecunc
+```
```
pkausw, line 214: I would split this into two boxes (makes copying easier)

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+    --version your_current_version \
+    --ml-model test_model_name \
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators calibrator_name \
+    --selector selector_name \
+    --dataset datasets_1,dataset_2,dataset_3,...
+
+law run cf.MLEvaluation \
+    --version v1 \
+    --ml-model test_model \
```
pkausw, line 211: Shouldn't it be

```suggestion
    --ml-model test_model_name \
```

?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators skip_jecunc
```
pkausw, line 213: Shouldn't this be

```suggestion
    --calibrators calibrator_name
```

?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+law run cf.MLTraining \
+    --version your_current_version \
+    --ml-model test_model_name \
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators calibrator_name \
+    --selector selector_name \
+    --dataset datasets_1,dataset_2,dataset_3,...
+
+law run cf.MLEvaluation \
+    --version v1 \
```
pkausw, line 210: Shouldn't this be

```suggestion
    --version your_current_version \
```

?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+    --ml-model test_model \
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators skip_jecunc
+```
+Most of these settings should sound familiar, if not look into the corresponding tutorial.
+`version` defines a setup configuration of your ML task, think more of a label than of an actual `version`.
+If you change the label, you
```
pkausw, line 217: Yes?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+
+# Optional useful functions:
+## Separate training and evaluation configuration for configs, calibrators, selector and producers:
+By default chosen configs, calibrators, selector and producer are used for both training and evaluation.
+Sometimes one do not want to share the same environment or does not need all the columns in evaluation as in training, this is where a separation of both comes in handy.
```
pkausw, line 222:
```suggestion
Sometimes one does not want to share the same environment or does not need all the columns in evaluation as in training.
This is where a separation of both comes in handy.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+
+To separate this behavior one need to define the training_{configs,calibrators, selector,producers}.
+These function take always the `config_inst` as first, and the requested_{configs,calibrators, selector,producers} as second parameter.
```
pkausw, line 225:
```suggestion
These functions take always the `config_inst` as first, and the requested_{configs,calibrators,selector,producers} as second parameter.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+If this function is defined the evaluation will use the give `config`,`calibrator`, `selector` or `producer`, while the training will use one defined in the function.
+
+In the following case, training will use a fixed selector and producer called default, while the calibrator uses an own defined calibrator called "skip_jecunc":
```
pkausw, line 228:
```suggestion
In the following case, training will use a fixed selector and producer called `default`, and a custom-defined calibrator called `skip_jecunc`:
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+```
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def training_calibrators
+:end-at: return ["skip_jecunc"]
+```
+## setup
+Setup is called at the end of `__init__` of your ML model.
+Within this function you can prepare operations that should stay open during the whole life time of your instance.
+An typical example for this would the opening of a file or dynamic appending of variables that are only relevant to your ML model, and no where else.
```
pkausw, line 248:
```suggestion
A typical example for this would be the opening of a file or dynamically appending variables that are only relevant to your ML model, and no where else.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+In the following example plotting variables are added to the ML model:
```
pkausw, line 249:
```suggestion
In the following example definitions needed to plot variables that are created in ML model are added to the `config_inst`:
```

docs/user_guide/sandbox.md
```diff
@@ -0,0 +1,58 @@
+
```
pkausw, line 1: File is not included in the toc of the user_guide page

pkausw (changes requested):
Nice PR @Bogdan-Wiederspan ! I have a couple of propositions, but I think this will greatly improve the understanding of the ML-related parts in columnflow :+1:

pkausw:
Also, can you fix the linting errors in this PR @Bogdan-Wiederspan ? Thanks!

docs/user_guide/ml.md
Bogdan-Wiederspan: I wanted to say that the person should save all ML related results and plots within the same directory as the related models instance.

But, you are right, I had a stroke mid sentence. Here a new suggestion.

```suggestion
It is of good practice to store your "machine-learning-instance" files within the directory of the models instance. To get the path to this directory use `task.target`.
```

docs/user_guide/ml.md
Bogdan-Wiederspan: Yes, it could be a good idea (from a didactic point of view) to move the filtering of the data to `open_input_files`, let `prepare_events` only handle thewhole preprocessing, if there is some.

docs/user_guide/ml.md
Bogdan-Wiederspan: yes :(
```suggestion
In the following example, the models prediction as well as the number of muons and electrons are saved, all the other columns in events are thrown away, since they are not present in `produce`:
```

docs/user_guide/ml.md
Bogdan-Wiederspan: that happens when one copy and paster :D.

Either of them, they only need to match.

docs/user_guide/ml.md
```diff
@@ -1,8 +1,258 @@
 # Machine Learning
 
-In this section, the users will learn how to implement machine learning in their analysis with
-columnflow.
+In this section, the users will learn how to implement machine learning in their analysis with columnflow.
```
Bogdan-Wiederspan, line 3:
```suggestion
In this section, the users will learn how to implement machine learning in their analysis with columnflow.

# How training happens in Columnflow: K-fold cross validation 
Machine learning in columnflow is implemented in a way that k-fold cross validation is enabled by default.
In k-fold cross validation the dataset is split in k-parts of equal size.
For each training, k-1 parts are used for the actual training and the remaining part is used for validation of the model.
This process is repeated k-times, resulting in the training of k-model instances.
In the end of the training columnflow will save all k-models, which are then usable for evaluation. 
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
 
-Note: which tasks, how to trigger them, how to train, test and plot networks and outputs
+# Configure your custom machine learning class:
+To create a custom machine learning (ML) class in columnflow, it is imperative to inherit from the MLModel class.
+This inheritance ensures the availability of functions to manage and access config and model instances, as well as the necessary producers.
```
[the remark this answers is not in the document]
Bogdan-Wiederspan: I added in the beginning an explanation about
What cross validation is, and also explained shortly what an models instance is.

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+:language: python
+:start-at: test_model
+:end-at: test_model
+```
+The second argument in {py::meth}`~columnflo.util.derive` is a `cls_dict` which configure your subclass.
+The `cls_dict` needs to be flat.
+The keys of the dictionary are set as class attributes and are therefore also accessible by using self.
+The configuration with derive has 2 main advantages:
+- manageability, since the dictionary can come from loading a config file and these can be changed fairly easy
+- flexibility, multiple settings require only different configuration files
```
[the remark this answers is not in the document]
Bogdan-Wiederspan:
```suggestion
- ==manageability==, since the dictionary can come from loading a config file and these can be changed fairly easy
- ==flexibility==, multiple settings require only different configuration files
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,258 @@
...
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def datasets
+:end-at: return set(dataset_inst)
+```
+
+
+## produces:
+By default, intermediate columns are not saved within the columnflow framework but are filtered out afterwards.
+If you want to prevent certain columns from being filtered out, you need to tell columnflow by writing them in the {py:meth}`~columnflow.ml.MLModel.produces`function.
```
Bogdan-Wiederspan, line 79:
```suggestion
If you want to prevent certain columns from being filtered out, you need to tell columnflow.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+This is always done by writing an iterable containing the name of the producer as string.
```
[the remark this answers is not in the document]
Bogdan-Wiederspan: Yes! I rewrote it to be more clear.

```suggestion
You can tell columnsflow, by return the names of all columns that should be preserved, as strings within an interable, in {py:meth}`~columnflow.ml.MLModel.produces`.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+More information can be found in the official documentation about [producers](building_blocks/producers).
+
+In the following example, I want to preserve the number of muons and electrons, we also want to preserve the output of the neural network.
+To avoid confusion, we are not producing the columns in this function, we only tell columnflow to not throwing them away.
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def produces
+:end-at: return preserved_columns
```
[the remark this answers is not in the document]
Bogdan-Wiederspan: see above, added an explanation about this in the beginning.

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+```
+ATTENTION: Sometimes one also want to preserve the used input and target features, for example to make plots.
+Remember that you already have a copy of all the input and target features in your parquet files used for training.
+In General it is not advised to create a new separate copy of these, otherwise it can be convenient to have everything within one place.
```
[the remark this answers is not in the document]
Bogdan-Wiederspan: It is a bad habit of me, coming from my masters time. I saved the models output together with my input and target features to further analyze them with ROOT. 
Resulting in wasting at least a factor 2x in disk space.

I just wanted to tell people, that they can actually save also the input and target feature together, but also wanted to discourage them to do so. 

I think I will deleted this part entirely.

docs/user_guide/ml.md
```diff
@@ -1,8 +1,258 @@
...
+More information can be found in the official documentation about [producers](building_blocks/producers).
+
+In the following example, I want to preserve the number of muons and electrons, we also want to preserve the output of the neural network.
+To avoid confusion, we are not producing the columns in this function, we only tell columnflow to not throwing them away.
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def produces
+:end-at: return preserved_columns
+```
+ATTENTION: Sometimes one also wants to preserve the used input and target features, for example to make plots.
```
Bogdan-Wiederspan, line 90:
```suggestion
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,258 @@
...
+Remember that you already have a copy of all the input and target features in your parquet files used for training.
```
Bogdan-Wiederspan, line 91:
```suggestion
```

docs/user_guide/ml.md
Bogdan-Wiederspan: These are helper functions I suggest people should design, thus they are not part of CF. This was the reason why I have no further explanation of these functions.

docs/user_guide/ml.md
```diff
@@ -1,8 +1,264 @@
...
-columnflow.
+In this section, the users will learn how to implement machine learning in their analysis with columnflow.
 
-Note: which tasks, how to trigger them, how to train, test and plot networks and outputs
+# How training happens in Columnflow: K-fold cross validation
+Machine learning in columnflow is implemented in a way that k-fold cross validation is enabled by default.
+In k-fold cross validation the dataset is split in k-parts of equal size.
+For each training, k-1 parts are used for the actual training and the remaining part is used for validation of the model.
+This process is repeated k-times, resulting in the training of k-model instances.
+In the end of the training columnflow will save all k-models, which are then usable for evaluation.
```
Bogdan-Wiederspan, line 10:
```suggestion
In the end of the training columnflow will save all k-models, which are then usable for evaluation.
An overview and further details about possible variations of k-fold cross validation can be found in the [sci-kit documentation](https://scikit-learn.org/stable/modules/cross_validation.html).
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,264 @@
...
+:language: python
+:start-at: test_model
+:end-at: test_model
+```
+The second argument in {py::meth}`~columnflow.util.DerivableMeta.derive` is a `cls_dict` which configures your subclass.
+The `cls_dict` needs to be flat.
+The keys of the dictionary are set as class attributes and are therefore also accessible by using `self`.
+The configuration with `derive` has two main advantages:
+- ==manageability==, since the dictionary can come from loading a config file and these can be changed fairly easy
+- ==flexibility==, multiple settings require only different configuration files
```
Bogdan-Wiederspan, line 26:
```suggestion
- ==flexibility==, multiple settings require only different configuration files
A possible configuration and model initialization for such a model could look like this:
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,264 @@
...
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: test_model
+:end-at: test_model
+```
+The second argument in {py::meth}`~columnflow.util.DerivableMeta.derive` is a `cls_dict` which configures your subclass.
+The `cls_dict` needs to be flat.
+The keys of the dictionary are set as class attributes and are therefore also accessible by using `self`.
+The configuration with `derive` has two main advantages:
+- ==manageability==, since the dictionary can come from loading a config file and these can be changed fairly easy
```
Bogdan-Wiederspan, line 25:
```suggestion
- manageability, since the dictionary can come from loading a config file and these can be changed fairly easy
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,264 @@
...
+- ==flexibility==, multiple settings require only different configuration files
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: hyperparameters =
+:end-at: test_model = TestModel.derive(
+```
+One can also simply define class variables within the model.
+This is can be useful for attributes that don't change often, for example to define the datasets your model uses.
+Since these are also class attributes, they are accessible by using `self` and also shared between all instances of this model.
```
Bogdan-Wiederspan, line 34:
```suggestion
Since these are also class attributes, they are accessible by using `self` and are also shared between all instances of this model.
```

docs/user_guide/ml_code.py
```diff
@@ -0,0 +1,280 @@
...
+ak = maybe_import("awkward")
+tf = maybe_import("tensorflow")
+np = maybe_import("numpy")
+
+
+law.contrib.load("tensorflow")
+
+
+class TestModel(MLModel):
+
```
Bogdan-Wiederspan, line 28:
```suggestion
# shared between all model instances
```

docs/user_guide/ml_code.py
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    def __init__(
+            self,
+            *args,
+            folds: int | None = None,
+            **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        # your instance variables
```
Bogdan-Wiederspan, line 43:
```suggestion
        # your instance variables
        # these are exclusive to your model instance
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,264 @@
...
+:language: python
+:start-at: def datasets
+:end-at: return set(dataset_inst)
+```
+
+
+## produces:
+By default, intermediate columns are not saved within the columnflow framework but are filtered out afterwards.
+If you want to prevent certain columns from being filtered out, you need to tell columnflow.
+You can tell columnsflow, by return the names of all columns that should be preserved, as strings within an interable, in {py:meth}`~columnflow.ml.MLModel.produces`.
```
Bogdan-Wiederspan, line 86:
```suggestion
You can tell columnflow, by let {py:meth}`~columnflow.ml.MLModel.produces` return the names of all columns that should be preserved, do this by define the names as strings within an interable.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+```
+
+
+## produces:
+By default, intermediate columns are not saved within the columnflow framework but are filtered out afterwards.
+If you want to prevent certain columns from being filtered out, you need to tell columnflow by writing them in the {py:meth}`~columnflow.ml.MLModel.produces`function.
+This is always done by writing an iterable containing the name of the producer as string.
+More information can be found in the official documentation about [producers](building_blocks/producers).
+
+In the following example, I want to preserve the number of muons and electrons, we also want to preserve the output of the neural network.
```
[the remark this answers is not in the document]
Bogdan-Wiederspan: Yes. This is a bad habbit of mine. I actually like to save the predictions within the same array like the given input, and target features. The main reason is to plot metrics to judge the predictive power. But I understand where the confusion comes from. I will rewrite that part to be more direct.

docs/user_guide/ml.md
```diff
@@ -1,8 +1,264 @@
...
+```
+
+
+## produces:
+By default, intermediate columns are not saved within the columnflow framework but are filtered out afterwards.
+If you want to prevent certain columns from being filtered out, you need to tell columnflow.
+You can tell columnsflow, by return the names of all columns that should be preserved, as strings within an interable, in {py:meth}`~columnflow.ml.MLModel.produces`.
+More information can be found in the official documentation about [producers](building_blocks/producers).
+
+In the following example, I want to preserve the number of muons and electrons, we also want to preserve the output of the neural network.
```
Bogdan-Wiederspan, line 89:
```suggestion
In the following example, I want tell columnflow to preserve the output of the neural network, but also the fold indices, that are used to create the trainings and test fold.
I do not store the input and target columns of that fold to save disk space, since they are already stored in the training set parquet file.

```

docs/user_guide/ml_code.py
```diff
@@ -0,0 +1,280 @@
...
+    def produces(self, config_inst: od.Config) -> set[Route | str]:
+        # mark columns that you don't want to be filtered out
+        ml_predictions = {f"{self.cls_name}.fold{fold}.{feature}"
+            for fold in range(self.folds)
+            for feature in self.target_columns}
+
+        util_columns = {f"{self.cls_name}.fold_indices"}
+
+        preserved_columns = ml_predictions | util_columns
+        return preserved_columns
```
Bogdan-Wiederspan, line 91:
```suggestion
    def produces(self, config_inst: od.Config) -> set[Route | str]:
        # mark columns that you don't want to be filtered out
        # save the networks prediction of a specific feature) for each fold
        # cls_name would be the name of your model
        ml_predictions = {f"{self.cls_name}.fold{fold}.{feature}"
            for fold in range(self.folds)
            for feature in self.target_columns}
       # save indices used to create the folds 
        util_columns = {f"{self.cls_name}.fold_indices"}

        # combine all columns to a unique set 
        preserved_columns = ml_predictions | util_columns
        return preserved_columns
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,266 @@
...
+```
+Thus, the choice to include these columns is your choice.
+
+## uses:
+In `uses` you define the columns that are needed by your machine learning model, and are forwarded to the ML model during the execution of the various tasks
+In this case we want to request the input and target features, as well as some weights:
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: def uses(
+:end-at: return used_columns
```
Bogdan-Wiederspan, line 106:
```suggestion
:end-at: return columns
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,267 @@
...
+
+law run cf.MLEvaluation \
+    --version your_current_version \
+    --ml-model test_model_name \
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators calibrator_name
+```
+Most of these settings should sound familiar, if not look into the corresponding tutorial.
+`version` defines a setup configuration of your ML task, think more of a label than of an actual `version`.
+If you change the label, you
```
Bogdan-Wiederspan, line 226:
```suggestion
If you change the version label, columnflow will rerun all dependencies that are unique for this label, this typically just means you will retrain a new model. You can then switch freely between both models version.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,267 @@
...
+
+# Optional useful functions:
+## Separate training and evaluation configuration for configs, calibrators, selector and producers:
+By default chosen configs, calibrators, selector and producer are used for both training and evaluation.
+Sometimes one does not want to share the same environment or does not need all the columns in evaluation as in training.
```
Bogdan-Wiederspan, line 231:
```suggestion
Sometimes one does not want to share the same environment or does not need all the columns in evaluation as in training.
Another possible scenario is the usage of different selectors or datasets, to be able to explore different phase spaces. 
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,269 @@
...
+For each training, k-1 parts are used for the actual training and the remaining part is used for validation of the model.
+This process is repeated k-times, resulting in the training of k-model instances.
+In the end of the training columnflow will save all k-models, which are then usable for evaluation.
+An overview and further details about possible variations of k-fold cross validation can be found in the [sci-kit documentation](https://scikit-learn.org/stable/modules/cross_validation.html).
 
-TODO
+# Configure your custom machine learning class:
+To create a custom machine learning (ML) class in columnflow, it is imperative to inherit from the {py:class}`~columnflow.ml.MLModel` class.
+This inheritance ensures the availability of functions to manage and access config and model instances, as well as the necessary producers.
+The name of your custom ML class can be arbitrary, since `law` accesses your machine learning model using a `cls_name` in {py::meth}`~columnflow.util.DerivableMeta.derive`, e.g.
```
pkausw, line 16:
```suggestion
The name of your custom ML class can be arbitrary, since `law` accesses your machine learning model using a `cls_name` in {py:meth}`~columnflow.util.DerivableMeta.derive`, e.g.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,269 @@
...
+```{literalinclude} ./ml_code.py
+:language: python
+:start-at: test_model
+:end-at: test_model
+```
+The second argument in {py::meth}`~columnflow.util.DerivableMeta.derive` is a `cls_dict` which configures your subclass.
```
pkausw, line 22:
```suggestion
The second argument in {py:meth}`~columnflow.util.DerivableMeta.derive` is a `cls_dict` which configures your subclass.
```

docs/user_guide/ml.md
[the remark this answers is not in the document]
pkausw: OK I guess, but you didn't actually do this, correct? Why?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,269 @@
...
+The `cls_dict` needs to be flat.
+The keys of the dictionary are set as class attributes and are therefore also accessible by using `self`.
+The configuration with `derive` has two main advantages:
+- manageability, since the dictionary can come from loading a config file and these can be changed fairly easy
+- flexibility, multiple settings require only different configuration files
+A possible configuration and model initialization for such a model could look like this:
```
pkausw, line 28: Add another line break, otherwise this is part of the previous list!

```suggestion

A possible configuration and model initialization for such a model could look like this:
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,269 @@
...
+In the {py:meth}`~columnflow.ml.MLModel.train` function, you implement the initialization of your models and the training loop.
+The `task` corresponding to the models training is {py:class}`~columnflow.tasks.ml.MLTraining`.
+By default {py:meth}`~columnflow.ml.MLModel.train` has access to the location of the models `inputs` and `outputs`.
+
+In columnflow, k-fold cross validation is enabled by default.
+The `self` argument in {py:meth}`~columnflow.ml.MLModel.train` referes to the instance of the fold.
+Using `self`, you have also access to the entire `analysis_inst`ance the `config_inst`ance of the current fold, and to all the derived parameters of your model.
+
+With this information, you can call and prepare the columns to be used by the model for training.
+In the following example a very simple dummy trainings loop is performed using the Keras fit function.
```
pkausw, line 145:
```suggestion
In the following example a very simple dummy training loop is performed using the Keras fit function.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+```
+It is recommended to start the path with `bash::`, to indicate that you want to source the {py:meth}`~columnflow.ml.MLModel.sandbox` with `bash`.
+How to actual write the setup and requirement file can be found in the section about [setting up a sandbox](building_blocks/sandbox).
+
+## datasets:
+In the {py:meth}`~columnflow.ml.MLModel.datasets` function, you specify which datasets are important for your machine learning model and which dataset instance should be extracted from your config.
+To use this function your datasets needs to be added to your campaign, as defined by the [Order](https://python-order.readthedocs.io/en/latest/>) Module.
+An example can be found [here](https://github.com/uhh-cms/cmsdb/blob/d83fb085d6e43fe9fc362df75bbc12816b68d413/cmsdb/campaigns/run2_2018_nano_uhh_v11/top.py).
+It is recommended to return this as `set`, to prevent double counting.
+In the following example all datasets given by the {ref} external config <init_of_ml_model> are taken, but also an additional is given.
```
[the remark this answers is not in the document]
pkausw: This is still not resolved. Alternatively, you can also use `one` instead of another `dataset`.

docs/user_guide/ml.md
pkausw: Can you split this into two boxes?

docs/user_guide/ml.md
```diff
@@ -1,8 +1,269 @@
...
+    --calibrators calibrator_name \
+    --selector selector_name \
+    --dataset datasets_1,dataset_2,dataset_3,...
+
+law run cf.MLEvaluation \
+    --version your_current_version \
+    --ml-model test_model_name \
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators calibrator_name
+```
```
pkausw, line 224: Can you split this into two boxes?

docs/user_guide/ml_code.py
```diff
@@ -0,0 +1,289 @@
+# coding: utf-8
```
pkausw, line 1: Might make sense to move this to a dedicated example directory, so users can also find and access this example

pkausw (changes requested):
Thanks @Bogdan-Wiederspan ! Just some additional comments

docs/user_guide/ml.md
```diff
@@ -1,8 +1,269 @@
...
 
-In this section, the users will learn how to implement machine learning in their analysis with
-columnflow.
+In this section, the users will learn how to implement machine learning in their analysis with columnflow.
 
-Note: which tasks, how to trigger them, how to train, test and plot networks and outputs
+# How training happens in Columnflow: K-fold cross validation
+Machine learning in columnflow is implemented in a way that k-fold cross validation is enabled by default.
+In k-fold cross validation the dataset is split in k-parts of equal size.
+For each training, k-1 parts are used for the actual training and the remaining part is used for validation of the model.
```
Bogdan-Wiederspan, line 8:
```suggestion
For each training, k-1 parts are used for the actual training and the remaining part is used to test the model.
```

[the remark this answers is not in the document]
Bogdan-Wiederspan:
```suggestion
For each training, k-1 parts are used for the actual training and the remaining part is used to test the model.
```

docs/user_guide/ml.md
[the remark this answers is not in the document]
Bogdan-Wiederspan: I simply forgot to add this. To the batch, also = WORD = does not highlight in this markdown engine. I will just highlight with bold instead of HTML mark.
```suggestion
- **manageability**, since the dictionary can come from loading a config file and these can be changed fairly easy
- **flexibility**, multiple settings require only different configuration files
```

docs/user_guide/ml.md
Bogdan-Wiederspan:
````suggestion
law run cf.MLTraining \
    --version your_current_version \
    --ml-model test_model_name \
    --config run2_2017_nano_uhh_v11_limited \
    --calibrators calibrator_name \
    --selector selector_name \
    --dataset datasets_1,dataset_2,dataset_3,...
```
```bash
law run cf.MLEvaluation \
    --version your_current_version \
    --ml-model test_model_name \
    --config run2_2017_nano_uhh_v11_limited \
    --calibrators calibrator_name
```
````

File: docs/user_guide/ml_code.py (hunk 53 above)
Comment (Bogdan-Wiederspan) [reply docs/user_guide/ml_code.py:1]: done, moved ml_code.py to an dir called example. Will switch all links to this file to the new directory in the next commit.

File: docs/user_guide/sandbox.md [hunk 55]
```diff
@@ -0,0 +1,58 @@
...
+## Where to Store the Corresponding Files
+
+For organizational purposes, it is recommended to store your setup and requirement files in `$CF_BASE/sandboxes/`.
+It is of good practice to follow the naming conventions used by Columnflow.
+Begin your setup file with "venv," for example, `venv_YOUR_ENVIRONMENT.sh`, and use only the environment name (without "venv") for your requirement file, e.g., `YOUR_ENVIRONMENT.txt`.
+
+## The Setup File
+
+Begin your setup file by referencing an existing setup file within the `$CF_BASE/sandboxes/` directory.
```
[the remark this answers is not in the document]
Bogdan-Wiederspan: This time they should just orient themself by looking at this file and copy the file over.
I think it is a good idea to give the user a default setup file in the default analysis repo. 
```suggestion
Begin your setup file by referencing an existing setup file within the `$ANALYSIS_BASE/sandboxes/` directory.
```

docs/user_guide/ml.md
```diff
@@ -1,8 +1,254 @@
...
+```bash
+law run cf.MLTraining \
+    --version your_current_version \
+    --ml-model test_model_name \
+    --config run2_2017_nano_uhh_v11_limited \
+    --calibrators calibrator_name \
+    --selector selector_name \
+    --dataset datasets_1,dataset_2,dataset_3,...
+
+law run cf.MLEvaluation \
```
Bogdan-Wiederspan:
````suggestion
law run cf.MLTraining \
    --version your_current_version \
    --ml-model test_model_name \
    --config run2_2017_nano_uhh_v11_limited \
    --calibrators calibrator_name \
    --selector selector_name \
    --dataset datasets_1,dataset_2,dataset_3,...
```
```bash
law run cf.MLEvaluation \
````

File: docs/user_guide/ml.md (hunk 14 above)
Comment (Bogdan-Wiederspan) [reply docs/user_guide/ml.md]: already done. This didnt showed up here because i marked the code block by myself, which counted as NEW comment. On this new comment the suggestion was applied to.

File: docs/user_guide/ml.md [hunk 57]
```diff
@@ -1,8 +1,271 @@
...
+It is recommended to start the path with `bash::`, to indicate that you want to source the {py:meth}`~columnflow.ml.MLModel.sandbox` with `bash`.
+How to actually write the setup and requirement files can be found in the section about [setting up a sandbox](building_blocks/sandbox).
+
+## datasets:
+In the {py:meth}`~columnflow.ml.MLModel.datasets` function, you specify which datasets are important for your machine learning model and which dataset instance(s) should be extracted from your config.
+To use this function your datasets needs to be added to your campaign, as defined by the [Order](https://python-order.readthedocs.io/en/latest/) Module.
+An example can be found [here](https://github.com/uhh-cms/cmsdb/blob/d83fb085d6e43fe9fc362df75bbc12816b68d413/cmsdb/campaigns/run2_2018_nano_uhh_v11/top.py).
+It is recommended to return this as `set`, to prevent double counting.
+In the following example all datasets given by the {ref}`external config <init_of_ml_model>` are taken, but also an additional is given.
+Note how the name of each dataset is used to get a `dataset instance` from your `config instance`.
```
Bogdan-Wiederspan, line 77:
```suggestion
In the following example all datasets given by the {ref} external config <init_of_ml_model> are taken, but also an additional dataset is given.
```

pkausw (approved):
Very nice, thanks @Bogdan-Wiederspan !

PR #375, opened 2023-12-22
allow passing custom plot-function to PlotCutflowVariables1D

### LINKED ISSUE #367: plot-function parameter not usable in PlotCutflowVariables1D
In the PlotCutflowVariables1D task, the `plot-function` parameter is currently not usable. Instead, we provide two hard-coded plotting functions based on the `--per-plot` parameter [1]. We should allow custom plotting functions here and resolve the correct default based on the `steps` parameter only if no plotting function has been set.

[1] https://github.com/columnflow/columnflow/blob/master/columnflow/tasks/cutflow.py#L582

mafrahm:
Closes #367


pkausw (approved):
LGTM :+1:

PR #376, opened 2023-12-22
allow customizing ylim via general_settings

mafrahm:
With these changes, one can customize the amount of whitespace created in a plot and the number of magnitudes in the y-range (for log scale).
Example:

```
law run cf.PlotVariables1D --version v1 --variables jet1_pt --processes tt_sl --remove-output 0,a,y --general-settings whitespace_fraction=0.1,magnitudes=2,yscale=log
```

Here is an example with whitespace_fraction=0.1 and magnitudes=6
plot__proc_ggHH_kl_1_kt_1_sl_hbbhww__cat_incl__var_mli_mbb

Of course, we could also implement these options as indendent task parameters.


columnflow/plotting/plot_all.py
```diff
@@ -135,6 +135,8 @@ def plot_all(
     shape_norm: bool = False,
     skip_legend: bool = False,
     cms_label: str = "wip",
+    whitespace_fraction: float = 0.3,
+    magnitudes: float = 4,
     **kwargs,
 ) -> plt.Figure:
     """
```
pkausw, line 142: Can you also add the new keyword arguments to the doc string?

columnflow/plotting/plot_all.py
```diff
@@ -192,8 +194,8 @@ def plot_all(
     # some default ylim settings based on yscale
     log_y = style_config.get("ax_cfg", {}).get("yscale", "linear") == "log"
 
-    ax_ymin = ax.get_ylim()[1] / 10**4 if log_y else 0.00001
-    ax_ymax = get_position(ax_ymin, ax.get_ylim()[1], factor=1.4, logscale=log_y)
+    ax_ymin = ax.get_ylim()[1] / 10**magnitudes if log_y else 0.0000001
```
pkausw, line 197: Why the change for the offset? Is there a specific reason for this?

pkausw (changes requested):
Thanks for the PR @mafrahm ! Just some minor changes necessary imho. I don't have a strong opinion about how to implement these new options specifically, but I think it's fine like this

columnflow/plotting/plot_all.py
mafrahm: It's not related to the rest of the PR, I think I just had some issue with the offset in my own analysis at some point because signal event yields were on the same order of magnitude as this aribrary offset

pkausw (approved):

PR #377, opened 2023-12-22
tests for columnflow.util

mafrahm:
I also fixed a minor issue in `pattern_matcher`, where regex expressions were considered as patterns as soon as the regex inlcudes a `*`

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -5,10 +5,126 @@ __all__ = ["UtilTest"]
 
 import unittest
 
-from columnflow.util import create_random_name
+from columnflow.util import (
+    create_random_name, maybe_import, MockModule, DotDict, Derivable,
+    safe_div, test_float, test_int, is_regex, is_pattern, pattern_matcher,
+)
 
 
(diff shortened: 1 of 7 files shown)


tests/test_util.py
```diff
@@ -5,10 +5,126 @@
...
+        })
+        self.assertEqual(my_dict.A, 1)
+        self.assertEqual(my_dict.C.D, 3)
+        my_dict = DotDict.wrap({
+            "A": 1,
+            "B": 2,
+            "C": [3, 4],
+        })
+        self.assertEqual(my_dict.C, [3, 4])
+
```
pkausw, line 103: Might make sense to also test that you can _not_ create a DotDict from something other than a dictionary-like type

pkausw (changes requested):
Thanks for this nice set of unit tests @mafrahm ! Just one comment, it might also make sense to add more checks (potentially in a future PR) for things that should _not_ be possible. Also, there seems to be an issue with the coverage test for this PR, can you resolve this?

pkausw (approved):
LGTM

PR #366, opened 2023-11-29
Merge Plotting guide into document_sprint branch

mafrahm:
This PR:
- adds a first draft of a plotting guide
- adds a "Task Overview" section
- updates the docs sandbox requirements (we can now add mermaid graphs and dropdowns)

PR #381, opened 2024-01-12
Added `Writing new tasks` and `Custom law.cfg file` section to docs

haddadanas:


pkausw:
Side note: This should probably be merged after PR #366

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+### The requirements:
+
+#TODO explain branches if not explained in law
+
+As noticed in the [code snipped](#example-task-class), a task can define three different requirement keywords, which have different functionality in the framework.
+
+The ```requires()``` methode in a task, which is inherited from {external+law:py:class}`law.task.base.Task`, defines all required outputs for a task to run on a branch level.
+If an output does not exist, the task responsible for this output will be scheduled and executed before the desired task is run.
+
+The {external+law:py:meth}`~law.tasks.ForestMerge.workflow_requires()` is analog to the former methode, however it defines the dependencies on a workflow level, which can be very useful, when working remotely and running jobs on remote clusters.
```
pkausw, line 119:
```suggestion
The {external+law:py:meth}`~law.tasks.ForestMerge.workflow_requires()` is analog to the former method.
However, it defines the dependencies on a workflow level, which can be very useful when working remotely and running jobs on remote clusters.
```

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+Every task defines a workflow, which contains all available branches of this task.
+When executing the workflow, the task will run in all (specified) branches.
+If the desired workflow requires the output of another workflow, the latter will be executed beforehand, which means that all needed branches (with no existing output) from the required workflow will be executed before any of the branches from the current workflow runs (Workflow runs vertically).
+This behaviour can be modified, by deactivating the ```pilot``` parameter.
+In this case, the workflow first run through the branches of the executed task after each other, where all dependencies of each branch are executed when needed by the branch (workflow propagates horizontally).
+This is very useful, when parallelising jobs on remote clusters, since multiple branches can run simultaneously, without the need to run all branches from one task beforehand.
+
+Last, the class attribute ```reqs```, that is a {py:class}`~columnflow.tasks.framework.base.Requirements` object, does not directly have to do with the other two requirements, but is rather for analyses specific configuration.
```
pkausw, line 127:
```suggestion
Finally, the class attribute ```reqs```, that is a {py:class}`~columnflow.tasks.framework.base.Requirements` object, does not directly have to do with the other two requirements, but is rather for analysis-specific configuration.
```

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+Strictly speaking, it is not needed for the task to run, however it offers a container for the requirements, which becomes very convenient, when for example a required task should be replaced by a customized subclass of the same task.
```
pkausw, line 128:
```suggestion
Strictly speaking, it is not needed for the task to run.
However it offers a container for the requirements, which becomes very convenient when for example a required task should be replaced by a customized subclass of the same task.
```

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+Instead of having to edit the requirements in all the class methods that depends on this requirement, the ```reqs``` attribute can be simply modified with:
+```python
+MyTask.reqs.RequiredTask = ModifiedRequiredTask
+```
+Therefore, using the class attribute ```reqs```, when working with the requirements within the class is conveninet.
```
pkausw, line 133:
```suggestion
Therefore, using the class attribute ```reqs```, when working with the requirements within the class is convenient.
```

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+
+As noticed in the [code snipped](#example-task-class), a task can define three different requirement keywords, which have different functionality in the framework.
+
+The ```requires()``` methode in a task, which is inherited from {external+law:py:class}`law.task.base.Task`, defines all required outputs for a task to run on a branch level.
+If an output does not exist, the task responsible for this output will be scheduled and executed before the desired task is run.
+
+The {external+law:py:meth}`~law.tasks.ForestMerge.workflow_requires()` is analog to the former methode, however it defines the dependencies on a workflow level, which can be very useful, when working remotely and running jobs on remote clusters.
+Every task defines a workflow, which contains all available branches of this task.
+When executing the workflow, the task will run in all (specified) branches.
+If the desired workflow requires the output of another workflow, the latter will be executed beforehand, which means that all needed branches (with no existing output) from the required workflow will be executed before any of the branches from the current workflow runs (Workflow runs vertically).
```
pkausw, line 122: Maybe some sort of illustration to show what `runs vertically` or `runs horizontally` means would be helpful here

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+
+![req plot](cf_plot.svg "image")
+
+### RunOnceTask
+
+The {external+law:py:class}`~law.tasks.RunOnceTask` is a special task that is used to run a task only once, which can be very useful in many situations.
+For example, when a task is required by multiple tasks or multiple branches of a workflow, but the task itself should only be executed once independent of the branches, or a task does not produce any file output but is needed to be marked as done (e.g. a task only produces text output on the CLI).
+
+This is implemented by (additionally) inheriting from {external+law:py:class}`~law.tasks.RunOnceTask`.
+Moreover, to mark the task as complete, the {external=law:py:meth}`~law.tasks.RunOnceTask.mark_complete()` method is to be added at the end of the ```run()``` method after the logic of the task is executed, or the complete```run()``` method is either to be wrapped by the {external=law:py:meth}`~law.tasks.RunOnceTask.complete_on_success` decorator.
```
pkausw, line 144:
```suggestion
Moreover, to mark the task as complete, the {external+law:py:meth}`~law.tasks.RunOnceTask.mark_complete()` method is to be added at the end of the ```run()``` method after the logic of the task is executed, or the complete ```run()``` method is either to be wrapped by the {external+law:py:meth}`~law.tasks.RunOnceTask.complete_on_success` decorator.
```

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+
+After creating a new task, make sure to add the new task to the law config file in your analysis folder, in order to run the task.
+The new task has to be specified under the ```[outputs]``` section as follows:
+
+```ini
+[outputs]
+
+lfn_sources: wlcg_fs_desy_store
+
+cf.MyNewTask: wlcg
```
pkausw, line 177: I think you only need the `cf.` prefix if you actually implement something within the columnflow name space. In praxis, I think users wouldn't want to do that though, right?

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+```
+### Executing the Task
+
+after sourcing the law environment, the task can be executed using the command:
+```bash
+law run cf.MyTask --param1 foo
+```
+The parameters can be given using the syntax ```--param1 foo```, where ```param1``` is the name of the parameter and ```foo``` is the value of the parameter.
+If the parameter's name include an underscore, the underscore is replaced by a hyphen (```--param_1 foo``` -> ```--param-1 foo```).
+If the parameter is a boolean, the parameter can be set to ```True``` by only typing ```--param-boolean```.
```
pkausw, line 188: It might be safer to use `param_boolean` in your example above, just to make sure that the user doesn't _need_ to append `boolean` to a parameter in order to make it a boolean (sounds dumb, but better save than sorry)

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+If the parameter is a list, the values can be given as a comma-separated list (```--param2 foo,bar,baz```).
 
 ## Custom law.cfg file for users without grid certificate (CMS specific?)
 
-TODO
+Users, who do not have a grid certificate, are not able to run jobs on the grid or store and fetch data from it.
+However, they can still run jobs locally.
+For this purpose, a custom law config file should be created in the analysis folder, which specifies a local dcache and the path to the nanoAoDs for every required campaing, and set the output locations for the tasks to local.
```
pkausw, line 195:
```suggestion
For this purpose, a custom law config file should be created in the analysis folder, which specifies a local dcache and the path to the nanoAODs for every required campaign, and set the output locations for the tasks to local.
```

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+It is recommended to create a copy of the default law config file and modify it as needed and then redirect the setup to the custom config file by setting the ```LAW_CONFIG_FILE``` environment variable in the `setup.sh` file to the path of the custom config file as follows:
```
pkausw, line 196:
```suggestion
It is recommended to create a copy of the default law config file, modify it as needed and then redirect the setup to the custom config file by setting the ```LAW_CONFIG_FILE``` environment variable in the `setup.sh` file to the path of the custom config file as follows:
```

I'm not sure if I would agree with the statement that the user should copy the initial law config file. Rather, the user should create a new file, inherit from the original law config file, and then overwrite the properties he/she needs to change. This is also what you show in your example, right? So I don't think you need to say anything about copying something.

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+
+```bash
+export LAW_CONFIG_FILE="${LAW_CONFIG_FILE:-${ANALYSIS_DIR}/law.nocert.cfg}"
+```
+
+The costum config file should contain the following sections:
```
pkausw, line 202: I would first define the nocert config and then show how to include this in the setup.sh file

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+[wlcg_campaigns_name]
+base: path/to/local/campaigns/nanoAOD/directory
+
+# define additional settings related to caching when working with this campaign 
+use_cache: $CF_WLCG_USE_CACHE
+cache_root: $CF_WLCG_CACHE_ROOT
+cache_cleanup: $CF_WLCG_CACHE_CLEANUP
+cache_max_size: 15GB
+cache_global_lock: True
+cache_mtime_patience: -1
```
pkausw, line 224: I wouldn't write about `dcache` since this is also very (DESY) specific. Maybe just talk about "defining different file systems"?

docs/user_guide/examples.md
pkausw, line 202: add link to config documentation? https://law.readthedocs.io/en/latest/config.html

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+
+# whether or not the ensure_proxy decorator should be skipped, even if used by task's run methods
+skip_ensure_proxy: True
+
+# complementary custom defined modules, which will only be included, when using the custom config file
+selection_modules: analysis.selection.{python_module_name}
+
+[outputs]
+
+lfn_sources: local_desy_dcache
```
pkausw, line 236:
```suggestion
lfn_sources: local_dcache
```

?

pkausw (changes requested):
Thanks for this nice PR @haddadanas ! Just some minor comments in line. Also, I suggest to 
- pull the latest version of the documentation_sprint_v2 branch
- maybe rename the section from `Example` to something more descriptive, like `Creating your own task`?

docs/user_guide/examples.md
```diff
@@ -1,14 +1,248 @@
...
+All task bases can be found under {py:mod}`~columnflow.tasks.framework`.
+The choice of the superclass highly depends on the purpose of the task.
+However, one needs to ensure to inherit from all necessary {py:mod}`~columnflow.tasks.framework.mixins` classes to provide all needed keywords and parameters.
+For example, if you have a task that needs a calibrator or depends on a task that requires a calibrator, you'll need to have the CalibratorMixin as a superclass, so that the calibrator parameter can be forwarded through the workflow tree.
+
+After making the choice of the superclass(es), the class is implemented at its core by the following methods (refer to the [luigi task documentation](https://luigi.readthedocs.io/en/stable/tasks.html) for more information on the methods):
+
+- {external+law:py:meth}`~law.task.base.BaseTask.run()`: This method contains the logic for executing the task.
+- `requires()`: This method returns a collection of tasks that this task depends on.
+- `output()`: This method returns a collection of output targets for the task.
```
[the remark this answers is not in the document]
haddadanas: Referenced the methodes in ForestMerge and WrapperTask, which have the same API BaseTask would have

docs/user_guide/examples.md
haddadanas: The `req plot`, which was not rendered correctly is the illustration for this :)

haddadanas:
Update:
added small API correction for the `DotDict` class in `columnflow.util`

pkausw (approved):
LGTM, thanks @haddadanas !

Issue #327, opened 2023-10-20
Add documentation for API: tasks


pkausw:
Add API documentation for module: `tasks`

pkausw:
closed by #386

Issue #326, opened 2023-10-20
Add documentation for API: plotting


pkausw:
Add API documentation for module: `plotting`

pkausw:
Closed by #366

pkausw:
closed by #386

Issue #325, opened 2023-10-20
Add documentation for API: ml


pkausw:
Complete API documentation for module `ml`

pkausw:
closed by #386

PR #386, opened 2024-01-17
Include rudimentary API docs for all modules

pkausw:
As discussed in our last meeting, we need to include all modules in the API docs.

This PR does that. Note that not all modules have meaningful doc strings or type hints yet.
Nevertheless, the updated version should give at least an overview of the modules and their member function/variables, which would be beneficial for all users.


pkausw:
Closes #325
Closes #326 
Closes #327

mafrahm (changes requested):
I think the `index.rst` for the plotting modules is still missing. Otherwise, this looks good to me.

pkausw:
> I think the `index.rst` for the plotting modules is still missing. Otherwise, this looks good to me.

The file is uploaded here: https://github.com/columnflow/columnflow/blob/features/api_docs/docs/api/plotting/index.rst (I think it doesn't show as part of the PR because it may have existed before?)

columnflow/tasks/framework/mixins.py
```diff
@@ -10,13 +10,12 @@
...
 import itertools
 from collections import Counter
-from typing import Iterable
 
 import luigi
 import law
 import order as od
 
-from columnflow.types import Sequence, Any
+from columnflow.types import Sequence, Any, Dict, Iterable, Tuple, Union, List
```
mafrahm, line 18: Why are we using these classes (`Dict`, `Tuple`, `List`) for typing instead of the default python types

mafrahm (approved):
In that case, I agree that we can merge this. Just left one more inline comment for clarification, but no changes are necessary.

columnflow/tasks/framework/mixins.py
pkausw: I think there was an issue with the typing not too long ago, which prompted us to centralize the type definitions in cf (see PR #305 ). I simply follow this approach here, though this needs to be followed-up on for other modules as far as I'm aware.

PR #390, opened 2024-02-01
generalize jer for pt-dependent SFs

mafrahm:
This small PR generalizes jer for Run3, since input variables for evaluating are now taken directly from the `evaluator` instead of being hard-coded.

It has been tested that results stay consistent on an exemplary Run2 file via the `cf_diff` script

### DIFF
--- a/columnflow/calibration/cms/jets.py
+++ b/columnflow/calibration/cms/jets.py
@@ -634,8 +634,8 @@ def jer(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
     # JER scale factors and systematic variations
     jersf = {}
     for syst in ("nom", "up", "down"):
-        variable_map["systematic"] = syst
-        inputs = [variable_map[inp.name] for inp in self.evaluators["sf"].inputs]
+        variable_map_syst = dict(variable_map, systematic=syst)
+        inputs = [variable_map_syst[inp.name] for inp in self.evaluators["sf"].inputs]
         jersf[syst] = ak_evaluate(self.evaluators["sf"], *inputs)
 
     # array with all JER scale factor variations as an additional axis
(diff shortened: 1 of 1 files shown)


columnflow/calibration/cms/jets.py
```diff
@@ -617,23 +620,23 @@ def jer(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
...
-        syst: ak_evaluate(
-            self.evaluators["sf"],
-            events.Jet.eta,
-            syst,
-        )
-        for syst in ("nom", "up", "down")
-    }
+    jersf = {}
+    for syst in ("nom", "up", "down"):
+        variable_map["systematic"] = syst
```
dsavoiu, line 637: This does the job but it's a bit messy IMO to update the `variable_map` from outside to loop in this way. Just in case in some future version of the code we decide to rely on the `variable_map` further down, it might lead to errors since the map ends up with the `systematic` key set to "down". It might be better to create a local copy adding the `systematic` key, e.g. something like this:
```suggestion
        variable_map = dict(variable_map, systematic=syst)
```

dsavoiu (changes requested):
Hi @mafrahm, thanks a lot for the PR! This looks good to me and should allow both pT-dependent and independent JER SFs to be used with the same code. It also nicely paves the way for future generalizations along the same lines in case we will ever need these for e.g. JECs as well :)

I just have one minor suggestion below.

columnflow/calibration/cms/jets.py
mafrahm: Good point, added with the latest commit

dsavoiu (approved):
Great, thanks for implementing the changes! It looks all good to me, I think we can go ahead with the merge now.

PR #393, opened 2024-02-06
implement pu_weight producer using correctionlib

mafrahm:


columnflow/production/cms/pileup.py
```diff
@@ -18,13 +20,98 @@
...
+        "NumTrueInteractions": indices,
+    }
+
+    for column_name, syst in (
+        ("pu_weight", "nominal"),
+        ("pu_weight_minbias_xs_up", "up"),
+        ("pu_weight_minbias_xs_down", "down"),
+    ):
+        # get the inputs for this type of variation
+        variable_map_syst = dict(variable_map, systematic=syst)
```
riga, line 52:
```suggestion
        variable_map_syst = {**variable_map, "systematic": syst}
```

Feels a bit more verbose.

columnflow/production/cms/pileup.py
```diff
@@ -18,13 +20,98 @@
...
+    self: Producer,
+    reqs: dict,
+    inputs: dict,
+    reader_targets: InsertableDict,
+) -> None:
+    """
+    Loads the pileup calculator from the external files bundle and saves them in the
+    py:attr:`pileup_corrector` attribute for simpler access in the actual callable.
+    """
+    logger.warning(
```
riga, line 85: I think we can remove the warning. If the producer results in the same weight vector (or even a fixed one in case the previous implementation was buggy), this would require explicit communication.

columnflow/production/cms/pileup.py
```diff
@@ -55,8 +142,8 @@ def pu_weight_requires(self: Producer, reqs: dict) -> None:
     reqs["pu_weights"] = CreatePileupWeights.req(self.task)
 
 
-@pu_weight.setup
-def pu_weight_setup(self: Producer, reqs: dict, inputs: dict, reader_targets: InsertableDict) -> None:
+@pu_weights_from_columnflow.setup
+def pu_weights_from_columnflow_setup(self: Producer, reqs: dict, inputs: dict, reader_targets: InsertableDict) -> None:
```
riga, line 146: Can you re-indent this?

riga (changes requested):
Nice work! This should be far more stable and also faster than the current way of deriving weights, with almost nothing to configure.

Just three super minor things inline.

columnflow/production/cms/pileup.py
pkausw: Just a comment from the side line:  the logger warning is the result of a discussion @mafrahm and I had offline. We concluded that it would be good to at least temporarily make users aware that the calculation of the pu weights has changed - we had talked about fading this out from one release to another or maybe leave it in for a couple of months just so that people have the chance to realize that this has changed. Personally, I'm still in favor for this, but I also don't have too strong feelings either. So if we decide to not show this at all, I can also live with that

mafrahm: I also have no strong opinion on this. I added that it is still recommended to use this Producer.
My reasoning for this warning was to make people aware of a (probably minor) source of inconsistency in results between versions of Columnflow when using this Producer. But I'd also be fine with removing the warning.

riga: Fine by me :+1:
I was mainly worried about this warning showing up too often (e.g. when running the producer locally for tasks with a lot of branches / files). I'd suggest to use `logger.warning_once(message_id: str, message: str)`.

riga: I will quickly add that to go ahead with this PR.

riga (approved):

PR #396, opened 2024-02-13
updates to the CreateYieldTable task

mafrahm:
- adds a `suffix` parameter
- adds a json output with raw event yields
- adds $ symbols around the yields for latex tables

### DIFF
--- a/columnflow/tasks/yields.py
+++ b/columnflow/tasks/yields.py
@@ -57,6 +57,10 @@ class CreateYieldTable(
         description="string parameter to define the normalization of the yields; "
         "choices: '', per_process, per_category, all; empty default",
     )
+    suffix = luigi.Parameter(
+        default=law.NO_STR,
+        description="Adds a suffix to the output name of the yields table; empty default",
+    )
 
     # upstream requirements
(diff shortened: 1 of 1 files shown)


columnflow/tasks/yields.py
```diff
@@ -57,6 +57,10 @@ class CreateYieldTable(
         description="string parameter to define the normalization of the yields; "
         "choices: '', per_process, per_category, all; empty default",
     )
+    suffix = luigi.Parameter(
```
riga, line 60: Can we name this `output_suffix`? I think such a suffix could be useful in many other tasks as well (e.g. plotting, but this should not be part of *this* PR of course), and since there are a couple other possible use cases for "suffixes", we could name this unambiguously right away.

riga (changes requested):
Looks good :+1: 

Just a minor inline comment regarding the name of the suffix parameter.

columnflow/tasks/yields.py
mafrahm: good point, it's changed now

riga (approved):

PR #400, opened 2024-02-27
add top_pt_weight producer

mafrahm:
This PR adds a new Producer for top pt reweighting. The reweighting is split into two parts:

- a Producer that reads and stores the gen top pt
- a Producer that uses the gen top pt to calculate and store the top pt weights.

In that way, a user is flexible which part to call during or after the selection.

The only configuration necessary for the top pt reweighting are the top pt reweighting parameters:

```python
    # https://twiki.cern.ch/twiki/bin/viewauth/CMS/TopPtReweighting#TOP_PAG_corrections_based_on_dat?rev=31
    cfg.x.top_pt_reweighting_params = {
        "a": 0.0615,
        "a_up": 0.0615 * 1.5,
        "a_down": 0.0615 * 0.5,
        "b": -0.0005,
        "b_up": -0.0005 * 1.5,
        "b_down": -0.0005 * 0.5,
    }
```

We could also provide these parameters as a default when not provided by the analysis.

The up and down variations are currently calculated by varying a and b in the same direction simultaneously, but we could also implement these uncertainties separately.

### DIFF
--- a/columnflow/production/cms/muon.py
+++ b/columnflow/production/cms/muon.py
@@ -63,13 +63,27 @@ def muon_weights(
     abs_eta = flat_np_view(abs(events.Muon.eta[muon_mask]), axis=1)
     pt = flat_np_view(events.Muon.pt[muon_mask], axis=1)
 
+    variable_map = {
+        "year": self.year,
+        "abseta": abs_eta,
+        "pt": pt,
+    }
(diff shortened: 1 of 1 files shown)


riga (approved):
Approved after reiterating privately.

PR #404, opened 2024-03-01
allow masking events via preparation producer

mafrahm:
With this small PR, it is possible to mask events used in MLTraining via the ml preparation producer.

Example:
```python
@producer(
    uses={some_categorizer},
)
def ml_preparation(
    self: Producer,
    events: ak.Array,
    stats: dict = {},
    fold_indices: ak.Array | None = None,
    ml_model_inst: MLModel | None = None,
    **kwargs,
) -> ak.Array:
    """
    Producer that is run as part of PrepareMLEvents to apply a selection
    """
    if self.task.task_family == "cf.PrepareMLEvents":
        # pass category mask to not use the full phase space in training
        events, mask = self[some_categorizer](events, **kwargs)
        logger.info(f"Select {ak.sum(mask)} from {len(events)} events for MLTraining")
        return events[mask]

    return events
```
    
The issue was that the mask was only applied to the `events` array, but not to the `fold_indices`. With this change, the mask will also be applied to the indices.
Since the `fold_indices` are now part of the `events` array, we could also remove it as a keyword argument


mafrahm:
After a discussion with @pkausw, we came to the conclusion that it might be better to pass a Categorizer instead of a Producer as part of the PrepareMLEvents. In that way, we can return a mask which can be applied by the task itself instead of doing that ourselves in the Producer.

The issue with this preparation function is that it is used for lots of different purposes (potentially applying a selection, producing columns and gathering stats). It is therefore not clear what would fit best.

@riga

riga (approved):
Looks good to me.

There is of course a potential for ambiguity as to what task array function should be triggered here. But from the cf side - calling a producer and only accepting `events` - this looks reasonable. Everything beyond that, i.e., what exactly happens inside the producer, should be up to the downstream analysis to decide.

PR #405, opened 2024-03-01
request only one workflow per variable in CreateDatacards

mafrahm:
Until now, the `CreateDatacards` task has simply requested one `MergeShiftedHistograms` workflow per category.

This PR updates the workflow requires from `CreateDatacards` such that only one `MergeShiftedHistograms` workflow is requested per dataset.
This might lead in producing some histograms for shifts that were not requested, but the runtime increase due to that should be rather small.


riga (approved):
Approved after reiterating privately.

PR #416, opened 2024-03-28
add overflow and underflow variable setting

### LINKED ISSUE #402: Overflow Bins for Plotting
Suggestion for adding overflow and underflow bins for the plotting of histograms. Ideally, this could be added as an additional parameter to choose whether or not these should be displayed. These would be especially useful for distributions with long tails.

mafrahm:
This PR adds the possibility to move entries from the overflow (underflow) bin into the last (first) visible bin of the histogram during plotting. This can be enabled by adding the `overflow` (`underflow`) auxiliary to the variable inst or via variable_settings, e.g.

```
--variable-settings var_name,overflow
```

Closes #402


riga:
@mafrahm Could you resolve the conflicts that appeared in the meantime? With that, everything looks good to me and we'd be ready to merge :+1:

riga (approved):
LGMT!

PR #419, opened 2024-04-19
Feature/slice histograms

### LINKED ISSUE #374: support slicing of histograms during plotting tasks
Similar to the `rebin` setting, we should implement slicing of histograms. To implement this, we need to enable passing multiple values per setting, since slicing usually requires a lower and upper bound. Both integers and complex numbers should be supported (*).
One possible syntax for passing multiple values per setting could look something like this:

```
--variable-settings var_name,slice=10j;20
```

where the `;` (to be discussed what symbol to use) would be another delimiter that needs to be resolved as part of the SettingsParameter: https://github.com/columnflow/columnflow/blob/master/columnflow/tasks/framework/parameters.py#L14

Note: how should we handle a combination of `rebin` and `slice`? First slicing, then rebinning? Or making it customizable somehow?

(*) reminder:
`h[slice(10, 20)]` means to select 10 bins starting at bin number 10 and ending with bin number 19.
`h[slice(10j, 20j)]` means to select all bins with x values greater than 10 and smaller than 20.
Combinations of integers and complex numbers are also possible, e.g. `h[slice(10j, 20)]`

mafrahm:
This PR adds the `slices` variable setting that allows slicing a histogram during Plotting. Example:

Exemplary task call from hh2bbww analysis:
```
law run cf.PlotVariables1D --variable-settings "jet1_pt,slice=5;15" --version prod1 --variables jet1_pt  --processes ggHH_kl_1_kt_1_dl_hbbhww --remove-output 0,a,y
```
This selects all bins from bin 5 to bin 15. You could also select bins from some x_min to x_max range by passing e.g. `--variable-settings "jet1_pt,slice=50j;200j"`

plot__proc_ggHH_kl_1_kt_1_dl_hbbhww__cat_incl__var_jet1_pt

This only slices the histogram itself but does not modify the x_min and x_max range of the plot since they are taken from the variable inst, but this can be modified via `--variable-settings "jet1_pt,x_min=50,x_max=200"`. Already having this functionality makes this PR a bit pointless, but it also adds some functionality to the SettingsParameters and slicing might become useful for other use cases.

Closes #374

### DIFF
--- a/columnflow/tasks/framework/parameters.py
+++ b/columnflow/tasks/framework/parameters.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import law
 
 from columnflow.util import try_float, try_complex, DotDict
+from columnflow.types import Iterable
 
 
 class SettingsParameter(law.CSVParameter):
@@ -32,7 +33,7 @@ class SettingsParameter(law.CSVParameter):
         pair = setting.split("=", 1)
         key, value = pair if len(pair) == 2 else (pair[0], "True")
         if ";" in value:
-            # split by semicolon and parse each value
+            # split by ";" and parse each value
             value = tuple(cls.parse_value(v) for v in value.split(";"))
         else:
             value = cls.parse_value(value)
@@ -51,7 +52,8 @@ class SettingsParameter(law.CSVParameter):
         return value
 
     @classmethod
-    def serialize_setting(cls, name: str, value: str) -> str:
+    def serialize_setting(cls, name: str, value: str | Iterable[str]) -> str:
+        value = ";".join(str(v) for v in law.util.make_tuple(value))
         return f"{name}={value}"
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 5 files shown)


columnflow/tasks/framework/parameters.py
```diff
@@ -31,16 +32,28 @@ class SettingsParameter(law.CSVParameter):
     def parse_setting(cls, setting: str) -> tuple[str, float | bool | str]:
         pair = setting.split("=", 1)
         key, value = pair if len(pair) == 2 else (pair[0], "True")
+        if ";" in value:
```
pkausw, line 35: Would it make sense to make the delimiter (in this case `;`) a member variable of the class?

pkausw (approved):
LGTM, thanks!

columnflow/tasks/framework/parameters.py
mafrahm: Yes! In that way we do not need to update the tests when changing this! I'll change this tomorrow

pkausw: Nice, thanks!

pkausw (approved):
Thanks for the changing, this looks excellent now :+1:

Issue #408, opened 2024-03-15
Jet corrector for different clustering algorithms


apaasch:
Currently the jet calibrator is implemented for the standard jet collection AK4(CHS/Puppi) with `event.Jets`.
There are analyses, however, that use a different type of jet collections such as AK8 or XCone jets.
A more general jet calibrator would be a great enhancment to allow those analyses to also use CF.

In case of XCone, it is sufficient to apply the centerally provided jet corrections onto a jet collection named differently than `event.Jets`.

PR #421, opened 2024-04-25
allow passing different b-score columns to btag_weights Producer

mafrahm:
This PR allows passing a different b-tagging discriminator to the `btag_weights` Producer


columnflow/production/cms/btag.py
```diff
@@ -135,6 +140,35 @@ def add_weight(syst_name, syst_direction, column_name):
     return events
 
 
+def get_b_score_column(btag_config: tuple[str, list, str]) -> str:
```
pkausw, line 143: I think if you support the old structure `tuple[str, list]` you should also form the type hints accordingly, i.e.

```suggestion
def get_b_score_column(btag_config: tuple[str, list, str] | tuple[str, list]) -> str:
```

columnflow/production/cms/btag.py
```diff
@@ -44,12 +48,13 @@ def btag_weights(
     *get_btag_file* can be adapted in a subclass in case it is stored differently in the external
     files.
 
-    The name of the correction set as well as a list of JEC uncertainty sources which should be
-    propagated through the weight calculation should be given as an auxiliary entry in the config:
+    The name of the correction set, a list of JEC uncertainty sources which should be
+    propagated through the weight calculation, and the column used for b-tagging should
+    be given as an auxiliary entry in the config:
```
pkausw, line 53: This does not cover your automatic handling of the column name needed for a given correction. Can you ammend this either in this docstring or the docstring of `get_b_score_column`?

columnflow/production/cms/btag.py
```diff
@@ -135,6 +140,35 @@ def add_weight(syst_name, syst_direction, column_name):
     return events
 
 
+def get_b_score_column(btag_config: tuple[str, list, str]) -> str:
+    """
+    Helper function to resolve the btag score column from the btag configuration.
+    """
```
pkausw, line 146: This docstring is not super helpful tbh. It would be better to also define the parameters and what is returned, e.g. 

```suggestion
    """
    Helper function to resolve the btag score column from the btag configuration.
    
    :param btag_config: Entry in auxiliary `config_inst.x.btag_sf`, see example :py:meth:`~columflow.production.cms.btag.btag_weights`. If tuple has less than 3 entries, the column name is derived from the name of the correction set.
    :returns: Name of column that is required for the calculation of this set of corrections.
    """
```

columnflow/production/cms/btag.py
```diff
@@ -210,7 +249,3 @@ def btag_weights_setup(
     )
     corrector_name = self.get_btag_config()[0]
     self.btag_sf_corrector = correction_set[corrector_name]
-
-    # check versions
-    if self.btag_sf_corrector.version not in (3,):
-        raise Exception(f"unsuppprted btag sf corrector version {self.btag_sf_corrector.version}")
```
pkausw, line 216: Why did we need this in the first place? Might this break something again when we remove this?

pkausw (changes requested):
Very nice feature @mafrahm , thanks a lot! I think this would also close #408 . Just some minor comments, should be easy to fix :+1:

pkausw (approved):
LGTM, thanks!

PR #422, opened 2024-05-02
add producers parameter to ProduceColumnsWrapper

mafrahm:


pkausw (approved):
Nice extension! I think similar extended Wrappers would make sense, not in the scope of this PR though. In any case, this gives an excellent example of how users can do it themselves

PR #424, opened 2024-05-02
Make `categorizer_map` regular `dict` to prevent silent `KeyErrors`.

dsavoiu:
This PR addresses an issue encountered in an analysis using dynamically defined categories.

Instead of the usual definition in the `init` method of a `TaskArrayFunction`, the categories were added dynamically in the `__call__` method just before the `category_ids` producer was called. Since the category instances were not present in the config at initialization time, the corresponding entries in the `category_ids.categorizer_map` were missing. This in turn should have caused a `KeyError` while looking up the categorizers for the dynamically added categories during `category_ids.__call__`, but this was suppressed since the `categorizer_map` was defined as a `defaultdict`, leading to the dynamically added categories being assigned the default (all-true) map and thus matching every event.

The above scenario is now prevented by making `categorizer_map` a regular `dict`, triggering a `KeyError` as expected.

### DIFF
--- a/columnflow/production/categories.py
+++ b/columnflow/production/categories.py
@@ -89,3 +89,6 @@ def category_ids_init(self: Producer) -> None:
             self.produces.add(categorizer)
 
             self.categorizer_map[cat_inst].append(categorizer)
+
+    # cast to normal dict to prevent silent failures on KeyError
+    self.categorizer_map = dict(self.categorizer_map)


mafrahm (approved):
I agree with this change since it helps catching errors in an analysis faster.

Regarding the use case of adding category instances during a call of some TaskArrayFunction, was this done on purpose or by mistake? Because currently I think there is now way around having to setup categories either directly in the config or in some init function since otherwise the category will not be known to the config when calling e.g. a plotting task.

dsavoiu:
Thanks for the review @mafrahm! The dynamic setup of categories in the call method was indeed an error and has since been moved to the init function. But this did not raise any exceptions, so we needed a while to figure out why the categories were not behaving as intended. With this fix there will now at least be a `KeyError` if this is attempted, so hopefully similar mistakes will be caught early on.

I'll go ahead and merge this then.

PR #426, opened 2024-05-06
multiply cferr columns with nominal btag weight

mafrahm:
For the `cferr` btag uncertainties, we only multiplied the SFs of all jets with flavor == 4.
Instead, we need to multiply the SFs of all jets with flavor == 4 with the nominal SFs we got from jets with flavor != 4.

This gets is especially clear when looking at events without any jets with flavor == 4. Previous to this fix, these events get assigned some btag_weight != 1, while the cferr variations are assigned a weight of 1.


pkausw (approved):
LGTM, thanks!

PR #427, opened 2024-05-06
Catch cases where rawFactor is NAN in jec calibrator.

jomatthi:
This PR adds a line setting the `Jet.rawFactor` to 0.0 in case its calculation results in NAN in the CMS jec calibrator. This can happen for jets with a very low pT.

### DIFF
--- a/columnflow/calibration/cms/jets.py
+++ b/columnflow/calibration/cms/jets.py
@@ -299,7 +299,8 @@ def jec(
     # apply full jet correction
     events = set_ak_column_f32(events, "Jet.pt", events.Jet.pt_raw * jec_factors)
     events = set_ak_column_f32(events, "Jet.mass", events.Jet.mass_raw * jec_factors)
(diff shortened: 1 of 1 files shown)


pkausw:
Hi @jomatthi , thanks for the PR! It would be very helpful to see an example of an event where this happens, just to document the issue a little better for the future. Can you provide a small example here?

dsavoiu:
Hi @pkausw,

if I remember correctly, @jomatthi was running into this edge case in the mttbar analysis. The way it happens there is probably due to how we do jet-lepton cleaning, which is by subtracting the four-vectors of matched leptons from the jet four-vector. In a small number of cases, the resulting jet four-vector after the cleaning is unphysical (the mass component turns imaginary), so we set the components manually to zero. The jets need to be recorrected after this step, which leads to a division by zero in the calibrator code when computing the `rawFactor`.

pkausw (approved):
Ah ok, thanks for clarifying @dsavoiu ! LGTM then, I'll proceed to merge. Thanks @jomatthi for the PR!

riga:
@all-contributors please add @jomatthi for code

PR #432, opened 2024-05-15
HistogramsUser base and inspection tasks

mafrahm:
This PR implements base tasks for loading histograms in Columnflow + two tasks that implement these base tasks for debugging purposes and as a simple exemplary usecase.

This task call loads histograms for only the nominal shift
```
law run cf.InspectHistograms --version v1 --variables jet1_pt --shift nominal --processes tt_sl
```

This task call loads histograms for nominal, murf_envelope_up, and murf_envelope_down
```
law run cf.InspectShiftedHistograms --version v1 --variables jet1_pt --shift-sources murf_envelope --processes tt_sl
```

### DIFF
--- a/columnflow/tasks/framework/histograms.py
+++ b/columnflow/tasks/framework/histograms.py
@@ -1,0 +1,154 @@
+# coding: utf-8
+
+"""
+Base tasks for working with multiple merged histograms.
+"""
+
+
+import law
+
+from columnflow.tasks.framework.base import Requirements, ShiftTask
+from columnflow.tasks.framework.mixins import (
+    CalibratorsMixin, SelectorStepsMixin, ProducersMixin, MLModelsMixin, WeightProducerMixin,
+    VariablesMixin, DatasetsProcessesMixin, CategoriesMixin,
+    ShiftSourcesMixin,
+)
+from columnflow.tasks.histograms import MergeHistograms, MergeShiftedHistograms
+from columnflow.util import dev_sandbox, maybe_import
+
+ak = maybe_import("awkward")
+
+
+class HistogramsUserBase(
+    DatasetsProcessesMixin,
+    CategoriesMixin,
+    VariablesMixin,
+    MLModelsMixin,
+    WeightProducerMixin,
+    ProducersMixin,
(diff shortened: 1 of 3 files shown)


columnflow/tasks/framework/histograms.py
```diff
@@ -0,0 +1,154 @@
...
+    def store_parts(self):
+        parts = super().store_parts()
+        parts.insert_before("version", "datasets", f"datasets_{self.datasets_repr}")
+        return parts
+
+    def load_histogram(self, dataset, variable):
+        histogram = self.input()[dataset]["collection"][0]["hists"].targets[variable].load(formatter="pickle")
+        return histogram
+
+    def slice_histogram(self, histogram, processes, categories, shifts, reduce_axes: bool = False):
```
pkausw, line 43: These are very useful function for a variaty of applications! It would be very nice if you could add doc strings and type hints in these two functions, since they are also intended to be used by users in the end

pkausw (changes requested):
Looks good, thanks @mafrahm ! I think this will make e.g. evaluating ABCD methods much more feasible. It would be great if you could implement the small changes that I mention inline 😬

columnflow/tasks/framework/histograms.py
```diff
@@ -0,0 +1,192 @@
...
+        processes: str | list[str],
+        categories: str | list[str],
+        shifts: str | list[str],
+        reduce_axes: bool = False,
+    ) -> hist.Hist:
+        """
+        Slice a histogram by processes, categories, and shifts.
+
+        This function takes a histogram and slices it based on the specified processes, categories, and shifts.
+        It returns the sliced histogram.
```
pkausw, line 73: Hm, I'm not sure how familiar users are with the concept of 'slicing a histogram' (I for one am not sure if I understand it correctly :D). Maybe we can be slightly more specific here:

```suggestion
        Slice a histogram by processes, categories, and shifts.

        This function takes a histogram and slices it based on the specified processes, categories, and shifts, i.e. all  other axes are integrated out.
        It returns the sliced histogram.
```

pkausw (approved):
Thanks 👍

PR #434, opened 2024-05-21
add ml_model_settings parameter

mafrahm:
This PR adds the 'ml_model_settings' parameter that writes key-value pairs directly into the `self.parameters` attribute.

Example:
```python
class MyModel(MLModel):
    def __init__(
        self,
        *args,
        **kwargs,
    ):
        # we cannot cast to dict on command line, but one can do this by hand
        ml_process_weights = self.parameters.get("ml_process_weights", {"st": 1, "tt": 2})
        if isinstance(ml_process_weight, tuple):
            ml_process_weights = {proc: int(weight) for proc, weight in [s.split(":") for s in ml_process_weights]}

        # store parameters of interest in the ml_model_inst, e.g. via the parameters attribute
        self.parameters = {
            "batchsize": int(self.parameters.get("batchsize", 1024)),
            "layers": tuple(int(layer) for layer in self.parameters.get("layers, (64, 64, 64)),
            "ml_process_weights": ml_process_weights
        }

        # create representation of ml_model_inst
        self.parameters_repr = law.util.create_hash(sorted(self.parameters.items()))

```

These parameters can then be changed on command line
```
law run cf.MLTraining --version v1 --ml-model MyModel --ml-model-settings "batchsize=2048,layers=32;32;32,ml_process_weights=st:1;tt:4"
```

Open questions:

1.) At the moment, the `ml_model_settings` is not implemented for the `MLModelsMixin`, therefore we can only use the "default" model (or create new model via `derive`) when e.g. creating histograms. We could add a `ml_models_settings` parameter to this mixin, but I'm not sure if this would be used much.
2.) Since the output of the model is now hashed, we might want to also automatically produce an output of the parameters or the `parameters_repr` such that ml trainings can always be reproduced if necessary.

### DIFF
--- a/columnflow/ml/__init__.py
+++ b/columnflow/ml/__init__.py
@@ -152,6 +152,13 @@ class MLModel(Derivable):
         if "configs" in kwargs:
             self._setup(kwargs["configs"])
 
+    def __str__(self):
+        """
+        Returns a string representation of this model instance. The string is composed of the class
+        name and the string representation of all parameters.
+        """
+        return f"{self.cls_name}__{self.parameters_repr}"
+
     @property
     def config_inst(self: MLModel) -> od.Config:
         if self.single_config and len(self.config_insts) != 1:
@@ -184,6 +191,24 @@ class MLModel(Derivable):
 
         # any other case
         return str(value)
+
+    @property
+    def parameters_repr(self: MLModel) -> str:
+        """
+        Returns a hash of string representation of all parameters. This is used to uniquely identify
+        a model instance based on its parameters.
+
+        :raises: Exception in case the parameters_repr changed after it was set.
+        :returns: String representation of all parameters.
+        """
+        parameters_repr = law.util.create_hash(self._join_parameter_pairs(only_significant=True))
+        if hasattr(self, "_parameters_repr") and self._parameters_repr != parameters_repr:
+            raise Exception(
+                f"parameters_repr changed from {self._parameters_repr} to {parameters_repr};"
+                "this should not happen",
+            )
+        self._parameters_repr = parameters_repr
+        return self._parameters_repr
 
     def _join_parameter_pairs(self: MLModel, only_significant: bool = True) -> str:
         """
--- a/columnflow/ml/__init__.py
+++ b/columnflow/ml/__init__.py
@@ -157,7 +157,10 @@ class MLModel(Derivable):
         Returns a string representation of this model instance. The string is composed of the class
         name and the string representation of all parameters.
         """
-        return f"{self.cls_name}__{self.parameters_repr}"
+        model_str = f"{self.cls_name}"
+        if self.parameters_repr:
+            model_str += f"__{self.parameters_repr}"
+        return model_str
 
     @property
     def config_inst(self: MLModel) -> od.Config:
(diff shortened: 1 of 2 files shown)


columnflow/ml/__init__.py
```diff
@@ -185,6 +195,26 @@ def _format_value(self: MLModel, value: Any) -> str:
...
+        :returns: String representation of all parameters.
+        """
+        if not self.parameters:
+            return ""
+        parameters_repr = law.util.create_hash(self._join_parameter_pairs(only_significant=True))
+        if hasattr(self, "_parameters_repr") and self._parameters_repr != parameters_repr:
+            raise Exception(
+                f"parameters_repr changed from {self._parameters_repr} to {parameters_repr};"
+                "this should not happen",
+            )
```
pkausw, line 214: I lik ethe mechanism to check for potential changes in the architecture. However, is this not sensitive to permutation within the `ml-model-settings` parameter? E.g., `--ml-model-settings "batchsize=2048,layers=32;32;32"` and `--ml-model-settings "layers=32;32;32,batchsize=2048"` should yield the same architecture, but I think the hashes would be different, right? Have you tested this by chance? Perhaps it's better to first sort the tuple entries

columnflow/tasks/framework/mixins.py
```diff
@@ -1060,8 +1061,18 @@ class MLModelMixinBase(AnalysisTask):
...
+        default=DotDict(),
+        description="settings passed to the init function of the ML model",
+    )
+
     exclude_params_repr_empty = {"ml_model"}
 
+    @property
+    def ml_model_repr(self):
+        """Returns a string representation of the ML model instance."""
+        return str(self.ml_model_inst)
```
pkausw, line 1074: Wouldn't it be better to overwrite the `__str__` property instead?

pkausw (changes requested):
Nice PR, thanks! Just some minor questions.

I agree that having a mechanism that dumps the current parameter settings automatically would be nice. However, I think the user can do this themselves at the end of e.g. the training loop. We might want to think about providing e.g. a decorator that can be used directly for this end and that wraps e.g. the training function of the model inst. But imho, this isn't critical for this PR. (Though it also shouldn't be super hard, so if you have some spare time feel free to add it ;) )

columnflow/ml/__init__.py
mafrahm: Hi @pkausw, I agree that the order should not matter, however, the `self.parameters` has explicitly been initialized as an OrderedDict. Maybe @riga can give a comment whether there is a reason to order the parameters. Otherwise, I'd also be in favour of sorting the parameters before hashing (which I am already doing in my own analysis)

columnflow/tasks/framework/mixins.py
mafrahm: We are overwriting the `__str__` property of the ml_model, but I don't think we want to overwrite this property from the task itself as this is really only the representation of the ml model

columnflow/ml/__init__.py
pkausw: I agree, this seems to be the safer thing to do. The change is also minimal, so I guess it should be easy to change this if necessary 👍

PR #448, opened 2024-06-10
Option to remove reduced events after merging.

riga:
When merging reduced events, the output of `cf.MergeReducedEvents` contains the exact same event content as its input `cf.ReduceEvents`, though - obviously - merging into fewer files. Since this effectively doubles the file size, it should be possible to remove the outputs of `cf.ReduceEvents` right away after merging.

This PR adds an option `--keep-reduced-events` to `cf.MergeReducedEvents`. Its default can be controlled through the `law.cfg` and is set to `True` there, but can be changed by analyses.


mafrahm (approved):
Looks good to me!

PR #449, opened 2024-06-10
Refactor pinning of output versions and locations.

riga:
This is a rather extensive PR that refactors the way that locations and versions of specific outputs can be pinned through the `law.cfg` or the `config_inst`'s and `analysis_inst`'s auxiliary entries.

This would solve a long-standing discussion about sharing outputs among analyzers and also across analyses.

I added sections on the scope and purpose of this feature to the documentation already built the branch: https://columnflow.readthedocs.io/en/feature-pinned_versions_and_locations/user_guide/best_practices.html

---

### Summary for versions

- Pinned versions have priority over what is passed upstream via `Task.req(self)` as part of the overlap between `Task` and `self`, but not over class-level CLI parameters if given (e.g. `--cf.Task-version ...`).
- Versions are looked up (in this order) at
  - `config_inst.x.versions`
  - `analysis_inst.x.versions`
  - `[versions]` section in `law.cfg`
- The lookup logic is described below.

### Summary for locations

- Pinned locations are chosen when tasks are defined by `self.target()` (which then defers to `self.local_target()` or `self.wlcg_target()` with the correct arguments).
- For the lookup, as before, locations are defined in the `[outputs]` section of the `law.cfg`.
- The lookup logic is described below.

### Lookup logic

The lookup of either locations or versions happens through nested dictionaries which can be defined in the config, analysis or `law.cfg` file. When defined in the `law.cfg`, this nested dictionary is constructed by splitting the option names at double underscores `__`. Hence, these two variants would be identical:

```python
cfg.x.versions = {
    "run3_*": {
        "cf.CalibrateEvents": "v2",
    },
    "cf.CalibrateEvents": "v1",
}
```

```ini
[versions]

run3_*__cf.CalibrateEvents: v2
cf.CalibrateEvents: v1
```

Now, tasks can define the keys that can be used to find specific values in these nested dictionaries (which can be considered trees for easier nomenclature). This is done through a new method `get_config_lookup_keys()`, implemented by the analysis base task, and extended by inheriting classes. For most of our tasks though, the keys will contain (in this order):

1. analysis name
2. config name
3. task family
4. dataset name
5. shift name

The lookup is done according to these keys using three rules:

- Not all keys have to match and can be skipped if they don't match
  - e.g. in the example above, only the *config name* and *task family* are defined
- Keys (non-leaf-nodes) in the tree can be patterns or regular expressions
- The lookup order is depth-first-search
  - i.e., the first matching node is explored first, and the first matching value is returned.

This way, definitions of pinned versions and locations can be very flexible.

Some examples:

```python
cfg.x.versions = {
    "cf.CalibrateEvents": {
        "nominal": "prod1",        # nominal shift at prod1
        "^.*(up|down)$": "prod2",  # all variations at prod2
    },
    "run3_22pre_*": {
        "cf.SelectEvents": {
            "tt_sl_powheg": "prod1",  # ttbar sl at prod1 in 22pre era
            "*": "prod2",             # rest at prod2
        },
    },
    "cf.SelectEvents": "prod2",  # prod2 for all other eras
}
```

### DIFF
--- a/analysis_templates/cms_minimal/__cf_module_name__/config/analysis___cf_short_name_lc__.py
+++ b/analysis_templates/cms_minimal/__cf_module_name__/config/analysis___cf_short_name_lc__.py
@@ -256,13 +256,9 @@ cfg.x.keep_columns = DotDict.wrap({
     },
 })
 
-# versions per task family, either referring to strings or to callables receving the invoking
-# task instance and parameters to be passed to the task family
-cfg.x.versions = {
-    # "cf.CalibrateEvents": "prod1",
-    # "cf.SelectEvents": (lambda cls, inst, params: "prod1" if params.get("selector") == "default" else "dev1"),
-    # ...
-}
+# pinned versions
+# (see [versions] in law.cfg for more info)
+cfg.x.versions = {}
 
 # channels
 # (just one for now)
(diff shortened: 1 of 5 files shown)


mafrahm (approved):
LGTM

riga:
Thanks for the fast review @mafrahm

PR #452, opened 2024-06-11
add checks to ensure proper event and object masks

maadcoen:
I added checks to the SelectionResult class which ensure that the following conditions when passing event masks or object mask to it:
- event mask should be a 1d Array of booleans
- object masks should be:
  -  a 2d Array of booleans
  - a 2d or 1d Array of integers 

All mask are checked for the presence of Nones (raising an error if found) and converted to pure type (no ? or option in the type).

Integer object masks are converted to be
- two dimensional (if original 1d, the user is notified about the conversion)
- jagged (type is of form <N * var * ... >, not <N * M * ... >)

These checks avoid errors like the following:
- objects mask is 1d integer: the object mask will be interpreted as an event index instead of an object index
- object mask is 2d integer, but of fixed dimensions (not jagged): the object mask might again be interpreted as selecting events, not obejcts
- event mask is a single boolean. For instance, when doing smth like ak.sum(leptons.pdgId == 11) == 1, which wil sum over all electrons in the entire sample because the axis=1 is omitted. The mask will most probably evaluate to False and kill all your events.
- object mask is of mixed type in only one (part of a) datafile, but pure in others (because of rare none values), resulting in merging issues.

### DIFF
--- a/columnflow/selection/__init__.py
+++ b/columnflow/selection/__init__.py
@@ -243,20 +243,82 @@ class SelectionResult(od.AuxDataMixin):
     """
 
     @classmethod
-    def check_nones_and_convert(cls, name, mask):
-        mask_type = str(mask.type)
-        if "?" in mask_type or "Option" in mask_type:
-            assert not ak.any(ak.is_none(mask)), f"mask {name} contains None values"
-            logger.info(f"mask {name} is of mixed type, but does not contain Nones: converting to pure type.")
-            mask = ak.fill_none(mask, 0)
+    def _check_nones_and_convert(cls, name, mask, optiontype):
+        assert not ak.any(ak.is_none(mask, axis=-1)), f"mask {name} contains None values"
+        logger.info(f"mask {name} is of mixed type, but does not contain Nones: converting to pure type.")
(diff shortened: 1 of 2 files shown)


riga:
Thanks @maadcoen ! 

@mafrahm and I will have a look shortly 👌

maadcoen:
I have already a question myself how to properly inspect the type of an array to see whether it potentially contains None values. Currently I do the following:
```
mask_type = str(mask.type)
if "?" in mask_type or "Option" in mask_type:
         ...
```
which feels awkward, and I am also not sure if it is really correct.

riga:
In these cases, I would suggest to open a question (or check for existing ones) to the [awkward array discussions](https://github.com/scikit-hep/awkward/discussions).

maadcoen:
> In these cases, I would suggest to open a question (or check for existing ones) to the [awkward array discussions](https://github.com/scikit-hep/awkward/discussions). Someone usually responds extremely fast :)

I did, and based on their answer I changed the code as follows. It's complicated, but it seems to me that is unavoidable since awkward typing is quite complicated bussiness. 
I added also a file with tests that show what is allowed and what is not with this setup.

PR #455, opened 2024-06-11
Handle empty LHE scale and pdf weight columns.

haddadanas:


riga:
Thanks for the changes @haddadanas, looks good!

I just tested locally and I would have a single commit that I would like to add to this PR. Could you give me permissions to append to the branch on your fork?

riga (approved):

PR #456, opened 2024-06-11
bug fix: was trying to convert jagged arrays to numpy views

maadcoen:
The changes from commit [cfec386](https://github.com/columnflow/columnflow/commit/cfec38626e2a9c0bf454a4ed2336d28c740c932a) made my analysis crash: the histograms always bin in category_ids which are store in a jagged awkward array. Consequently, applying np.asarray to it fails. The changes below restore the original behaviour.


riga:
Hi @maadcoen , thanks for the PR!

We should indeed make sure that ragged shapes should be handled correctly :+1:

Just for my understanding, how do the arrays in the dictionary look your case? Is there one ragged one and the other ones are regularly shaped, or do they all have the same ragged shape?

Depending on that, I'd have a comment on your change of the values to be binned, which are now all move slightly to the left, whereas we only want this behavior for entries that end up *exactly* on the right most edge.r

maadcoen:
the dictionary always looks like the following:
```
data = {
    "category": generally jagged array, unless categories cover all events,
    "process": 1d regular,
    "shift": I am not sure about this one, I haven't worked with shifts so far,
    "weight": 1d regular,
    VARIABLE_NAME: can be jagged (e.g. Jet.pt) or regular (like missing pt)
}
```

columnflow/columnar_util.py
```diff
@@ -1201,16 +1201,14 @@ def fill_hist(
             if name not in data:
                 raise ValueError(f"missing data for histogram axis '{name}'")
 
-    # create numpy views for all data arrays
-    data = {name: np.asarray(data[name]) for name in data}
+    # create awkward views for all data arrays
+    data = {name: ak.Array(data[name]) for name in data}
```
riga, line 1205: This should be removed.

```suggestion
```

columnflow/columnar_util.py
```diff
@@ -1201,16 +1201,14 @@ def fill_hist(
...
 
     # correct last bin values
     for ax in correct_last_bin_axes:
         right_egde_mask = data[ax.name] == ax.edges[-1]
         if np.any(right_egde_mask):
-            # work on a copy to not change input values in-place
-            data[ax.name] = data[ax.name].copy()
-            data[ax.name][right_egde_mask] -= ax.widths[-1] * 1e-5
+            data[ax.name] = data[ax.name] - ax.widths[-1] * 1e-5
```
riga, line 1211: GitHub does not let me make a suggestion here for some reason, but the loop should be:

```
# correct last bin values
for ax in correct_last_bin_axes:
    right_egde_mask = ak.flatten(data[ax.name], axis=None) == ax.edges[-1]
    if np.any(right_egde_mask):
        data[ax.name] = data[ax.name].copy()
        flat_np_view(data[ax.name])[right_egde_mask] -= ax.widths[-1] * 1e-5
```

riga (changes requested):
I made two suggestions that should do the trick. Ragged ak array layouts are kept and in case there are values at the exact right most edge value, the values to be binned are copied and changed in place, but only at the specified values.

With that, we could go ahead and merge.

columnflow/columnar_util.py
maadcoen: ok, I see now the problem with what I had. It's been changed.

maadcoen: true, sorry! Fixed that now

riga:
I think something went wrong on your end in the commit history.
We also stumbled upon a case a couple minutes ago where we need to fill histograms with irregularly shaped input data, so I would quickly open a second PR that is essentially a copy of *this* PR with fixed history, and merge it right away :+1:

maadcoen:
ok, thanks!

PR #460, opened 2024-06-12
correct bug in normalization setup when no cross section is available for a process

nprouvost:
add process to lookup table only if process has a cross section


riga (approved):
Nice, thanks!

Issue #308, opened 2023-10-12
Resolving groups of plotting parameters general_settings and process_settings


mafrahm:
At the moment the resolving of groups for plotting parameters `general_settings` [1] and `process_settings` [2] only works, when giving a single group, e.g. ```--process-settings my_group``` resolves the group, but ```--process-settings my_group:other_process,some_setting=True``` does not resolve the group, but interprets `my_group` as another process.

It would be nice, if we could make the `resolve_config_default_and_groups` function [3] also usable for the SettingsParameter and MultiSettingsParameter.

[1] https://github.com/columnflow/columnflow/blob/5939cb0a1f70e909f2df6c46bbed593b873a19e8/columnflow/tasks/framework/plotting.py#L80-L87
[2] https://github.com/columnflow/columnflow/blob/5939cb0a1f70e909f2df6c46bbed593b873a19e8/columnflow/tasks/framework/plotting.py#L322-L329
[3] https://github.com/columnflow/columnflow/blob/5939cb0a1f70e909f2df6c46bbed593b873a19e8/columnflow/tasks/framework/base.py#L419

riga:
I think we could follow to paths here:

1. If there is a single value that could be interpreted as a group, check if it's a group and load its contents if it is indeed a group.
2. We use a "reference-like" syntax to mark a value explicitly as a group. I'm thinking of something similar as the config referencing in law or yaml, i.e. "my_value: &other_value". This would look like `--process-settings "&my_group:other_process,some_setting=True"`.

Thinking about the implications of both paths, I'd definitely vote for 2 as it's far more explicit. However, it does require values to be quoted on the command line, but I think that's a fair trade-off.

Including @dsavoiu here as well.

mafrahm:
I feel like there could be three things to resolve:

1. resolving full groups of settings, where each group is a dict of processes and their corresponding settings (e.g. ```--process-settings proc1,some_setting=True:some_setting_group:another_setting_group```). This should be the only case where we only pass a single value before adding the next `:`, since passing a process without settings is useless
2. resolving groups of processes (e.g. ```--process-settings top,some_setting=True``` should be resolved to ```--process-settings st,some_setting=True:tt,some_setting=True```)
[...]

PR #461, opened 2024-06-13
Improve process settings handling.

riga:
This PR slightly changes how process settings are applied during plotting.

Currently, the process name used in the process settings parameter for plotting, e.g. `--process-settings "hh_ggf_hbb_htt_kl1_kt1,unstack"`, requires that the process name matches exactly the name of the process that is plotted.

However, it would be nice if we could rely on the processes' parent-child relations to assign these settings, especially not to be forced to always remember the sometimes long-ish process names. In the example above, if one knows that there's no other hh processes contributing to the plot, a simple `--process-settings "hh,unstack"` should do the trick.

With the changes in this PR, this is now possible.

### DIFF
--- a/columnflow/plotting/plot_util.py
+++ b/columnflow/plotting/plot_util.py
@@ -57,7 +57,7 @@ def get_cms_label(ax: plt.Axes, llabel: str) -> dict:
 
 def apply_settings(
     instances: Iterable[od.AuxDataMixin],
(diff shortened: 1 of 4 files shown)


mafrahm (approved):
Looks good to me!

Also mentioning that in theory you could also use groups of processes for that, but there is also still an open issue regarding that topic (https://github.com/columnflow/columnflow/issues/308)

PR #468, opened 2024-06-13
allow stitching in normalization_weights Producer

mafrahm:
Changes:

- `process_ids` is no longer automatically called in the `normalization_weights` Producer
- stitching is performed separately for each process that is included in the `process_id` column and the `sum_mc_weights_per_process`
- the process xsecs can be calculated from the inclusive sample (`get_xsecs_from_inclusive_dataset=True`) or via the cmsdb xsecs (`get_xsecs_from_inclusive_dataset=False`)
- I kept `normalization_weights` as the unstitched one for consistency.


riga:
I hope it's ok I switched it to draft mode for now, but of course, remove the label again when the PR can be reviewed :)

columnflow/production/normalization.py
```diff
@@ -76,31 +173,41 @@ def normalization_weights_setup(
...
-    selection_stats = self.task.cached_value(
-        key="selection_stats",
-        func=lambda: inputs["selection_stats"]["collection"][0]["stats"].load(formatter="json"),
-    )
+    normalization_selection_stats = {
+        dataset: self.task.cached_value(
+            key=f"selection_stats_{dataset.name}",
+            func=lambda: inp["collection"][0]["stats"].load(formatter="json"),
+        )
+        for dataset, inp in inputs[self.selection_stats_key].items()
```
mafrahm, line 187: Hey @riga, just to make sure: this will cache the selection stats per dataset, therefore even though we run this setup function multiple times (both for stitched and unstitched normalization), the cached values will still not produce any unexpected behaviour, right?

columnflow/production/normalization.py
```diff
@@ -14,35 +18,112 @@
...
+        raise Exception("inclusive dataset not found")
+
+    unmatched_processes = {p for p in process_map if not incl_dataset.has_process(p, deep=True)}
+    if unmatched_processes:
+        raise Exception(f"processes {unmatched_processes} not found in inclusive dataset")
+
+    return incl_dataset
+
+
+def get_stitching_datasets(config_inst: od.Config, dataset_inst: od.Dataset) -> set[od.Dataset]:
```
riga, line 48: This should return a list for type consistency.

columnflow/production/normalization.py
```diff
@@ -14,35 +18,112 @@
...
-    events = self[process_ids](events, **kwargs)
-
     # get the lumi
     lumi = self.config_inst.x.luminosity.nominal
 
     # read the cross section per process from the lookup table
     process_id = np.asarray(events.process_id)
+
+    unique_process_ids = set(process_id)
+    invalid_ids = unique_process_ids.difference(self.xs_table.rows[0])
```
riga, line 113:
```suggestion
    invalid_ids = unique_process_ids - set(self.xs_table.rows[0])
```

columnflow/production/normalization.py
```diff
@@ -76,31 +173,41 @@ def normalization_weights_setup(
...
-        func=lambda: inputs["selection_stats"]["collection"][0]["stats"].load(formatter="json"),
-    )
+    normalization_selection_stats = {
+        dataset: self.task.cached_value(
+            key=f"selection_stats_{dataset.name}",
+            func=lambda: inp["collection"][0]["stats"].load(formatter="json"),
+        )
+        for dataset, inp in inputs[self.selection_stats_key].items()
+    }
+    # if necessary, merge the selection stats
```
riga, line 189:
```suggestion
    # if necessary, merge the selection stats across datasets
```

columnflow/production/normalization.py
```diff
@@ -76,31 +173,41 @@ def normalization_weights_setup(
...
     max_id = max(process_inst.id for process_inst in process_insts)
 
     # ensure that the selection stats do not contain any process that was not previously registered
-    unregistered_process_ids = [
-        int(process_id) for process_id in selection_stats["sum_mc_weight_per_process"]
-        if int(process_id) > max_id
-    ]
+    allowed_ids = set(int(process_id) for process_id in merged_selection_stats["sum_mc_weight_per_process"])
+
+    unregistered_process_ids = allowed_ids.difference({p.id for p in process_insts})
```
riga, line 210:
```suggestion
    unregistered_process_ids = allowed_ids - {p.id for p in process_insts}
```

columnflow/production/normalization.py
```diff
@@ -53,16 +134,32 @@ def normalization_weights_requires(self: Producer, reqs: dict) -> None:
     Adds the requirements needed by the underlying py:attr:`task` to access selection stats into
     *reqs*.
     """
-    # TODO: for actual sample stitching, we don't need the selection stats for that dataset, but
-    #       rather the one merged for either all datasets, or the "stitching group"
-    #       (i.e. all datasets that might contain any of the sub processes found in a dataset)
+    self.selection_stats_key = f"{'stitched_' if self.allow_stitching else 'norm_'}selection_stats"
```
riga, line 137: I don't quite get why there has to be a dynamic attribute that is used in the requirements dict. Maybe I'm missing something, but would it be possible to drop it?

columnflow/production/normalization.py
```diff
@@ -14,35 +18,112 @@
...
+        raise Exception(
+            "sum of event weights does not match the sum of event weights per process",
+        )
+
+    # compute the branching ratios
+    branching_ratios = {
+        int(process_id): sum_weights / sum_mc_weight
+        for process_id, sum_weights in sum_mc_weight_per_process.items()
+    }
+    # TODO: use keys of BRs to check if all processes are covered
```
riga, line 80: It would be good to have this check in place in this PR.

riga (changes requested):
Very nice! This should be extremely helpful to all analyses.

I just have a couple minor things inline.
Also, could you open a follow-up issue to remind ourselves to add a section in the documentation for this?

columnflow/production/normalization.py
mafrahm: If you use the same key for both the stitched and unstitched reqs function, the requirements from one Producer will overwrite the requirements from the other ones, resulting in getting other requirements than expected for one of the two Producers.

columnflow/production/normalization.py
mafrahm: I don't think you can check this at that point because it is unknown which types of process ids are expected to be used in the end, therefore you can only do this check during the event loop itself where you know which process ids are covered.

columnflow/production/normalization.py
mafrahm: I checked and after a small bugfix, it ran through and the results were the same as before :+1:

columnflow/production/normalization.py
riga: Very true. We should revisit the contextualized requirements/inputs again at some point...

riga (approved):

PR #470, opened 2024-06-14
Add mechanism to hook into histograms (e.g. for ABCD methods)

riga:
This PR introduces a mechanism to hook into the creation (or rather usage) of histograms, with scenarios in mind that should cover ABCD methods.

In addition, the way that produced plot files are named is improved as well, since this was required by the addition of hist hooks (more below).

### Hist hooks

A hist hook can be defined in the new aux entry `hist_hooks` of the config.

```python
def example_hook(
    task: AnalysisTask,
    hists: dict[od.Process, hist.Hist],
) -> dict[od.Process, hist.Hist]:
    # add new objects to hists and return it
    return hists


cfg.x.hist_hooks = {
    "example": example_hook,
}
```

Now, all non-cutflow plot tasks support a parameter `--hist-hook NAME` which can trigger the specified hook right before the plotting function is invoked, and also right before certain axis are projected or selected. In the example of the `cf.PlotVariables1D`, the histograms being passed to the hook still contain the `category` and, optionally, the `shift` axes.

This way, hist hooks can easily be used to inject new histograms from data-driven background estimations based on existing histograms.

A trivial example in which a process "QCD" should be estimated as half of the "tt" contribution would look like:

```python
def example_hook(
    task: AnalysisTask,
    hists: dict[od.Process, hist.Hist],
) -> dict[od.Process, hist.Hist]:
    # create a new "QCD" process, if not already done in the config itself
    import order as od
    qcd = od.Process("qcd", id="+", label="QCD", color1=(244, 93, 66))

    # create the qcd shape
    tt_hist = hists[self.config_inst.processes.n.tt]
    hists[qcd] = tt_hist * 0.5

    return hists
```

### Plot parts

Since the above mechanism introduces yet another optional parameter that should be included in the output name when set, I added a `plot_parts()` interface, similar to what we already do with `store_parts()`. `plot_parts()`, however, only define which parts are written in the file name itself, separated by two underscores as was done manually before.

### DIFF
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -309,7 +309,7 @@ class PlotCutflow(
 
     def plot_parts(self) -> law.util.InsertableDict:
         parts = super().plot_parts()
-        parts["category"] = self.branch_data
+        parts["category"] = f"cat_{self.branch_data}"
         return parts
 
     def output(self):
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -616,8 +616,8 @@ class PlotCutflowVariables1D(
 
     def plot_parts(self) -> law.util.InsertableDict:
         parts = super().plot_parts()
-        parts["category"] = self.branch_data.category
-        parts["variable"] = self.branch_data.variable
+        parts["category"] = f"cat_{self.branch_data.category}"
+        parts["variable"] = f"var_{self.branch_data.variable}"
         return parts
 
     def output(self):
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -711,8 +711,8 @@ class PlotCutflowVariables2D(
     def plot_parts(self) -> law.util.InsertableDict:
         parts = super().plot_parts()
         parts["processes"] = self.processes_repr
-        parts["category"] = self.branch_data.category
-        parts["variable"] = self.branch_data.variable
+        parts["category"] = f"cat_{self.branch_data.category}"
+        parts["variable"] = f"var_{self.branch_data.variable}"
         return parts
 
     def output(self):
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -307,21 +307,35 @@ class PlotCutflow(
             for d in self.datasets
         }
 
+    def plot_parts(self) -> law.util.InsertableDict:
+        parts = super().plot_parts()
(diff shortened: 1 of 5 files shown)


columnflow/tasks/cutflow.py
```diff
@@ -307,21 +307,35 @@ def requires(self):
             for d in self.datasets
         }
 
+    def plot_parts(self) -> law.util.InsertableDict:
+        parts = super().plot_parts()
+        parts["category"] = self.branch_data
```
mafrahm, line 312: is there a reason to remove the "cat_" prefix?
```suggestion
        parts["category"] = f"cat_{self.branch_data}"
```

columnflow/tasks/cutflow.py
```diff
@@ -595,25 +614,32 @@ class PlotCutflowVariables1D(
         "default: processes",
     )
 
+    def plot_parts(self) -> law.util.InsertableDict:
+        parts = super().plot_parts()
+        parts["category"] = self.branch_data.category
+        parts["variable"] = self.branch_data.variable
```
mafrahm, line 620:
```suggestion
        parts["category"] = f"cat_{self.branch_data.category}"
        parts["variable"] = f"var_{self.branch_data.variable}"
```

columnflow/tasks/cutflow.py
```diff
@@ -682,16 +708,20 @@ class PlotCutflowVariables2D(
         add_default_to_description=True,
     )
 
+    def plot_parts(self) -> law.util.InsertableDict:
+        parts = super().plot_parts()
+        parts["processes"] = self.processes_repr
+        parts["category"] = self.branch_data.category
+        parts["variable"] = self.branch_data.variable
```
mafrahm, line 715:
```suggestion
        parts["category"] = f"cat_{self.branch_data.category}"
        parts["variable"] = f"var_{self.branch_data.variable}"
```

mafrahm (commented):
I left some minor inline comments, but generally this looks good to me.

I think we still had some open points to discuss, e.g. what happens when you want to only plot a sub-set of processes but need all background processes to do the background estimation of a certain process.

riga:
Thanks for the suggestions, we should indeed keep the prefixes.

> I think we still had some open points to discuss, e.g. what happens when you want to only plot a sub-set of processes but need all background processes to do the background estimation of a certain process.

Here, one can use the different purposes of `--datasets` and `--processes`, with the first one deciding which *data* is requested, and the second one configuring *what* is actually plotted. If this isn't enough, we could target this in a subsequent PR.

PR #471, opened 2024-06-14
Make bin shifting optional.

riga:
This PR adds a parameter to the normal and cutflow plotting tasks that controls the behavior of the histogram filling regarding the treatment of entries whose values are exactly at the bin edge of the right most bin.

With this PR, the behavior defaults to an automatic evaluation (based on the axis traits), but can be actively enabled or disabled.


columnflow/tasks/framework/parameters.py
```diff
@@ -22,6 +22,15 @@
...
 
+shift_last_bin_inst = law.OptionalBoolParameter(
+    default=None,
+    significant=False,
+    description="whether to shift entries that have the exact value of the right-most bin edge "
+    "slightly to the left such that they end up in the last instead of the overflow bin; when "
+    "'None', shifting is performed for all variable axis that are continuous and non-circular; "
+    "default: None",
+)
+
```
dsavoiu, line 33: Not a huge concern, but it seems that the only way for users to actively select the automatic decision logic here based on axis traits (i.e. passing `shift_lаst_bin=None` to `fill_hist`) is to omit the `--shift-last-bin` parameter, right? In general, I think it would be good to avoid having three semantically distinct states for an `OptionalBoolParameter` and would instead advocate to always map the default `None` to either `True` and `False` in a transparent way.

A good alternative could be to use a `ChoiceParameter` with three choices like `{"false", "true", "аuto"}` and have it resolved to a boolean via `resolve_param_values` or in the `fill_hist` function itself.

dsavoiu (commented):
Hi @riga,

thanks a lot for implementing this! I think it's very useful to have this as a task parameter for use cases where the bin shift is not desired.

The changes look good to me overall, I just have a small remark below that we should maybe discuss before merging.

In addition, I would maybe also think about introducing a flag for this in the _analysis config_ and setting the default of the `shift_last_bin` parameter based on that. In that way analyses could have an entry in `law.cfg` to control the behavior globally without needing to add it on the command-line. Is that something you would consider adding to this PR?

columnflow/tasks/framework/parameters.py
riga: > Not a huge concern, but it seems that the only way for users to actively select the automatic decision logic here based on axis traits (i.e. passing shift_lаst_bin=None to fill_hist) is to omit the --shift-last-bin parameter, right?

It would also be possible to pass None explicitly, i.e. `--shift-last-bin None`, so in this sense, the `OptionalBoolParameter` is *like* a choice parameter, but still with boolean parsing enabled if the value is not `"None"`.

I think we use that in other places, too, for the exact same reason of a boolean switch with a context-specific default (@mafrahm I think these were plotting related options, right?).

riga:
Thanks for the review @dsavoiu !

> In addition, I would maybe also think about introducing a flag for this in the analysis config and setting the default of the shift_last_bin parameter based on that. In that way analyses could have an entry in law.cfg to control the behavior globally without needing to add it on the command-line. Is that something you would consider adding to this PR?

Yes, that sounds like a good idea. To be consistent with the other "elsewhere-defined" defaults, it could also make sense to keep the default in the `[analysis]` section of the config (like default analysis, config, producer, etc ,...). Would that also work for you?

columnflow/tasks/framework/parameters.py
dsavoiu: I see, thanks for the explanation @riga! I was not aware that the `OptionalBoolParameters` supported `None` as a value. Then all is fine :)

dsavoiu:
Hi @riga, yes, having it in the `analysis` section to be more consistent with the other defaults sounds good to me. Now that I think about it, we should then also probably choose a more descriptive name for this parameter/flag, since it may not be immediately clear that it refers to the histogram filling behavior just from the name. I would suggest something along the lines of `histogram_last_bin_inclusive`, or similar. Could even be `histogram_shift_last_bin`, although I'm wondering if the `shift` part is maybe not confusing, since we also use `shift` to mean something completely different.

In any case, the parameter and config key should probably be called the same, modulo a possible `default_` prefix for the latter.

Any thoughts?

riga:
Makes sense to me. How about

- `fill_hist` argument: `include_last_edge`
- Task parameter: `--include-last-edge` (I would drop the `histogram-` prefix since the parameter is only present for histogramming tasks anyway)
- Default in law.cfg: `analysis::default_histogram_include_last_edge`

?

columnflow/columnar_util.py
```diff
@@ -1156,24 +1156,33 @@ def fill_hist(
...
+    The data is assumed to be structured in the same way as the histogram axes. If
+    *last_edge_inclusive* is *True*, values that would land exactly on the upper-most bin edge of an
+    axis are shifted into the last bin. If it is *None*, the behavior is determined automatically
+    and depends on the variable axis type. In this case, shifting is applied to all continuous,
+    non-circular axes.
     """
     if fill_kwargs is None:
         fill_kwargs = {}
 
+    # helper do decide whether the variable axis qualifies for shifting the last bin
```
dsavoiu, line 1173:
```suggestion
    # helper to decide whether the variable axis qualifies for shifting the last bin
```

columnflow/tasks/framework/parameters.py
```diff
@@ -22,6 +22,20 @@
...
+    "analysis",
+    "default_histogram_last_edge_inclusive",
+    None,
+)
+last_edge_inclusive_inst = law.OptionalBoolParameter(
+    default=_default_last_bin_edge_inclusive,
+    significant=False,
+    description="whether to shift entries that have the exact value of the right-most bin edge "
+    "slightly to the left such that they end up in the last instead of the overflow bin; when "
+    "'None', shifting is performed for all variable axis that are continuous and non-circular; "
```
dsavoiu, line 35:
```suggestion
    "'None', shifting is performed for all variable axes that are continuous and non-circular; "
```

dsavoiu (approved):
Thanks a lot @riga for implementing the comments! I just noticed some typos, but otherwise I would say this looks good to merge.

riga:
Thanks for catching the typos!

PR #474, opened 2024-06-17
.copy not defined for awkward arrays. Use ak.copy instead

maadcoen:


riga (approved):
Thanks @maadcoen ! Let's merge.

PR #475, opened 2024-06-17
BUG: `tenacity` version 8.4.0 doesn't work

haddadanas:
`tenacity` current version 8.4.0 doesn't work with `columnflow` (or rather the current `luigi` version). Rerunning the github linting actions will directly lead to the error.
This PR set tenacity to the previous release until Spotify fixes the problem (Issue for Spotify already exists [luigi#3292](https://github.com/spotify/luigi/issues/3292))

The Error
```
 columnflow/__init__.py:11: in <module>
    import law
modules/law/law/__init__.py:35: in <module>
    import luigi
data/software/venvs/venv_columnar_dev_3cbb5aff/lib/python3.9/site-packages/luigi/__init__.py:34: in <module>
    from luigi import rpc
data/software/venvs/venv_columnar_dev_3cbb5aff/lib/python3.9/site-packages/luigi/rpc.py:33: in <module>
    from tenacity import Retrying, wait_fixed, stop_after_attempt
data/software/venvs/venv_columnar_dev_3cbb5aff/lib/python3.9/site-packages/tenacity/__init__.py:653: in <module>
    from tenacity.asyncio import AsyncRetrying  # noqa:E402,I100
E   ModuleNotFoundError: No module named 'tenacity.asyncio'
```

### DIFF
--- a/sandboxes/cf.txt
+++ b/sandboxes/cf.txt
@@ -1,5 +1,6 @@
 # version 11
 
+tenacity==8.3.0
 luigi~=3.5.1
 scinum~=2.1.1
 six~=1.16.0
--- a/sandboxes/cf.txt
+++ b/sandboxes/cf.txt
@@ -1,6 +1,6 @@
 # version 11
 
-tenacity==8.3.0
+tenacity!=8.4.0
 luigi~=3.5.1
 scinum~=2.1.1
 six~=1.16.0


haddadanas:
The bug has been fixed from `tenacity` side (with version 8.4.1). But maybe it would make sense to fix the version. Else this PR can be closed :smile:

sandboxes/cf.txt
```diff
@@ -1,5 +1,6 @@
 # version 11
 
+tenacity==8.3.0
```
riga, line 3: Instead of pinning a specific version, I would just exclude the version that is known to fail.

```suggestion
tenacity!=8.4.0
```

riga (changes requested):
Good catch! I made an inline suggestion though to exclude the failing version.

riga (approved):
:+1:

PR #476, opened 2024-06-18
Linting the docs :)

haddadanas:
This PR fixes all linting issues in the cf-docs and `README` files. There are fixes for the markdown files (from the `pymarkdown` linting) as well as for all warning and errors occurring in `sphinx` when building the docs.

Some rules have been disabled in the config file, since they are widely used (e.g. html tables in the README etc.). Should this not be wanted, I am ready to make further changes :smile: 

The PR also add missing code tests to `run_coverage.sh`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,7 @@
-<!-- marker-before-logo -->
-
-<p align="center">
+<h1 align="center">
   <img src="https://media.githubusercontent.com/media/columnflow/columnflow/master/assets/logo_dark.png#gh-light-mode-only" width="480" />
   <img src="https://media.githubusercontent.com/media/columnflow/columnflow/master/assets/logo_bright.png#gh-dark-mode-only" width="480" />
-</p>
-
-<!-- marker-after-logo -->
+</h1>
 
 <!-- marker-before-badges -->
 
@@ -68,7 +64,7 @@ bash -c "$(curl -Ls https://raw.githubusercontent.com/columnflow/columnflow/mast
 
 At the end of the setup, you will see further instructions and suggestions to run your first analysis tasks (example below).
 
-```
+```text
 Setup successfull! The next steps are:
 
 1. Setup the repository and install the environment.
@@ -107,7 +103,6 @@ Setup successfull! The next steps are:
 
 For a better overview of the tasks that are triggered by the commands below, checkout the current (yet stylized) [task graph](https://github.com/columnflow/columnflow/wiki#default-task-graph).
 
-
 ## Projects using columnflow
 
 - [hh2bbtautau](https://github.com/uhh-cms/hh2bbtautau): HH → bb𝜏𝜏 analysis with CMS.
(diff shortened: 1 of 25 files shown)


haddadanas:
For the test to not fail (there are some files in law, which are scanned and are not linted) we have to either:

- apply the linting for all submodules (PR are there :wink: ) or
- switch the `pymarkdown` command to scan only cf-files (in  [`tests/run_docs`](https://github.com/haddadanas/columnflow/blob/c88fbee791f6b7e0a3e07dd3ca0d143c2578f87f/tests/run_docs#L23)):
```
    pymarkdown --config "${cf_dir}/.markdownlint" scan "${cf_dir}/README.md" "${cf_dir}/docs/*.md" \
        "${cf_dir}/docs/*/*.md" "${cf_dir}/docs/*/*/*.md" "${cf_dir}/docs/*/*/*/*.md"
```

README.md
```diff
@@ -1,11 +1,7 @@
-<!-- marker-before-logo -->
```
riga, line 1: Could you leave the markers in there? They are important for the sphinx docs.

docs/task_overview/introduction.md
```diff
@@ -36,50 +36,50 @@ the name of the dataset to be searched for, as defined in the analysis config is
...
 
 - {py:class}`~columnflow.tasks.calibration.CalibrateEvents`: Task to implement corrections to be
-applied on the datasets, e.g. jet-energy corrections. This task uses objects of the
-{py:class}`~columnflow.calibration.Calibrator` class to apply the calibration. The argument
-```--calibrator``` followed by the name of the Calibrator
-object to be run is needed for this task to run. A default value for this argument can be set in
-the analysis config. Similarly, the ```--shift``` argument can be given, in order to choose which
-corrections are to be used, e.g. which variation (up, down, nominal) of the jet-energy corrections
-are to be used.
+    applied on the datasets, e.g. jet-energy corrections. This task uses objects of the
```
riga, line 39: Could you change the breaking here so that each sentence is on one line? I think for markdown files, editors should be set up in a way similar to writing large tex files, i.e., one line per sentence and the editor wraps lines automatically at some width (but only visually).

riga (changes requested):
Nice work @haddadanas !

I made two comments inline, one of which I want to repeat here since this might be a more general discussion.

1. In markdown files (and just as for other formats that have plain text like tex files) we should stick to the "one sentence per line" pattern. The only other alternative is breaking lines at certain widths, but this ultimately causes overcomplicated diffs. For instance, when a single word as just in some line, which might cause the next N lines to be realigned and rewrapped, resulting in N line changes, whereas it should be only one. Instead, when using "one line per sentence", editors should be set up such that they do visual line breaks instead. This also plays nicely with enumerations, lists, etc, through some indentation in the next line.

2. Could you exclude the linting from looking into submodules? I saw you opened PRs in law/order, but there I would probably go for different linting rules, since the docs use a mix or rst/md.

docs/task_overview/introduction.md
haddadanas: I'll do this for all the documentation :+1: 
One question: Should the one sentence per line be done regardless of the resulting width or should we also break after a width of 120?

README.md
riga: I just saw that don't make use of this header anyways in docs/index.rst, so all good :+1:

riga (approved):
Thanks for handling this! Looks like a lot of work 🫡

PR #480, opened 2024-06-28
Small fix for the `GetDatasetLFNs` task

haddadanas:


riga (approved):
:+1:

This should make it easier to use the task from within tasks that are not `DatasetTask`'s themselves!

PR #481, opened 2024-06-30
allow building histograms in SelectEvents

mafrahm:


riga (changes requested):
I like the idea, and the code is quite concise (good that `merge_counts` uses `+=` that hists implement ;) ), but I have one request.

Adding this feature *as is* would essentially invalidate previous selections as a new output is expected that will be missing. Could you add a switch that enables the creation (or rather, the declaration of the hists output) of the hists, have it configurable through the law.cfg and disable it by default?

For the law.cfg, we used this recently here as well:

- definition: https://github.com/columnflow/columnflow/pull/471/files#diff-861327112d2f2c52479fbcddd5e2cea944e41e5860ed4cfdb8bc88bc4580e358R42
- usage: https://github.com/columnflow/columnflow/pull/471/files#diff-6d5b3624097f05d89980ed25b4d713dcb22d2ac2fa341da2aa47844202d2be02R25

Also, what would be the long-term plan? Do we fully rely on hists for storing counts / sum of event weights?

mafrahm:
Hi @riga, keeping the existing outputs validated is a good point! I'd just flag the `hists` outputs from SelectEvents and MergeSelectEvents as optional (and just return an empty dict if the input is missing in MergeSelectEvents). In that way, selections will from now on always produce the (possibly empty) hists output such that we can remove this flag again in some future version.

mafrahm:
For future plans, I'm not 100% sure. I already implemented the renormalized btag weights in my analysis and this works quite well, seems to be faster as before and is much easier to maintain.

For other normalizations (pu, pdf, scale, and the normalization weights), we could keep using the json since these normalizations do not require that many different bins (basically only the process ids) and therefore it should not be a big difference whether we use one or the other.

I'd in any case keep the option to store jsons, since having some stats in a human readable format is quite nice for quick and simple checks (e.g. storing number of events after each step is nice for synchronization exercises)

riga:
> For other normalizations (pu, pdf, scale, and the normalization weights), we could keep using the json since these normalizations do not require that many different bins (basically only the process ids) and therefore it should not be a big difference whether we use one or the other.

> I'd in any case keep the option to store jsons, since having some stats in a human readable format is quite nice for quick and simple checks (e.g. storing number of events after each step is nice for synchronization exercises)

Yeah, I had the same impression. I think the path depends a bit on how we might use these features, so let's defer this to a future discussion.

riga:
I just quickly renamed the option in the law.cfg (it was named differently in the tasks) and also added it to the minimal example.

PR #482, opened 2024-07-01
Update sandboxes, adjust dak/ak IO.

### LINKED ISSUE #469: Update cf and columnar sandboxes
Before the 0.2.3 release, we should update our main sandboxes and potentially adjust internals.

riga:
This PR updates all central sandboxes to use latest dependencies.

Most packages are referenced through the "compatible release" modifier `~=X.Y.Z` (meaning `>= X.Y.Z,==X.Y.*`, so only allowing patch releases), whereas columnar packages (uproot/awkward/dask_awkward/coffea) are pinned to exact versions since they are still somewhat moving targets.

I tested the entire workflow with these updates which, as expected, failed due to some core changes in said columnar packages. Therefore I made some changes deep in the IO, as well as some more high level adjustments, so that everything is running smoothly now. Maybe you can confirm @mafrahm, but I think the memory consumption during event handling is improved.

Closes #469.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -2560,11 +2560,9 @@ class DaskArrayReader(object):
         self.close()
 
     def __len__(self: DaskArrayReader) -> int:
-        return (
-            len(self.dak_array)
-            if self.dak_array.known_divisions
-            else dak.num(self.dak_array, axis=0).compute()
-        )
+        if not self.dak_array.known_divisions:
+            self.dak_array.eager_compute_divisions()
+        return len(self.dak_array)
 
     @property
     def closed(self: DaskArrayReader) -> bool:
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -2529,9 +2529,7 @@ class DaskArrayReader(object):
         self: DaskArrayReader,
         path: str,
         open_options: dict | None = None,
-        # TODO: we want to go back to SLICES as soon as possible, but currently there is an issue
-        #       that we load all events into memory for each chunk when using SLICES
-        materialization_strategy: MaterializationStrategy = MaterializationStrategy.PARTITIONS,
+        materialization_strategy: MaterializationStrategy = MaterializationStrategy.SLICES,
     ):
         super().__init__()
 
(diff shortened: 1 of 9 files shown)


columnflow/columnar_util.py
```diff
@@ -2529,9 +2529,7 @@ def __init__(
         self: DaskArrayReader,
         path: str,
         open_options: dict | None = None,
-        # TODO: we want to go back to SLICES as soon as possible, but currently there is an issue
-        #       that we load all events into memory for each chunk when using SLICES
-        materialization_strategy: MaterializationStrategy = MaterializationStrategy.PARTITIONS,
+        materialization_strategy: MaterializationStrategy = MaterializationStrategy.SLICES,
```
riga, line 2532: `dak_array[start:end]` seems to be working now without memory leak, so back to slicing.

columnflow/columnar_util.py
```diff
@@ -3045,38 +3049,39 @@ def open_coffea_root(
         else:
             raise Exception(f"'{source}' cannot be opened as coffea_root")
 
-        return (source, tree.num_entries)
+        return ((source, tree_name), tree.num_entries)
```
riga, line 3052: It is now required to pass the tree name in the coffea events factory, so the name that is identified here, should be returned and reused by the read method below.

columnflow/selection/__init__.py
```diff
@@ -275,13 +275,17 @@ def __iadd__(self: SelectionResult, other: SelectionResult | None) -> SelectionR
...
         # helper to create a view without behavior
         def deepcopy_without_behavior(struct: T) -> T:
-            return copy.deepcopy(law.util.map_struct(
-                (lambda obj: ak.Array(obj, behavior={}) if isinstance(obj, ak.Array) else obj),
+            return law.util.map_struct(
+                (lambda obj: (
+                    ak.Array(obj, behavior={})
+                    if isinstance(obj, ak.Array)
+                    else copy.deepcopy(obj)
+                )),
```
riga, line 283: The `deepcopy` was in the wrong spot before. Also, ak arrays do no longer support deepcopying.

mafrahm (approved):
This looks like there are only internal changes, are there also any changes to be expected on the user side?

If not, I think we can merge this.

riga:
Indeed, all changes are internal and tested. Nothing should change for users.

PR #485, opened 2024-07-03
Make el9 the default on NAF.

riga:
This PR makes Alma9/el9 the default on the NAF htcondor batch system.

The changes are pretty straight forward.

**Switching to el9**

At this time, naf-cms* login nodes 15 and higher are already on el9, so to switch, naf users would have to login only to these nodes and reinstall their entire software stack once (doing `CF_REINSTALL_SOFTWARE=1 source setup.sh ...` once).

**Staying at centos7**

To stay on the old OS, one could export `CF_HTCONDOR_FLAVOR=naf_el7` or set `--htcondor-flavor naf_el7` per task. **However**, I think this would only be a temporary solution since also login nodes ≤ 14 will be moved in the very near future so I guess we should all transition asap anyway.


haddadanas (approved):

mafrahm (approved):
LGTM

riga:
Ok great. Then let's make this the default now.

PR #486, opened 2024-07-03
Add btag sf config, allowing for additional corrector args.

riga:
This PR generalizes the btag weight producer in two regards:

- A `BTagSFConfig` (dataclass) object is introduced to have a more verbose way of configuring things for the `btag_weight` producer. The latter's behavior is fully backwards compatible and also accepts tuples as before.
- This config object accepts `corrector_kwargs` which can be a dictionary which is passed to the correction set evaluation. This allows for configuring additional values potentially required by different correction sets (e.g. wp based SFs) to be passed to the evaluation (e.g. the name of the wp).

### DIFF
--- a/columnflow/production/cms/btag.py
+++ b/columnflow/production/cms/btag.py
@@ -193,7 +193,7 @@ def btag_weights(
             flavor_mask = flavor != 4
 
         # prepare arguments
-        args = {
+        variable_map = {
             "systematic": syst_name if syst_name == "central" else f"{syst_direction}_{syst_name}",
             "flavor": flavor[flavor_mask],
             "abseta": abs_eta[flavor_mask],
@@ -203,7 +203,10 @@ def btag_weights(
         }
 
         # get the flat scale factors
-        sf_flat = self.btag_sf_corrector(*(args[inp.name] for inp in self.btag_sf_corrector.inputs))
+        sf_flat = self.btag_sf_corrector(*(
+            variable_map[inp.name]
+            for inp in self.btag_sf_corrector.inputs
+        ))
 
         # insert them into an array of ones whose length corresponds to the total number of jets
         sf_flat_all = np.ones(n_jets_all, dtype=np.float32)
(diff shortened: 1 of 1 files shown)


columnflow/production/cms/btag.py
```diff
@@ -147,14 +192,18 @@ def add_weight(syst_name, syst_direction, column_name):
             # apply to all but c flavor
             flavor_mask = flavor != 4
 
+        # prepare arguments
+        args = {
```
mafrahm, line 196: this dict is currently named `variable_map` in most of our correctionlib-based Producers. Would be nice to stay consistent here, but not necessary

columnflow/production/cms/btag.py
```diff
@@ -6,18 +6,60 @@
...
+                f"'{self.correction_set}' is highly discouraged",
+            )
+
+    @classmethod
+    def new(
+        cls,
+        obj: BTagSFConfig | tuple[str, list[str]] | tuple[str, list[str], str],
+    ) -> BTagSFConfig:
+        # purely for backwards compatibility with the old tuple format
+        return obj if isinstance(obj, cls) else cls(*obj)
```
mafrahm, line 60: Could we also allow passing a simple dictionary?

mafrahm (approved):
Two minor inline comments. Looks good to me otherwise

columnflow/production/cms/btag.py
riga: In principle yes, but thinking about it, I would prefer these dataclass-style container objects since they allow for some build-in parsing and validation. In the long-term I would even remove the backwards compatibility and no longer accept tuples.

Did you have a special use case in mind that would require dicts?

columnflow/production/cms/btag.py
riga: True. I renamed it to be consistent.

columnflow/production/cms/btag.py
mafrahm: Not really, just laziness such that a user does not need to import the `BTagSFConfig` for their own analysis as the checks from the dataclass object runs anyways when the `get_btag_config` function is called in the init.

PR #488, opened 2024-07-04
Route brace expansion

riga:
This PR provides brace expansion to all places where we handle multiple user defined routes / columns. As a result, it becomes possible to write e.g.

```python
@producer(
   uses={"Jet.{pt,eta,phi}"},
)
def ...
```

in CSPs, or

```python
cfg.x.keep_columns = DotDict.wrap({
    "cf.ReduceEvents": {
        "{Jet,FatJet}.{pt,eta,phi,mass,btagPNet*}",
```

in the config. On our end, this would make longish configurations way easier to understand and maintain.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -366,6 +366,15 @@ class Route(od.TagMixin):
             return self.column == other
         return False
 
+    def __lt__(self, other: Route | str | Sequence[str | int | slice | type(Ellipsis) | list]) -> bool:
+        if isinstance(other, Route):
+            return self.fields < other.fields
+        if isinstance(other, (list, tuple)):
+            return self.fields < tuple(other)
+        if isinstance(other, str):
+            return self.column < other
+        return False
+
     def __bool__(self) -> bool:
         return len(self._fields) > 0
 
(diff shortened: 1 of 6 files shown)


columnflow/tasks/reduction.py
```diff
@@ -113,7 +113,15 @@ def run(self):
...
         skip_columns: set[str] = set()
         for c in self.config_inst.x.keep_columns.get(self.task_family, ["*"]):
-            for r in (self.find_keep_columns(c) if isinstance(c, ColumnCollection) else {Route(c)}):
+            # expand collections and braces in strings
+            if isinstance(c, ColumnCollection):
+                routes = self.find_keep_columns(c)
+            elif isinstance(c, str):
+                routes = set(map(Route, law.util.brace_expand(c)))
+            else:
+                routes = {Route(c)}
```
mafrahm, line 122: This code block of determining the routes seems to repeat quite often. Does it make sense to refactor this?

columnflow/columnar_util.py
```diff
@@ -366,6 +366,15 @@ def __eq__(self, other: Route | str | Sequence[str | int | slice | type(Ellipsis
             return self.column == other
         return False
 
+    def __lt__(self, other: Route | str | Sequence[str | int | slice | type(Ellipsis) | list]) -> bool:
```
mafrahm, line 369: is this used anywhere?

mafrahm (approved):
LGTM, just two minor inline comments

columnflow/columnar_util.py
riga: It's needed when sorting sequences of Route objects. We already have `__eq__`, so `__lt__` was the only missing piece.

columnflow/tasks/reduction.py
riga: Yup, did that in a new method :+1:

PR #489, opened 2024-07-05
FIX `stack` argument for the scaling of processes in the plotting

haddadanas:
Small bug, where the data is added to the area of the stacked processes, when calculating the integral of the stack.
In line 110 `hists.items()` is required, because the `unstack` and `is_data` flags are carried by the process instances and not by the histograms.


riga (approved):
:+1:

PR #490, opened 2024-07-08
Plotting updates

riga:
This PR provides a couple updates for the plotting routines:

- A `round_dynamic` function is added that is used by the determination of the scale factor when a process setting contains `scale=stack`. Before, the value was rounded to the closest "10", but sometimes values are in the order of O(10k) which requires a smaller precision. This is handled now by `round_dynamic`.
- A generic `inject_label` is added which can inject substrings into labels (e.g. process labels) according to some predefined rules (e.g. "replace `__PLACEHOLDER__`" or "inject before parentheses").
- Plots have a `--debug-plot` parameter now. When set, an ipython debugger is invoked after the plot function is called which allows to visualize and optionally change plot styles interactively.

### DIFF
--- a/columnflow/plotting/plot_util.py
+++ b/columnflow/plotting/plot_util.py
@@ -133,7 +133,7 @@ def inject_label(
             elif c == "(":
                 in_parentheses -= 1
             if not in_parentheses:
-                return f"{label[:i]}{inject} {label[i:]}"
+                return f"{label[:i]} {inject} {label[i:]}"
 
     # otherwise, just append
     return f"{label} {inject}"
--- a/columnflow/plotting/plot_util.py
+++ b/columnflow/plotting/plot_util.py
@@ -6,6 +6,7 @@ Some utils for plot functions.
 
 from __future__ import annotations
 
+import re
 import operator
 import functools
 from collections import OrderedDict
@@ -26,6 +27,7 @@ label_options = {
     "wip": "Work in progress",
     "pre": "Preliminary",
     "pw": "Private work",
+    "pwip": "Private work in progress",
     "sim": "Simulation",
     "simwip": "Simulation work in progress",
     "simpre": "Simulation preliminary",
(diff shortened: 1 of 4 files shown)


columnflow/plotting/plot_util.py
```diff
@@ -55,6 +57,97 @@ def get_cms_label(ax: plt.Axes, llabel: str) -> dict:
...
+    elif value <= 500:
+        ref = 20
+    else:
+        ref = 50
+
+    # round and return
+    return sign * int(round(value / ref) * ref * 10**mag_over_1k)
+
+
+def inject_label(
```
mafrahm, line 108: Can you provide some examples or tests for the specific usecases? I do not yet understand for what the `before_parentheses` mode is needed.

riga: In our signal samples, we usually have trailing parentheses to denote model parameters as in "HH_ggf (kl=1, kt=1)", and here one would naturally want to have the scaling factor before parameters as it could otherwise be interpreted as e.g. "parameters times 10". However, I just set the default to False to make this an opt-in behavior.

columnflow/plotting/plot_util.py
```diff
@@ -55,6 +57,97 @@ def get_cms_label(ax: plt.Axes, llabel: str) -> dict:
...
+    """
+    Rounds a *value* at various scales to a subjective, sensible precision. Rounding rules:
+
+        - 0 -> 0
+        - (0, 1) -> round to 0.1
+        - [1, 20]: round to 1
+        - (20, 100]: round to 5
+        - (100, 200]: round to 10
+        - (200, 500]: round to 20
+        - (500, 1000]: round to 50
```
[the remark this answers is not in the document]
riga: I thought about that (for symmetry reasons :) ), but rounding to 2 is never used anywhere. But then I agree, we should maybe remove the rounding to 20 as well.

columnflow/plotting/plot_util.py
```diff
@@ -55,6 +57,86 @@ def get_cms_label(ax: plt.Axes, llabel: str) -> dict:
...
+    if before_parentheses and label.endswith(")"):
+        in_parentheses = 1
+        for i in range(len(label) - 2, -1, -1):
+            c = label[i]
+            if c == ")":
+                in_parentheses += 1
+            elif c == "(":
+                in_parentheses -= 1
+            if not in_parentheses:
+                return f"{label[:i]}{inject} {label[i:]}"
```
mafrahm, line 136: Missing space?
```suggestion
                return f"{label[:i]} {inject} {label[i:]}"
```

columnflow/plotting/plot_util.py
```diff
@@ -112,17 +196,23 @@ def get_stack_integral() -> float:
...
+            scale_factor = round_dynamic(get_stack_integral() / h.sum().value)
         if try_int(scale_factor):
             scale_factor = int(scale_factor)
             hists[proc_inst] = h * scale_factor
-            proc_inst.label += f" x{scale_factor}"
+            proc_inst.label = inject_label(
+                proc_inst.label,
+                rf"$\times${scale_factor}",
+                placeholder="SCALE",
+            )
```
mafrahm, line 212: After switching the default, you should probably pass this argument as well.
```suggestion
            proc_inst.label = inject_label(
                proc_inst.label,
                rf"$\times${scale_factor}",
                placeholder="SCALE",
                before_parentheses=True,
            )
```

columnflow/plotting/plot_util.py
```diff
@@ -55,6 +58,87 @@ def get_cms_label(ax: plt.Axes, llabel: str) -> dict:
...
+    Rounds a *value* at various scales to a subjective, sensible precision. Rounding rules:
+
+        - 0 -> 0
+        - (0, 1) -> round to 1 significant digit
+        - [1, 20]: round to 1
+        - (20, 100]: round to 5
+        - (100, 500]: round to 10
+        - (500, 1000]: round to 50
+        - (1000, 5000]: round to 100
+        - (5000, 10000]: round to 500
```
mafrahm, line 72:
```suggestion
        - (100, 200]: round to 10
        - (200, 1000]: round to 50
        - (1000, 2000]: round to 100
        - (2000, 10000]: round to 500
```

columnflow/plotting/plot_util.py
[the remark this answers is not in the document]
mafrahm: I'd personally prefer to keep the step, but since this type of rounding is only used for plotting parameters and not to round any physical quantities, I'm fine with it.

mafrahm (commented):
Left some final suggestions. Otherwise I'm fine with merging

columnflow/plotting/plot_util.py
riga: Mh, I've been thinking about this again, and I'd vote for re-introducing the 20/200 steps since maximum possible divergence is just too large. Say, at a value of 225, the rounding to 200 or 250 would cause a change of roughly 12.5%, which is quite a lot.

riga: After talking to @dsavoiu we came up with a simpler and more accurate way. We could just round to 1 significant digits below 10, and to two significant digits above:

- 0 -> 0 (int)
- (0, 1) -> round to 1 significant digit (float)
- [1, 10) -> round to 1 significant digit (int)
- [10, inf): round to 2 significant digits (int)

mafrahm: That's much easier to comprehend, I like it. The only cases one could still discuss are the steps from e.g. 1 to 2 or 0.1 to 0.2, since there we can still get a rounding of 50%. But it's probably fine as well since it is now clear that we round to 1 significant digit.

riga: Indeed. And these additional cases are maybe quite use-case specific. Maybe let's revisit them once we stumble upon them.

mafrahm (approved):

PR #492, opened 2024-07-08
Raise exception on missing cross sections in normalization weight producer

riga:
This PR changes the warning / error behavior in the normalization weight producer.

Right now, if the cross section of any process is missing, a warning is printed regardless of whether the process is used or not. In addition, there is another warning shown right before an invalid process id is encountered. If this is the case, an exception is implicitly raised 2 lines below that when the xs lookup table is actually accessed.

This PR changes that behavior. No warnings are shown during the producer setup. Instead, an explicit exception is raised when an invalid id is encountered with the exact reason, invalid ids, and ids known to the lookup table.

### DIFF
--- a/columnflow/production/normalization.py
+++ b/columnflow/production/normalization.py
@@ -239,6 +239,7 @@ def normalization_weights_setup(
             xs_table[0, process_inst.id] = process_inst.get_xsec(self.config_inst.campaign.ecm).nominal
 
     self.xs_table = xs_table
+    self.xs_process_ids = set(self.xs_table.rows[0])
 
 
 @normalization_weights.init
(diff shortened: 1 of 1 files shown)


columnflow/production/normalization.py
```diff
@@ -102,19 +102,20 @@ def normalization_weights(self: Producer, events: ak.Array, **kwargs) -> ak.Arra
...
+    # get the lumi and read the process id column
     lumi = self.config_inst.x.luminosity.nominal
-
-    # read the cross section per process from the lookup table
     process_id = np.asarray(events.process_id)
 
+    # ensure all ids were assigned a cross section
     unique_process_ids = set(process_id)
-    invalid_ids = unique_process_ids - set(self.xs_table.rows[0])
+    invalid_ids = unique_process_ids - self.xs_process_ids
```
mafrahm, line 111: this attribute does not exist

mafrahm (changes requested):
One inline comment, if that one is fixed, we can merge this

columnflow/production/normalization.py
riga: Right, my bad, I forgot to stage this hunk. It's added now.

mafrahm (approved):

PR #493, opened 2024-07-08
Configurable IO settings.

riga:
This PR allows to control the chunk and pool size settings of our IO handler from various points.

So far, the only option to control these two values is the global law.cfg, e.g.

```
[analysis]
chunked_io_chunk_size: 50000
chunked_io_pool_size: 2
```

However, it might be beneficial to configure these values on a per-task or per-calibrator/selector/producer (CSP) level. One use case is a producer that calls a machine learning model that requires a lot of memory which - for chunk sizes around 50k - might just be too much to handle by normal batch nodes.

To achieve this, each CSP now has the option to define a `max_chunk_size`, and the tasks identify the minimum of sizes within a CSP "tree". This value is then passed to the ChunkedIOHandler. I left out `pool_size` here for now, as I don't think this might become a useful option to set on CSP level.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -2797,12 +2797,16 @@ class ChunkedIOHandler(object):
         ["chunk", "chunk_pos"],
     )
 
+    default_chunk_size = law.config.get_expanded_int("analysis", "chunked_io_chunk_size", 50000)
+    default_pool_size = law.config.get_expanded_int("analysis", "chunked_io_pool_size", 2)
+
     def __init__(
         self,
         source: Any,
+        *,
         source_type: str | Sequence[str] | None = None,
-        chunk_size: int = law.config.get_expanded_int("analysis", "chunked_io_chunk_size", 50000),
-        pool_size: int = law.config.get_expanded_int("analysis", "chunked_io_pool_size", 4),
+        chunk_size: int = default_chunk_size,
+        pool_size: int = default_pool_size,
         open_options: dict | Sequence[dict] | None = None,
         read_options: dict | Sequence[dict] | None = None,
         read_columns: set | Sequence[set] | None = None,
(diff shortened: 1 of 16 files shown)


columnflow/tasks/calibration.py
```diff
@@ -104,6 +104,10 @@ def run(self):
         reader_targets = self.calibrator_inst.run_setup(reqs["calibrator"], inputs["calibrator"])
         n_ext = len(reader_targets)
 
+        # determine the maximum chunk size
+        chunk_sizes = (dep.max_chunk_size for dep in self.calibrator_inst.walk_deps(include_self=True))
+        chunk_size = min((s for s in chunk_sizes if isinstance(s, int)), default=None)
```
mafrahm, line 109: These two lines are repeated for every task. I feel like this could be a (cached) property of the TaskArrayFunction Class, which could be used directly in Line 140.

mafrahm (approved):
LGTM, just one minor inline comment

columnflow/tasks/calibration.py
riga: Then we also have to make `max_chunk_size` an attribute of `TaskArrayFunction` instead of the inheriting CSPs. But I agree, this is a rather generic feature that should live in the parent class.

riga: Added in last commit. I guess this requires a re-review.

columnflow/columnar_util.py
```diff
@@ -2437,6 +2462,19 @@ def __call__(
 
         return result
 
+    def get_min_chunk_size(self) -> int | None:
```
mafrahm, line 2465: probably irrelevant but could this be a cached property?

```suggestion
    @functools.cached_property
    def get_min_chunk_size(self) -> int | None:
```

mafrahm: ah no I though this is called in a loop, but it should in any case only be called once per task.

PR #494, opened 2024-07-10
allow creating histograms with different axes types

### LINKED ISSUE #437: Allow building histograms with different types of axes
At the moment, each variable that is passed to `cf.CreateHistograms` will automatically create a variable axis. It would be nice if we can also support some other types of Histogram axes (see [Hist](https://hist.readthedocs.io/en/latest/user-guide/axes.html) docu for possible types of axes).
- `Boolean`, `StrCategory`, and `IntCategory` might be used if someone needs a second categorical axis, e.g. for trigger studies.
- `Integer` might be nice for variables with discrete x values, e.g number of jets.

Implementing the creation of histograms with different axes types should be rather straightforward. We just need to add some auxiliary entry to our variable insts that is used to define how the additional axis is created.

mafrahm:
closes #437 

**ToDo:**

- [x] Define function to create customizable hist axes
- [x] Use function in CreateHistograms
- [ ] Use function in CreateCutflowHistograms
- [ ] Implement tests for creating histograms
    - [ ] Variable
    - [ ] Integer
    - [ ] IntCat
    - [ ] StrCat
    - [ ] Boolean
    - [ ] Regular
- [ ] OPTIONAL: helper functions for histogram filling
- [ ] OPTIONAL: plotting functions for different types of axes

### DIFF
--- a/columnflow/hist_util.py
+++ b/columnflow/hist_util.py
@@ -1,0 +1,101 @@
+# coding: utf-8
+
+"""
+Collection of helper functions for creating and handling histograms.
+"""
+
+from __future__ import annotations
+
+import law
+import order as od
+
+from columnflow.util import maybe_import
+
+hist = maybe_import("hist")
+np = maybe_import("numpy")
+ak = maybe_import("awkward")
+
+logger = law.logger.get_logger(__name__)
+
+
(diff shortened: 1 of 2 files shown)


pkausw (changes requested):
Please fix conflicts with master first

pkausw (approved):

PR #495, opened 2024-07-10
Scoped CSP requirements and inputs.

riga:
This PR solves the issue of potentially ambiguous / redundant requirements of task array functions. Calibrators/selectors/producers can define custom requirements which were - until now - added to a single dictionary so that collisions between requirements of different CSPs could occur.

Now, every CSP receives its own dictionary that can be filled with requirements. In order not to require identical tasks multiple times, our main tasks remove redundant tasks from a flattened list of requirements.

### DIFF
--- a/columnflow/tasks/production.py
+++ b/columnflow/tasks/production.py
@@ -50,7 +50,6 @@ class ProduceColumns(
         return reqs
 
     def requires(self):
-        print(f"requirements evaluated, {self!r}")
         return {
             "events": self.reqs.ProvideReducedEvents.req(self),
             "producer": law.util.make_unique(law.util.flatten(self.producer_inst.run_requires())),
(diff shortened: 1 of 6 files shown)


columnflow/tasks/production.py
```diff
@@ -43,14 +45,15 @@ def workflow_requires(self):
         reqs["events"] = self.reqs.ProvideReducedEvents.req(self)
 
         # add producer dependent requirements
-        reqs["producer"] = self.producer_inst.run_requires()
+        reqs["producer"] = law.util.make_unique(law.util.flatten(self.producer_inst.run_requires()))
 
         return reqs
 
     def requires(self):
+        print(f"requirements evaluated, {self!r}")
```
mafrahm, line 53: Remove print statement?
```suggestion
```

mafrahm (approved):
LGTM, I also checked this in my own analysis and running different normalization Producers seems to work :+1:

PR #496, opened 2024-07-11
Fix ProvideReducedEvents

riga:
ProvideReducedEvents is meant to provide reduced events through its output, which is dynamically assigned to be either the output of `ReduceEvents` or `MergeReduceEvents` depending on various conditions. It also dynamically defines its requirements to trigger the right task at the right time in order to have an at least partially known dependency tree ahead of runtime.

However, we observed an issue where datasets were merged that contained only a single file, which is an unnecessary indirection. This PR fixes that. To summarize, the aforementioned conditions are now:

- When merging is explicitly skipped (`--skip-merging`) or the number of input files is known to be 1 while the merging is also not forced (`--force-merging`), require `reduced` events **without the need to request the reduction stats file**.
- Otherwise:
  - The reduction stats file is needed.
  - When merging is forced, require merged events.
  - Otherwise
    - If the file merging (how many input files make up one merged file) is known (i.e., the reduction stats are already present), require either reduced (`file_merging == 1`) or merged events (`file_merging > 1`).
    - If not, do not declare a static dependency since a decision cannot be made yet and defer to the dynamic dependency definition in the run method.

The implementation of these conditions looks cumbersome at first, but since they are rather complex and potentially not easy to maintain, I used a code structure that exactly matches the summary above, while at the same time, requiring and reading the stats file only if really necessary.

### DIFF
--- a/columnflow/tasks/reduction.py
+++ b/columnflow/tasks/reduction.py
@@ -531,7 +531,7 @@ class ProvideReducedEvents(
 
     @law.workflow_property(setter=True, cache=True, empty_value=0)
     def file_merging(self):
-        if self.skip_merging:
+        if self.skip_merging or self.dataset_info_inst.n_files == 1:
             return 1
 
         # check if the merging stats are present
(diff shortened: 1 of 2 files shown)


columnflow/tasks/reduction.py
```diff
@@ -564,54 +564,75 @@ def _req_merged_reduced_events(self, **params) -> law.Task:
...
 
     @workflow_condition.output
     def output(self):
-        if self.skip_merging or self.force_merging:
-            req = self.requires()
+        # determine the definitive requirement
+        if (
+            self.skip_merging or
+            self.dataset_info_inst.n_files == 1 or
+            (not self.force_merging and self.file_merging == 1)
```
mafrahm, line 620: When `self.force_merging` is True, we should require the merged reduced events even if the number of files is 1, right?

```suggestion
            self.skip_merging or
            (not self.force_merging and (self.dataset_info_inst.n_files == 1 or self.file_merging == 1))
```

riga: Good point. And I guess this can be done even easier by just using `self.requires()["events"]` which is definitely known at this point. I did that in ab9c1ac9.

mafrahm: true, this is much better :+1:

mafrahm (approved):

PR #497, opened 2024-07-11
Collect CSP sandbox through dependencies.

riga:
This PR makes use of the new `TaskArrayFunction.walk_deps` method to solve a long-standing issue. So far, in our tasks we overwrote the sandbox parameter with that of the requested CSP. However, for that lookup, we only used the sandbox value by the CSP itself regardless of its dependencies.

I added a new method, `get_sandbox()`, which is traversing dependencies and collection sandbox values. In case no or a single sandbox was found, *None* or the single sandbox are returned. When more than one unique sandbox was found, there is a configurable collision handling that would either issue a warning or raise an exception. More elaborate methods could be included in the future.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -1674,15 +1674,25 @@ class ArrayFunction(Derivable):
         """
         return dep_cls in self.deps
 
-    def walk_deps(self, include_self: bool = False) -> Generator[ArrayFunction, None, None]:
+    def walk_deps(
+        self,
+        depth_first: bool = False,
+        include_self: bool = False,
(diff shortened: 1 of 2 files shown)


mafrahm (approved):
LGTM

not sure how easy it would be, but if multiple sandboxes are requested, could one check if there is a sandbox that includes the required packages from all other sandboxes?

riga:
I was thinking about the exact same thing, but I guess this involves attributing flags to sandboxes like "is_compatible_with_sandbox_XYZ". This requires some discussions and sounds like a future PR :)

PR #499, opened 2024-07-12
add Selector for jet veto maps

mafrahm:
documentation: https://cms-jerc.web.cern.ch/Recommendations/#jet-veto-maps

The procedure still needs to be extended for 2023postBPix, which I'll do as soon as we have impelemented the corresponding campaign in the cmsdb.

### DIFF
--- a/columnflow/selection/cms/jets.py
+++ b/columnflow/selection/cms/jets.py
@@ -6,10 +6,8 @@ Selection modules for jets.
 
 from __future__ import annotations
 
-from collections import defaultdict
-from typing import Tuple
-
 import law
+import math
 
 from columnflow.util import maybe_import, InsertableDict
 from columnflow.columnar_util import set_ak_column, optional_column as optional
@@ -29,14 +27,12 @@ logger = law.logger.get_logger(__name__)
     },
     produces={"Jet.veto_maps"},
     get_veto_map_file=(lambda self, external_files: external_files.jet_veto_map),
-    exposed=True,
 )
 def jet_veto_map(
     self: Selector,
     events: ak.Array,
-    stats: defaultdict,
     **kwargs,
-) -> Tuple[ak.Array, SelectionResult]:
+) -> tuple[ak.Array, SelectionResult]:
     """
     Selector that applies the Jet Veto Map to the jets and stores the result as a new column ``Jet.veto_maps``.
     Additionally, the ``jet_veto_map`` step is added to the SelectionResult that masks events containing
@@ -44,14 +40,13 @@ def jet_veto_map(
     For users that only want to remove the jets from the veto map, the ``veto_map_jets`` object
     is added to the SelectionResult.
 
-
     Requires an external file in the config
     under ``jet_veto_map``:
 
     .. code-block:: python
 
         cfg.x.external_files = DotDict.wrap({
-            "jet_veto_map": ("/afs/cern.ch/user/m/mfrahm/public/mirrors/jsonpog-integration-a332cfa/POG/JME/2022_Summer22EE/jetvetomaps.json.gz", "v1"), # noqa
+            "jet_veto_map": ("/afs/cern.ch/user/m/mfrahm/public/mirrors/jsonpog-integration-a332cfa/POG/JME/2022_Summer22EE/jetvetomaps.json.gz", "v1"),  # noqa
         })
 
     *get_veto_map_file* can be adapted in a subclass in case it is stored differently in the external files.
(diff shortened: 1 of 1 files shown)


columnflow/selection/cms/jets.py
```diff
@@ -0,0 +1,137 @@
...
+    if "postBPix" in self.config_inst.campaign.name:
+        # in postBPix, we need to run the veto map with type=jetvetomap_bpix and subtract this from
+        # the result of the nominal jet veto map
+        raise NotImplementedError("Jet Veto Map for 2023 postBPix not implemented yet")
+
+    # add the veto map result to the events
+    events = set_ak_column(events, "Jet.veto_maps", veto_map_result)
+    results = SelectionResult(
+        steps={"jet_veto_map": ak.sum(veto_map_jet_mask, axis=1) >= 1},
+        objects={"Jet": {"veto_map_jets": veto_map_jet_mask}},
```
riga, line 103: I don't think we should provide objects in "helper" selectors by default, but it should be a user decision. Also, since events with vetoed jets are to be discarded altogether, creating objects by default would be a bit misleading.

columnflow/selection/cms/jets.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+
+    return events, results
+
+
+@jet_veto_map.requires
+def jet_veto_map_requires(self: Selector, reqs: dict) -> None:
+    if "external_files" in reqs:
+        return
```
riga, line 112: I just noticed that we can skip these checks in all `*_requires` methods since #495. We should maybe do this in a second PR or just on master. But could you test this here already?

riga (changes requested):
Nice!

There are some minor inline comments as usual, and I think we need to discuss two items:

- Identification of specific CMS campaigns (e.g. "BPix", "EE", ...) through tags.
- Creating of selection objects in helper selectors.

columnflow/selection/cms/jets.py
riga: Btw, to "return" the actual per-jet mask you could also add it to the aux dict of the SelectionResult.

columnflow/selection/cms/jets.py
```diff
@@ -0,0 +1,137 @@
...
+    },
+    produces={"Jet.veto_maps"},
+    get_veto_map_file=(lambda self, external_files: external_files.jet_veto_map),
+    exposed=True,
+)
+def jet_veto_map(
+    self: Selector,
+    events: ak.Array,
+    stats: defaultdict,
+    **kwargs,
```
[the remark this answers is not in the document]
mafrahm: Yes, that would be possible, but the recommendation is to veto the event as soon as we find one jet which fulfills the mask and is part of the veto map. Passing a mask might lead users to believe that they can keep their event as long as they do not select jets that are part of this map.

I'd therefore prefer to not include such an optional mask.

columnflow/selection/cms/jets.py
```diff
@@ -0,0 +1,137 @@
...
+        "phi": jet.phi,
+    }
+
+    inputs = [variable_map[inp.name] for inp in self.veto_map.inputs]
+
+    # apply the veto map
+    veto_map_result = ak.where(
+        jet_mask,
+        self.veto_map(*inputs),
+        -1,
```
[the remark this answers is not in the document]
mafrahm: This is the default value that is filled when our `jet_mask` is False, just to separate between jets that were not even considered from the veto map and jets that are tagged as jets we do not need to veto (which are assigned a value of 0, while jets that should be vetoed are assigned a value of 100 as far as I have seen)

columnflow/selection/cms/jets.py
mafrahm: Good point, returning an aux is much better

columnflow/selection/cms/jets.py
mafrahm: Yes, looks like it. This function is called 6 times in total (when running SelectEvents for a single branch) and each time the input `reqs` is an empty DotDict.

columnflow/selection/cms/jets.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+
+    # get the Jet veto mask (events containing such a jet should be vetoed)
+    veto_map_jet_mask = (veto_map_result > 0)
+
+    if "bpix" in self.config_inst.campaign.x("postfix", "").lower():
```
riga, line 96: Isn't the "post" bpix era just called "BPix", and the "pre" era misses this value? If so, I'd go for the explicit value check `self.config_inst.campaign.x("postfix", "").lower() == "bpix"`.

riga (changes requested):
One last minor thing inline, then we can merge afterwards.

columnflow/selection/cms/jets.py
mafrahm: done :+1:

riga (approved):

PR #500, opened 2024-07-15
Fix and streamline jet veto maps selection.

riga:
This PR updates the jet veto map selector:

- fixed (flipped) veto map check.
- moved 2023 bpix check to init.
- used `run` and `postfix` aux entries of campaigns.
- merged `veto_map_result` and `veto_map_jet_mask` which were identical, and removed latter from the aux result since it's already stored inside the events chunk.

### DIFF
--- a/columnflow/selection/cms/jets.py
+++ b/columnflow/selection/cms/jets.py
@@ -16,6 +16,7 @@ from columnflow.selection import Selector, SelectionResult, selector
 np = maybe_import("numpy")
 ak = maybe_import("awkward")
 
+
 logger = law.logger.get_logger(__name__)
 
 
@@ -25,7 +26,7 @@ logger = law.logger.get_logger(__name__)
         "Muon.{pt,eta,phi,mass,isPFcand}",
         optional("Jet.puId"),
     },
-    produces={"Jet.veto_maps"},
+    produces={"Jet.veto_map_mask"},
     get_veto_map_file=(lambda self, external_files: external_files.jet_veto_map),
 )
 def jet_veto_map(
@@ -53,7 +54,6 @@ def jet_veto_map(
 
     documentation: https://cms-jerc.web.cern.ch/Recommendations/#jet-veto-maps
     """
-
     jet = events.Jet
     muon = events.Muon[events.Muon.isPFcand]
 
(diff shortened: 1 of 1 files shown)


columnflow/selection/cms/jets.py
```diff
@@ -53,61 +54,64 @@ def jet_veto_map(
...
-    # get the Jet veto mask (events containing such a jet should be vetoed)
-    veto_map_jet_mask = (veto_map_result > 0)
-
-    if self.config_inst.campaign.x("postfix", "").lower() == "bpix":
-        # in postBPix, we need to run the veto map with type=jetvetomap_bpix and subtract this from
-        # the result of the nominal jet veto map
-        raise NotImplementedError("Jet Veto Map for 2023 postBPix not implemented yet")
+    # apply the veto map to selected jets
+    # (a map value of 0 means the jet is to be vetoed)
+    veto_mask = jet_mask & (self.veto_map(*inputs) == 0)
```
mafrahm, line 90: No, this is wrong. An event is to be vetoed when a jet with veto map value > 0 survives the jet preselection.
```suggestion
    # (a map value of >0 means the jet is to be vetoed)
    bad_jet = jet_mask & (self.veto_map(*inputs) > 0)
    good_jet = jet_mask | (self.veto_map(*inputs) == 0)
```

Afterwards, we need to define the selection step accordingly
```
# check that no bad jets are present or that all jets are good jets.
jet_veto_map_step = ak.sum(bad_jet, axis=1) == 0
jet_veto_map_step = ak.sum(good_jet, axis=1) == ak.num(good_jet)
```

mafrahm (changes requested):
One very important inline comment regarding the result. Everything else looks good to me

columnflow/selection/cms/jets.py
riga: I fixed it, please have a look :+1:

columnflow/selection/cms/jets.py
```diff
@@ -53,61 +54,69 @@ def jet_veto_map(
...
+    # evalute the veto map only for selected jets
+    # (a map value of != 0 means the jet is vetoed)
+    veto_mask = jet_mask
+    flat_veto_mask = flat_np_view(veto_mask)
+    flat_veto_mask[flat_veto_mask] = ak.flatten(self.veto_map(*inputs) != 0)
 
-    # get the Jet veto mask (events containing such a jet should be vetoed)
-    veto_map_jet_mask = (veto_map_result > 0)
+    # store the per-jet veto mask
+    events = set_ak_column(events, "Jet.veto_map_mask", veto_mask)
```
mafrahm, line 96: to me it looks like this is still just the `jet_mask`. I think you need to use the `layout_ak_array` function to transform the `flat_veto_mask` back into the correct shape.

riga: The `flat_veto_mask` is really just a view, so values changed in the view directly translate to changed values in the original, structured `veto_mask`.

mafrahm (approved):

PR #501, opened 2024-07-15
Fix selector steps default resolution.

riga:
This PR fixes the default value resolution of the `--selector-steps` parameter.
Before, there was no default, leading to the `steps` field of the selection results. However, the order is arbitrary and usually never reasonable by chance.

The parameter now defaults to config values, and falls back to the "random" steps when nothing is found.

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -230,7 +230,7 @@ class CalibratorsMixin(ConfigTask):
     calibrators = law.CSVParameter(
         default=(RESOLVE_DEFAULT,),
         description="comma-separated names of calibrators to be applied; default: value of the "
-        "'default_calibrator' config in a 1-tuple",
+        "'default_calibrator' config",
         brace_expand=True,
         parse_empty=True,
     )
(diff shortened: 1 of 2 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -881,7 +881,8 @@ class ProducersMixin(ConfigTask):
 
     producers = law.CSVParameter(
         default=(RESOLVE_DEFAULT,),
-        description="comma-separated names of producers to be applied; empty default",
+        description="comma-separated names of producers to be applied; default: value of the "
+        "'default_producer' config in a 1-tuple",
```
mafrahm, line 885: I'd remove this last part because this implies to me that I can only pass a single value via my `default_producer` config. Same for the other parameters.
```suggestion
        description="comma-separated names of producers to be applied; default: value of the "
        "'default_producer' config",
```

mafrahm (changes requested):
LGTM, just one minor inline comments regarding the parameter description

columnflow/tasks/framework/mixins.py
riga: Removed 👍

mafrahm (approved):

PR #503, opened 2024-07-16
changed the way, the BRs of stitched processes are calculated

haddadanas:
Switched from a direct calculation of the Branching ratios between leaf processes and inclusive processes to a step-wise method, where in-between datasets are used to improve on the low statistics one may encounter for low populated processes.


haddadanas:
The errors on the BRs are calculated, however they are not saved within the column, because the sparse matrices used as lookup-table do not support `object` as `dtype`:
```
ValueError: object dtype is not supported by sparse matrices
```

Issue #502, opened 2024-07-15
Align task array functions that use correctionlib


riga:
We have various (mostly cms-specific) task array functions that use correctionlib:

- calibrators: jec, jer, met_phi
- selectors: jet_veto_map
- producers: sfs for btag, electron, muon, weights for pu, scale, pdf variations

They are configured through objects stored directly on `order.Config` objects in a way that is usually documented per function.

However, as we've recently seen for btag weights, these configurations may change in the future, e.g. when generalizing implementations. To make this future proof, we should have lightweight data classes store and validate these settings internally, while providing an optional interface for using dictionaries or even tuples as before that are understood by these containers for backwards compatibility.

riga:
Done by #628.

PR #505, opened 2024-07-17
Add TEC calibrator.

riga:
As the title suggests, this PR adds a tau energy calibrator in the same format as our other cms specific calibrator functions. The `TECConfig` object is still standalone and not inheriting from any central base config. This will be adjusted with #502.

### DIFF
--- a/columnflow/calibration/cms/tau.py
+++ b/columnflow/calibration/cms/tau.py
@@ -1,0 +1,259 @@
+# coding: utf-8
+
+"""
+Tau energy correction methods.
+"""
+
+from __future__ import annotations
+
+import functools
+import itertools
+from dataclasses import dataclass, field
+
+from columnflow.calibration import Calibrator, calibrator
+from columnflow.calibration.util import propagate_met
+from columnflow.util import maybe_import, InsertableDict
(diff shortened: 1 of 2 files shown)


columnflow/calibration/cms/tau.py
```diff
@@ -0,0 +1,259 @@
...
+    if self.with_uncertainties:
+        scales_up = np.ones_like(dm_mask, dtype=np.float32)
+        scales_up[dm_mask] = self.tec_corrector(*args, "up")
+        scales_down = np.ones_like(dm_mask, dtype=np.float32)
+        scales_down[dm_mask] = self.tec_corrector(*args, "down")
+
+    # custom adjustment 1: reset where the matching value is unhandled
+    # custom adjustment 2: reset electrons faking taus where the pt is too small
+    mask1 = (match < 1) | (match > 5)
+    mask2 = ((match == 1) | (match == 3) & (pt <= 20.0))
```
mafrahm, line 142: Not sure about the details, but this currently puts the SF of all taus with match == 1 to 1.0
```suggestion
    mask2 = ((match == 1) | (match == 3)) & (pt <= 20.0)
```

riga: Good catch :+1:

PR #506, opened 2024-07-17
Fix input lfn localization in ReduceEvents

riga:
This PR fixes a critical issue in our `ReduceEvents` task. So far, all input files except the actual nano file were localized using the `@law.decorator.localize` decorator. However, the nano file was localized dynamically and in the scenario of an analysis using remote lfns without wlcg file caching, this resulted in a situation where the chunked loading complained about a missing local file. This bug only appeared in `ReduceEvents` since `{Calibrate,Select}Events` already used an improved method.

This PR adds the same localization method to `ReduceEvents` and streamlines the implementation.

In addition, I slightly aligned the `{Calibrate,Select}Events` tasks to match the new structure.

@jomatthi


mafrahm (approved):
LGTM

PR #507, opened 2024-07-18
allow customizing repr of CSPWs

mafrahm:
This PR allows overwriting the string representation of CSPWs that is stored as part of the output path. 

Example: I'd like to set a `version` in my Producer that I can update separate from the Producer class name to signal my workflow to re-run everything that depends on this particular Producer.

```python
@memoize
def patch_cspw_string_repr():
    def TaskArrayFunction_str(self):
        version_str = f"V{self.version}" if hasattr(self, "version") else ""
        return f"{self.cls_name}{version_str}"

    TaskArrayFunction.__str__ = TaskArrayFunction_str
```

This can of course already be achieved by updating the Producer's `cls_name`, but this might lead to having to update this Producer name at many different code locations as well.

The PR also allows more advanced versioning of CSPWs, e.g. there might be a list of significant parameters that could be turned into a hash.

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -34,7 +34,8 @@ logger = law.logger.get_logger(__name__)
 
 
 class CalibratorMixin(ConfigTask):
-    """Mixin to include a single :py:class:`~columnflow.calibration.Calibrator` into tasks.
+    """
+    Mixin to include a single :py:class:`~columnflow.calibration.Calibrator` into tasks.
 
     Inheriting from this mixin will give access to instantiate and access a
     :py:class:`~columnflow.calibration.Calibrator` instance with name *calibrator*,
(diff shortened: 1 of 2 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -370,6 +375,18 @@ def calibrator_insts(self) -> list[Calibrator]:
             self._calibrator_insts = self.get_calibrator_insts(self.calibrators, {"task": self})
         return self._calibrator_insts
 
+    @property
+    def calibrators_repr(self) -> str:
+        """
+        Return a string representation of the calibrators.
+        """
+        part = "none"
```
riga, line 383: Could you name this differently, e.g. `r` or so? The `part` might be misleading (although it made sense where it previously came from ;) ). Same below.

columnflow/tasks/framework/mixins.py
```diff
@@ -188,6 +188,11 @@ def calibrator_inst(self) -> Calibrator:
 
         return self._calibrator_inst
 
+    @property
+    def calibrator_repr(self):
+        """Return the name of the calibrator."""
```
riga, line 193: This should state

```
"""
Return a string representation of the calibrator.
"""
```

If it was always the name, there would be no need for a customization hook / property.
Same below.

riga (changes requested):
This looks good and useful!

Just two slightly picky inline comments on variable names and doc strings.

riga (approved):

PR #508, opened 2024-07-18
Improve the way BRs are calculated for stitched datasets

haddadanas:


riga (approved):
Looks good!

Merging now since all local tests passed, but @mafrahm you might still want to have an eye on this.

columnflow/production/normalization.py
```diff
@@ -59,24 +63,85 @@ def get_stitching_datasets(self: Producer) -> list[od.Dataset]:
...
+                continue
 
-    if not sum_mc_weight == sum(sum_mc_weight_per_process.values()):
-        raise Exception(
-            "sum of event weights does not match the sum of event weights per process",
-        )
+            proc_ids = [child_proc.id] if is_leaf else [p.id for p in child_proc.get_leaf_processes()]
+            # compute the uncertainty on the branching ratio using number of events
+            _br = N(sum(num_events_per_process.get(str(proc_id), 0) for proc_id in proc_ids)) / N(num_events)
+            # uncertainty is independent of the mc weights, so we can use the same relative uncertainty
```
mafrahm, line 107: I am rather certain that this statement is wrong. We can probably make this assumption because most of the event weights are of similar order of magnitude, but if we had a sample where some weights are significantly larger than others, then these weight will dominate our uncertainty (extreme case: there is one weight much larger than all other weights combined --> in that case, the uncertainty will be completely dominated by that single event)

columnflow/production/normalization.py
```diff
@@ -59,24 +63,85 @@ def get_stitching_datasets(self: Producer) -> list[od.Dataset]:
...
+                (is_leaf and str(child_proc.id) not in sum_mc_weight_per_process) or
+                (not is_leaf and child_proc.id not in proc_ds_map)
+            ):
+                continue
 
-    if not sum_mc_weight == sum(sum_mc_weight_per_process.values()):
-        raise Exception(
-            "sum of event weights does not match the sum of event weights per process",
-        )
+            proc_ids = [child_proc.id] if is_leaf else [p.id for p in child_proc.get_leaf_processes()]
```
mafrahm, line 104: Here it is assumed that the `sum_mc_weight_per_process` dict covers all sub-processes up to the leaf processes. Instead, we should probably take the "leaf processes" of the sub-set of processes that are covered by our stichting datasets.

PR #511, opened 2024-07-22
Fix hist and plot requirements.

riga:
This PR fixes the dependency structure of histogramming and plotting tasks.

- The `MergeHistograms` **workflow** required exactly what its branch 0 required (via `reqs["hists"] = self.as_branch().requires()`), and branches required the `CreateHistograms` workflow, **however**, with it's `--workflow` parameter not set to `local`. This could have lead to situations where a remote job tries to submit new remote jobs. It's fixed now by defining requirements for workflow and branches separately from each other.

- The plot tasks currently require merged histograms with identical settings in both their workflow and branch requirements. This is not changed by this PR as the decision is made on the most abstract plotting base tasks, so further discussions needed. However, to accelerate the initial dependency tree building, this PR drops the workflow requirements on specialized plot tasks.

- `PlotShiftedVariablesPerProcess1D` is meant to be a wrapper tasks, triggering multiple `PlotShiftedVariables1D` with different `--processes` parameters. However, it is currently also a `PlotShiftedVariables1D` task which makes it a workflow with all bells and whistles (proxy object, custom output and completeness definitions, it's own run method, etc.). Although this provides the convenience of having parameters defined directly on wrapper task level, mixing wrapper and workflow features can lead to weird scenarios. It's fixed for now by this PR, however, also dropping convenient parameter access.

### DIFF
--- a/columnflow/tasks/plotting.py
+++ b/columnflow/tasks/plotting.py
@@ -214,7 +214,8 @@ class PlotVariablesBaseSingleShift(
         reqs = super().workflow_requires()
 
         # no need to require merged histograms since each branch already requires them as a workflow
-        reqs.pop("merged_hists", None)
+        if self.workflow == "local":
+            reqs.pop("merged_hists", None)
 
         return reqs
 
@@ -322,7 +323,8 @@ class PlotVariablesBaseMultiShifts(
         reqs = super().workflow_requires()
 
         # no need to require merged histograms since each branch already requires them as a workflow
-        reqs.pop("merged_hists", None)
+        if self.workflow == "local":
+            reqs.pop("merged_hists", None)
 
         return reqs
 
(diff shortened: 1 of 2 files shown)


columnflow/tasks/plotting.py
```diff
@@ -210,6 +210,14 @@ def create_branch_map(self):
             for cat_name in sorted(self.categories)
         ]
 
+    def workflow_requires(self):
+        reqs = super().workflow_requires()
+
+        # no need to require merged histograms since each branch already requires them as a workflow
+        reqs.pop("merged_hists", None)
```
mafrahm, line 217: are you sure we can do this? I'd assume this means that when running plotting tasks via job submission, job dependencies will only be resolved after having already submitted the corresponding branch (meaning that if requirements are not already met, the whole workflow will run in each submitted job)

riga: We can also change this to only have an effect for local workflows.

PR #513, opened 2024-07-22
Accept multiple hist hooks.

riga:
This PR is very straight forward.

Before, we accepted only a single `--hist-hook NAME` to add/remove/change histograms right before plotting or datacard creation. With this PR, multiple comma-separated `--hist-hooks NAME,NAME,...` can be used which receive the histograms in the exact order they are defined on the command line.


mafrahm (approved):
LGTM

PR #516, opened 2024-07-23
Support for wlcg targets with local mirror.

riga:
So far we offer support for local targets and wlcg targets on remote resources. However, many clusters are set up in a way that a local, read-only mount exists for remote targets. 

For this purpose, there is a new "wrapper" target now called `MirroredTarget`'s. All read requests are forwarded to local representations if they exist. If not, the remote *original* resource is queried.

This PR changes the target definition (in `AnalysisTask.target()`) to understand this new `wlcg_mirrored` target type used in the law.cfg file.


mafrahm (approved):
LGTM

PR #518, opened 2024-07-25
allow stitching using non-leaf processes in sum_mc_weight_per_process

mafrahm:
The stiched normalization weights Producer expected that only leaf process IDs are present in the `sum_mc_weight_per_process` dict. This PR should fix that.

### DIFF
--- a/columnflow/production/normalization.py
+++ b/columnflow/production/normalization.py
@@ -98,16 +98,16 @@ def get_br_from_inclusive_dataset(
             # skip processes that are not covered by any dataset or irrelevant for the used dataset
             # (identified as leaf processes that have no occurrences in the stats)
             # (or as non-leaf processes that are not in the stitching datasets)
-            is_leaf = str(child_proc.id) in sum_mc_weight_per_process.keys()
+            is_leaf = child_proc.is_leaf_process
+            child_in_weight_procs = str(child_proc.id) in sum_mc_weight_per_process
             if (
                 (is_leaf and str(child_proc.id) not in sum_mc_weight_per_process) or
                 (not is_leaf and child_proc.id not in proc_ds_map)
             ):
                 continue
 
-            proc_ids = [child_proc.id] if is_leaf else [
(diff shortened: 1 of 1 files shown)


columnflow/production/normalization.py
```diff
@@ -98,17 +98,22 @@ def get_br_from_inclusive_dataset(
...
                 (is_leaf and str(child_proc.id) not in sum_mc_weight_per_process) or
                 (not is_leaf and child_proc.id not in proc_ds_map)
             ):
                 continue
 
-            proc_ids = [child_proc.id] if is_leaf else [p.id for p in child_proc.get_leaf_processes()]
+            proc_ids = [child_proc.id] if is_leaf else [
+                p.id for p, _, _ in child_proc.walk_processes()
+                if p.id in sum_mc_weight_per_process.keys()
+            ]
```
haddadanas, line 111:
```suggestion
            proc_ids = [child_proc.id] if (is_leaf or child_in_weight_procs) else [
                p.id for p, _, _ in child_proc.walk_processes() if str(p.id) in sum_mc_weight_per_process
            ]
```

columnflow/production/normalization.py
```diff
@@ -98,17 +98,22 @@ def get_br_from_inclusive_dataset(
             # skip processes that are not covered by any dataset or irrelevant for the used dataset
             # (identified as leaf processes that have no occurrences in the stats)
             # (or as non-leaf processes that are not in the stitching datasets)
-            is_leaf = child_proc.is_leaf_process
+            is_leaf = str(child_proc.id) in sum_mc_weight_per_process.keys()
```
haddadanas, line 101:
```suggestion
            is_leaf = child_proc.is_leaf_process
            child_in_weight_procs = str(child_proc.id) in sum_mc_weight_per_process
```
I would not change the definition of is_leaf, since we need it in the if statement in line 102, but I would add a new variable

riga (approved):

PR #519, opened 2024-07-26
added feature to skip categories by custom function

haddadanas:
The category_ids producer has now the parameter skip_category_func, which by default is None, however, can be assigned a callable, to skip categories


riga (approved):
LGTM!

riga (approved):

PR #523, opened 2024-08-06
Overflow fix

JulesVandenbroeck:
Small fix to be able to include overflow bin in histogram plots. Before overflow bins were only included in case a variabe_inst.x.slice was defined.


JulesVandenbroeck:
This was meant to be a merge with the master branch on GhentAnalysis github instead of the central columnflow.

PR #524, opened 2024-08-07
Allow running over empty Nano files

mafrahm:
There are some NanoAOD files (*) with 0 events that fail when running through our workflow. This PR fixes that by skipping these files when running calibrators/selectors or when merging files.
Additionally, I also added a check for inconsistencies in number of fields when merging ReduceEvents outputs.

(*) example: file 0 from `data_muoneg_f`in the run3_2022_postEE_nano_v12 campaign:
```
# DAS
/MuonEG/Run2022F-22Sep2023-v1/NANOAOD
```
```
# file
/store/data/Run2022F/MuonEG/NANOAOD/22Sep2023-v1/50000/4d76213a-ef14-411a-9558-559a6df3f978.root
```

### DIFF
--- a/columnflow/tasks/calibration.py
+++ b/columnflow/tasks/calibration.py
@@ -139,7 +139,8 @@ class CalibrateEvents(
                 events = update_ak_array(events, *cols)
 
                 # just invoke the calibration function
-                events = self.calibrator_inst(events)
+                if len(events) != 0:
+                    events = self.calibrator_inst(events)
 
                 # remove columns
                 events = route_filter(events)
--- a/columnflow/tasks/reduction.py
+++ b/columnflow/tasks/reduction.py
@@ -484,18 +484,21 @@ class MergeReducedEvents(
 
     def merge(self, inputs, output):
         inputs = [inp["events"] for inp in inputs]
+        metadata = [pq.read_metadata(inp.path) for inp in inputs]
+
+        # empty nano files are not considered for merging
+        empty_nano = lambda meta: (meta.num_rows == 0) & (meta.num_columns <= 3)
 
         # check if the number of columns is consistent
-        metadata = [pq.read_metadata(inp.path) for inp in inputs]
-        unique_num_fields = set(meta.num_columns for meta in metadata if not meta.num_rows == 0)
+        unique_num_fields = set(meta.num_columns for meta in metadata if not empty_nano(meta))
         if len(unique_num_fields) > 1:
             raise Exception(
                 "cannot merge input files with inconsistent number of columns; found files with "
                 f"the following number of columns: {', '.join(map(str, unique_num_fields))}",
             )
 
-        # remove empty input files
-        cleaned_inputs = [inp for inp, meta in zip(inputs, metadata) if not meta.num_rows == 0]
+        # remove empty nano input files
+        cleaned_inputs = [inp for inp, meta in zip(inputs, metadata) if not empty_nano(meta)]
 
         # merge the input files
         law.pyarrow.merge_parquet_task(
--- a/columnflow/tasks/reduction.py
+++ b/columnflow/tasks/reduction.py
@@ -488,7 +488,7 @@ class MergeReducedEvents(
         with law.localize_file_targets(
             inputs, cache=False, mode="r",
         ) as localized_inputs:
-            
+
             metadata = [pq.read_metadata(inp.abspath) for inp in localized_inputs]
 
             # empty nano files are not considered for merging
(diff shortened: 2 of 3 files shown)


pkausw:
Thanks for this PR! @mafrahm and I had an offline chat about this project, and came up with a couple of things that should be checked before merging (collecting it here to keep track)
- [ ] does the check of the meta data work with remote files in the merging step?
- [ ] how costly would it be to perform a check for empty files already in the GetDatasetLFNs step?
- [ ] how costly would it be to compare the (sets of) names of the columns in the files to merge instead of the number of columns? 

Let me know if I forgot a check!

pkausw:
>     how costly would it be to perform a check for empty files already in the GetDatasetLFNs step?

Update: I have tested how much time it would take to directly check the number of events based on the dasgoclient. For this, I modified the `get_dataset_lfns` function:

```python
def get_dataset_lfns_filter_nevents(
            dataset_inst: od.Dataset,
            shift_inst: od.Shift,
            dataset_key: str,
        ) -> list[str]:
            """
            Get the LNF information with the ``dasgoclient``.

            :param dataset_inst: Current dataset instance, currently not used.
            :param shift_inst: Current shift instance, currently not used.
            :param dataset_key: DAS key identifier for the current dataset.
            :raises Exception: If query with ``dasgoclient`` fails.
            :return: The list of LFNs corresponding to the dataset with the identifier *dataset_key*.
            """
            import subprocess
            import json
            code, out, _ = law.util.interruptable_popen(
                f"dasgoclient --query='file dataset={dataset_key}' -json",
                shell=True,
                stdout=subprocess.PIPE,
                executable="/bin/bash",
            )
            if code != 0:
                raise Exception(f"dasgoclient query failed:\n{out}")

            das_info = json.loads(out.strip())
            valid_files = list(filter(
                lambda x: x["file"][0]["is_file_valid"]==1 and x["file"][0]["nevents"] >0, das_info
            ))
            return [
                x["file"][0]["name"]
                for x in valid_files
            ]
 ```
 
 I did a small-scale timing study with and without this function using the command
 
 ```bash
 time law run cf.GetDatasetLFNs --config run3_2022_postEE_cms_v12 --dataset data_muoneg_f --remove-output 0,a,y
 ```
 
 5 times. Here are the results:
 
 | | time (original) | time (nevents check) |
 | --- | --- | --- |
 | real | 21.1396 +- 0.607 | 11.8054 +- 0.3181 |
 | user | 24.4952 +- 0.43 | 12.8302 +- 0.3898 |
 | sys |  1.6606 +- 0.0832 | 0.6584 +- 0.0123 | 
 
 
 These values have to be taken with a grain of salt I guess - e.g. I don't really understand why the query becomes faster. Bottom line though is that performing this check doesn't seem to take longer, which would have the nice side effect that we don't need to consider the empty files in the branch map to begin with. 
 
 This of course would mean that we need to perform such a filter already on cmsdb level, since the files that the `dasgoclient` query delivers doesn't match with the number of files specified in the corresponding cmsdb campaign. I had a short offline discussion with @riga, and we think it would be nice to add broken/empty files as auxiliary information to the individual datasets. This way, we would be able to track these files and update the cmsdb entry when/if these issues are resolved in the future. For the dataset in question, this could look something like this
 
 ```python
 cpn.add_dataset(
    name="data_muoneg_f",
    id=14784482,
    is_data=True,
    processes=[procs.data_muoneg],
    keys=[
        "/MuonEG/Run2022F-22Sep2023-v1/NANOAOD",  # noqa
    ],
    n_files=95,
    n_events=38219969,
    aux={
        "era": "F",
        "broken_files": [
            '/store/data/Run2022F/MuonEG/NANOAOD/22Sep2023-v1/50000/4d76213a-ef14-411a-9558-559a6df3f978.root',
            '/store/data/Run2022F/MuonEG/NANOAOD/22Sep2023-v1/50000/4fb72196-3b02-4499-8f6c-a54e15692b32.root'
        ]
    },
)
```

In this example, the number of files to process would be `n_files - len(broken_files`. What do you think @mafrahm ?

pkausw:
Update on this item:

>     * [x]  does the check of the meta data work with remote files in the merging step?

-> no, it doesn't. I have pushed a fix that first localizes first the inputs and the output of the task, tested with CERN eos as remote target

pkausw:
I looked into the final point on the list @mafrahm . I implemented an explicit check for the column names and ran the MergeReducedEvents task over all 93 files of the above mentioned dataset. The timings I took indicate that there isn't a large run time difference between the versions - indeed, the version of the code with the new check seems to run ~10s faster on average than the current implementation (though the statistics is very limited of course).

With this in mind, I don't see a reason why we shouldn't check this. If you agree, I can push my implementation, but I'm also happy to discuss of course :+1:

riga:
Closed as discussed in todays meeting.

PR #526, opened 2024-08-07
Fix over/underflow handling for histograms

riga:
The over- and underflow handling implemented in our plot utils is returning a copy of the input histogram. However, the portion where handling is triggered treats it as being in-place. As a result, produced plots are always shown without flow handling, independent of the variable settings.

This PR updates `use_flow_bins` to consistently return a copy of the histogram (regardless of the flags), and changes adjust to code to handle it as a copy. This fixes the plot functions.

@aalvesan


mafrahm (approved):
LGTM, thank you for the fix!

PR #527, opened 2024-08-09
Add `strict` option to `find_config_objects`.

dsavoiu:
This PR adds a `strict` option to the `find_config_objects` method of framework base tasks. If set, an exception if any of the entries in `names` does not match at least one config object.

It could be useful to enforce this in some cases to make debugging easier, e.g. for typos in variable names. As an example of how this could work, an `allow_missing_variables` flag is added to the `VariablesMixin` class. Previously, if unknown or invalid `--variables` were supplied, they were silently skipped, whereas now an exception is raised.


riga:
Just pasting the outcome of last meetings discussion here to keep track of changes:

- Add `--strict-lookup` (bool) option directly to `ConfigTask`, with its default being extracted from `law.cfg` (perhaps from the `analysis` section).
- Use this flag in the config lookup methods.

dsavoiu:
Hi @riga, thanks for the chat yesterday! I had lost track of this PR, but I think it would be nice if we could take another look at this soon.

Particularly since #566 is expected to introduce larger changes to the task inheritance, it could also be a good time to understand if in the new `MultiConfigTask` approach we rely on any assumptions when doing config object lookups, and if it has any implications for this PR.

Regarding a `--strict-lookup` option for `ConfigTask` and corresponding `law.cfg` entry: I see how it could be useful to make the lookup behavior configurable in some way, but I'm not sure if a sensible default can be found for all situations where `find_config_objects` is called, since the desired behavior can in principle depend on the task.

In this PR I added a class variable for `VariablesMixin` as an example, since we probably never want to silently ignore variable specifications that don't match any known variable. However, in other cases where the `find_config_objects` function is used, it may well be that silent skipping is more desirable (e.g. when applying the lookup to multiple configs that can define different objects).

We should examine these use-cases and see if we can provide a more targeted way to control the lookup behavior. In the mean time, however, I would suggest to at least merge the patch adding the `strict` kwarg to `find_config_objects`. What do you think?

PR #528, opened 2024-08-09
Minor fixes for weight producers.

dsavoiu:
This PR adds some minor fixes related to weight producers.

* a custom `skip_func` is no longer ignored when `{data,mc}_only` are not given
* `CreateHistograms` checks if a custom `skip_func` exists before attempting to evaluate it
* the `empty` weight producer now returns `events` in addition to the weight array

### DIFF
--- a/columnflow/weight/__init__.py
+++ b/columnflow/weight/__init__.py
@@ -81,6 +81,7 @@ class WeightProducer(TaskArrayFunction):
                     raise Exception(
                         f"weight producer {cls_name} received both mc_only and data_only",
                     )
+
                 if mc_only or data_only:
                     if cls_dict.get("skip_func"):
                         raise Exception(
@@ -88,6 +89,7 @@ class WeightProducer(TaskArrayFunction):
                             "mc_only or data_only are set",
                         )
 
+                if "skip_func" not in cls_dict:
                     def skip_func(self):
                         # check mc_only and data_only
                         if getattr(self, "dataset_inst", None):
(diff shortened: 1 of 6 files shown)


pkausw (approved):
Thanks for the PR @dsavoiu ! LGTM. Though there is one thing I'm not sure about: I seem to remember a discussion about return type of weight producers (`weight` array vs `events, weight` tuple). Since both versions are currently implemented in the code (implementation in `empty` weight producer vs in `histograms`) I'm not quite sure which one we decided on in the end. Maybe @riga or @mafrahm remembers. If this is the return type we want to go for this PR should be ready to merge, but better make sure again

mafrahm:
Hi @pkausw, I am rather certain that we decided on `events, weight` as the output signature. I think the main reason was to prioritize consistency and flexibility over the possibility of user errors.

columnflow/weight/__init__.py
```diff
@@ -81,13 +81,15 @@ def update_cls_dict(cls_name, cls_dict, get_attr):
...
                     )
+
                 if mc_only or data_only:
                     if cls_dict.get("skip_func"):
                         raise Exception(
                             f"weight producer {cls_name} received custom skip_func, but either "
                             "mc_only or data_only are set",
                         )
 
+                if "skip_func" not in cls_dict:
```
riga, line 92: I had to think about the implications here a bit more, but I think this is what we want, i.e.

- complain if `mc_only` or `data_only` **and** a `skip_func` is set, and then
- create a `skip_func` if none is defined so far.

We should also add this subtle change to the other decorators that offer a `skip_func`. I can do that @dsavoiu and amend this PR, unless you want to go ahead :+1:

riga: Added this just now in 81ca127 for consistency with calibrators, selectors and producers.

riga (approved):
Let's go.

PR #529, opened 2024-08-09
Check all collections for colorbars when setting norm.

dsavoiu:
This PR contains a minor fix related to setting the scale on colorbars for 2D plots.

Previously, the `norm` object describing the scale was only applied the colorbar associated with the last collection on an axis, which did not consider the case where multiple collections were returned in a different order.

Currently, the `norm` is applied to all colorbars found for any collection.


mafrahm (approved):
LGTM

PR #530, opened 2024-08-13
Fix naf condor bug

riga:
This PR provides a hotfix for a cluster of bugs on the NAF and inside central lcg software stacks:

- python2.7 packages are added to naf worker nodes which breaks the consistency of our own software stack → thus, I removed all path fragments from `PATH` and `PYTHONPATH` variables that contain `"python2.7"`.
- the reason here is - most likely - that the alma9 lcg setup misses important libraries (such as libjson-c) that are needed by gfal → I initialized the central conda/mamba setup if available, which provides its own stack that includes a working version of gfal.

---

Not necessarily related, but while debugging (on a tangent as it turned out), I noticed that we do not have the option yet to overwrite the call methods of array functions, so I added a `@call` decorator.


pkausw (approved):
Thanks for the PR @riga ! I tested this in my local setup and noticed that `CF_CERN_USER_FIRSTCHAR` was missing in the remote env. Added it now to the bootstrap script, LGTM now :+1:

PR #531, opened 2024-08-15
Inference model improvements.

riga:
This PR provides a bunch of improvements to the inference model interface without breaking any previous behavior. It looks like a lot of changes, but most of them are actually just removing `InferenceModel` from `self: InferenceModel` type hints which is discouraged. Changes in particular:

- Cleaned up yaml representers in the custom `YamlDumper` class. The representer for `ParameterTransformations` was actually missing so yaml conversion did not work before. This is fixed now.
- `config_mc_datasets` and `config_data_datasets` can contain patterns now that are resolved in the `CreateDatacards` task. This is far more convenient than repeating longish lists in various places.
- The `cleanup` method is not called automatically anymore since it needs to be configurable. This should not break anything however. It accepts a `keep_parameters` option now to keep parameters in groups. Without that, the cleanup routine removes parameters that were not added to the datacard but perhaps through a custom combine physics model.
- Block formatting for parameter groups is fixed.
- Observed data events were written into the datacard as a float, though always with a trailing `.0`. This is converted to an integer now as it should be.

### DIFF
--- a/columnflow/inference/__init__.py
+++ b/columnflow/inference/__init__.py
@@ -188,26 +188,35 @@ class InferenceModel(Derivable):
     # optional initialization method
     init_func = None
 
-    class YamlDumper(yaml.SafeDumper or object):
+    class YamlDumper(yaml.SafeDumper):
         """
         YAML dumper for statistical inference models with ammended representers to serialize
         internal, structured objects as safe, standard objects.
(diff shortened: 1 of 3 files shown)


mafrahm (approved):
LGTM

PR #533, opened 2024-08-16
Fix cumulate

maadcoen:
mistake in cumulation function: reverse cumulation failed on any axis that wasn't the first


maadcoen:
sorry, wasn't meant to be pulled here...

PR #534, opened 2024-08-17
MultiConfigTask

### LINKED ISSUE #391: MultiConfigTask
There are many tasks that might require to access columns/histograms from multiple campaigns simultaneously.

- Machine Learning
- creating datacards
- rebinning
- plotting (e.g. combining 2022preEE and 2022postEE)

There is already the `MLModelTrainingMixin` [1] that allows us to train DNNs with multiple campaigns. However, this mixin cannot be reused for other types of MultiConfigTasks.
Instead we should build one central `MultiConfigTask` base task.

The `MLModelTrainingMixin` currently resolves Calibrators, Selector and Producers (CSPs) for each year individually. Therefore, we cannot rely on the already defined mixins for CSPs but have to reimplement the resolving of each of those.
This could be simplified by removing the resolving for each year individually, since users can simply write their CSPs such that they are year-independent. Such a Producer might look like this:

```python
@producer
def producer_run2(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
    year = self.config_inst.campaign.x.year

    if year == 2016:
        events = self[producer_2016](events, **kwargs)
    elif year == 2017:
        events = self[producer_2017](events, **kwargs)
    elif year == 2018:
        events = self[producer_2018](events, **kwargs)


def producer_run2_init(self: Producer) -> None:
    year = self.config_inst.campaign.x.year

    if year == 2016:
        self.uses.add(producer_2016)
        self.produces.add(producer_2016)
    elif year == 2017:
        self.uses.add(producer_2017)
        self.produces.add(producer_2017)
    elif year == 2018:
        self.uses.add(producer_2018)
        self.produces.add(producer_2018)
```

The main changes that are required in our current configs/tasks is that our mixins for CSPs + ML need to inherit from the `AnalysisTask` instead of the `ConfigTask` and that the default CSPs need to be defined in the `analysis_inst` instead of the `config_inst`

@riga 

[1] https://github.com/columnflow/columnflow/blob/e786a350b9d86e0c3e80b1e7ef710ff9fef81d9e/columnflow/tasks/framework/mixins.py#L1014

mafrahm:
This PR restructures the inheritance structure a bit to allow using mixins from CSPs in MultiConfigTasks.

- Mixins classes from CSPs are now inheriting from the AnalysisTask instead of the ConfigTask
    - this means that default CSPs now need to be set in the analysis_inst instead of the config_inst
- The MLTraining task now inherits from the CalibratorsMixin, SelectorMixin and ProducersMixin
    - This simplifies the resolving of CSPs for MLTraining as this is now taken care via the mixin classes
    - The MLTraining now needs to require the same CSPs for each config. If necessary, CSPs can be configured config dependant as shown in #391
- With MultiConfigTasks, there might be tasks that do not define a `config_inst`. Since many CSP inits expect that a config_inst is present, I added some code that sets the `config_inst` attribute to the first value of the `config_insts`. However, this might lead to inconsistencies between configs so we should keep an eye on that.

TODO:

- [x] Fix tests
- [ ] Decide on how to handle inits for MultiConfigTask
- [ ] Decide which Mixins should be AnalysisTask and which should be ConfigTask (WeightProducerMixin and InferenceModelMixin?)

Closes #391

### DIFF
--- a/analysis_templates/cms_minimal/__cf_module_name__/config/analysis___cf_short_name_lc__.py
+++ b/analysis_templates/cms_minimal/__cf_module_name__/config/analysis___cf_short_name_lc__.py
@@ -109,11 +109,11 @@ for dataset_name in dataset_names:
 verify_config_processes(cfg, warn=True)
 
 # default objects, such as calibrator, selector, producer, ml model, inference model, etc
-cfg.x.default_calibrator = "example"
-cfg.x.default_selector = "example"
-cfg.x.default_producer = "example"
+ana.x.default_calibrator = "example"
+ana.x.default_selector = "example"
+ana.x.default_producer = "example"
 cfg.x.default_weight_producer = "example"
-cfg.x.default_ml_model = None
+ana.x.default_ml_model = None
 cfg.x.default_inference_model = "example"
 cfg.x.default_categories = ("incl",)
 cfg.x.default_variables = ("n_jet", "jet1_pt")
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -1612,10 +1612,15 @@ class ArrayFunction(Derivable):
     ) -> None:
         super().__init__()
 
-        if not hasattr(self, "config_inst") and hasattr(self, "config_insts"):
-            self.config_inst = self.config_insts[0]
-        elif not hasattr(self, "config_inst") and not hasattr(self, "config_insts"):
-            return
+        # NOTE: it might be necessary/advantageous to skip CSP init functions when no config_inst is present
+        # currently this is not needed since all tasks are either a *ConfigTask* or a
+        # *MultiConfigTask* (which sets config_inst = config_insts[0]). However, we should keep
+        # in mind that this might be problematic in the future.
+
+        # if not hasattr(self, "config_inst") and hasattr(self, "config_insts"):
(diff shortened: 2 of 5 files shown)


mafrahm:
I just ran a test analysis from `SelectEvents` up to `PlotVariables1D` with `MLTraining` included and everything seems to be working.

JulesVandenbroeck:
Dear Mathis,

I am doing a run 3 (2022 & 2023) analysis and was looking into making combined plots of all 4 campaigns together. I was wondering if `MultiConfigTask` created in this branch is stable enough to incorporate in my analysis for this combined plotting task or if I could help with testing in case it is not?

Kind Regards,
Jules

mafrahm:
Hi @JulesVandenbroeck, great to hear that there are other people that could benefit from these changes!
Right now, there is no task for the multi config plotting, but it would be nice if you could prototype this using the `MultiConfigTask`, that would be a really good use case to test. One thing that should be changed for a `MultiConfigTask` is that the `WeightProducerMixin` should also be inheriting from `AnalysisTask` instead of `ConfigTask`. I am certain there will be other complications (e.g. variables are still defined per config, so what is being used in a `MultiConfigPlot` task?), so feel free to message me on Mattermost if you have anything to discuss.

mafrahm:
closed because superseded by #566

PR #537, opened 2024-08-21
Fix issues with CMSSW sandboxes

### LINKED ISSUE #535: cmssw_columnar.sh sandbox not working in htcondor workflow
### Discussed in https://github.com/columnflow/columnflow/discussions/525

<div type='discussions-op-text'>

<sup>Originally posted by **JulesVandenbroeck** August  7, 2024</sup>
Hi,

I require the cmssw_columnar.sh sandbox for pyROOT BDT purposes in my calibration task
. I have setup the sandbox and it works fine when running locally. However when running the task with `--workflow htcondor` venv setup does not work properly as the error `ModuleNotFoundError: No module named 'numpy'` is shown in the job log. I tried uninstalling the cmssw_columnar software saved in `data` as well as rerun cf.BundleCMSSWSandbox. I also placed the job log file in attachment.

Kind Regards,
Jules Vandenbroeck

[stdall_7To8.txt](https://github.com/user-attachments/files/16522801/stdall_7To8.txt)</div>

pkausw:
As was pointed out in issue #535 and discussions #525 and #442 , CMSSW sandboxes currently don't work as intended. There are two causes for this behavior that are fixed by this PR:

- the `cmssw_columnar` sandbox doesn't work on htcondor. This is due to overlapping dependencies between the columnar and dev sandboxes when the columnar requirements are installed with `pip install`. Fixed by performing the installation in isolation mode (`pip install -I`)
- the CMSSW sandboxes can only process local files because `gfal2` is not available. This is a result of manually removing the path to the micromamba environment from the `PYTHONPATH`. Fixed by not doing that anymore.

This is a very small PR, but I think a review would be beneficial mainly because of the second point - based on the comment in the code, there was a reason for this choice. I have tested the implementation in this PR and found no issues, but perhaps I missed something.

Closes #535

### DIFF
--- a/sandboxes/_setup_cmssw.sh
+++ b/sandboxes/_setup_cmssw.sh
@@ -333,7 +333,8 @@ setup_cmssw() {
 
     # prepend persistent path fragments again to ensure priority for local packages and
     # remove the conda based python fragments since there are too many overlaps between packages
-    export PYTHONPATH="${CF_PERSISTENT_PATH}:$( echo ${PYTHONPATH} | sed "s|${CF_CONDA_PYTHONPATH}||g" )"
+    # export PYTHONPATH="${CF_PERSISTENT_PATH}:$( echo ${PYTHONPATH} | sed "s|${CF_CONDA_PYTHONPATH}||g" )"
+    export PYTHONPATH="${CF_PERSISTENT_PATH}:$( echo ${PYTHONPATH} )"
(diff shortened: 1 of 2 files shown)


riga:
Closed following a private discussion. Keeping certain python packages coming with the conda env can (and has) lead to inconsistencies within the CMSSW env (this was observed before in a version clash between numpy in conda and tf in cmssw).

PR #539, opened 2024-08-22
Include hist hook for datacards

pkausw:
In the current version, the `cf.CreateDatacards` task cannot use hist hooks. This PR implements this feature. The histograms are first loaded in the same format as in the plotting routines for consistency. Moreover, process objects can now indicate that they should be dereived dynamically with a new property called `is_dynamic`. The datasets attached to dynamic processes are not considered when resolving the task requirements, which should reduce unnecessary overhead.

Additionally, this PR also updates the doc strings of the `InferenceModel` class. No new information was added, but now the parameters should be parsed properly.

This PR needs to be tested a little more, but should be mergable soon.


mafrahm:
Hi @pkausw, I just thought about the usecase of rebinning for hist hooks in datacards. We typically use the same observable/variable in multiple categories, but we would like to perform the rebinning for each category separately. Would that be possible with your PR?

pkausw:
> Hi @pkausw, I just thought about the usecase of rebinning for hist hooks in datacards. We typically use the same observable/variable in multiple categories, but we would like to perform the rebinning for each category separately. Would that be possible with your PR?

In principle it would be possible since you can create new histograms from scratch. Not sure if this nice and easy to understand, but I guess that is up to the analyzers :+1:

columnflow/tasks/cms/inference.py
```diff
@@ -220,15 +230,16 @@ def run(self):
...
-                        "category": [
-                            hist.loc(c.id)
-                            for c in leaf_category_insts
-                            if c.id in h.axes["category"]
-                        ],
+                        # "category": [
+                        #     hist.loc(c.id)
+                        #     for c in leaf_category_insts
+                        #     if c.id in h.axes["category"]
+                        # ],
```
mafrahm, line 237:
```suggestion
```

mafrahm (approved):

pkausw:
After talking to @riga, I adopted the docs part of this PR such that the `automodule` command for sphinx also works with empty `__all__` lists. This PR should now be ready for merging

riga (changes requested):
Thanks!

I just had two minor comments for which I provided code suggestions, looks good otherwise.

pkausw:
Thanks for the review @riga, I implemented your suggestions :+1:

riga (approved):
LGTM!

PR #540, opened 2024-08-22
Round the value for Lumi for 2d Plots

haddadanas:


riga (approved):
Thanks for the fix!

We had this already in place for 1D plots so it's nice to see the same fix being applied to all 2D things.

PR #548, opened 2024-10-14
filter dataset lfns based on broken_files aux

mafrahm:
Uses aux info from the cmsdb that is added in this PR: https://github.com/uhh-cms/cmsdb/pull/63


pkausw (approved):
LGTM, thanks :+1:

PR #553, opened 2024-10-23
Corrected paths used in union.py task

Bogdan-Wiederspan:
After fixing mirror targets in law (ff594e7670d236b6676773d6e5d5eb6b42c4d4c5), the paths used in union.py should resolve to absolute path and not only the path.


riga:
@Bogdan-Wiederspan This is indeed an important generalization, especially for local targets whose root is not `/` but a specific mount and directory as defined in the target fs' `base`.

I ported to changes to other places in `tasks/` where `.path` was used before. Can you check if your tests still run with these changes?

Bogdan-Wiederspan:
Did the a rerun from scratch. Looks good to me.

riga (approved):
Ok, thanks!

PR #556, opened 2024-10-24
Update muon and electron SF producers (backward compatible)

nprouvost:
create ElectronSFConfig and MuonSFConfig objects and update producer accordingly

also add Trigger SF Producer


mafrahm:
Regarding your TODO, I don't think that we need any code changes to cover the trigger SFs. You can simply derive new producers to produce different sets of correction weights. I am already doing that in my analysis:

https://github.com/uhh-cms/hh2bbww/blob/master/hbw/production/weights.py#L117-L128

nprouvost:
> Regarding your TODO, I don't think that we need any code changes to cover the trigger SFs. You can simply derive new producers to produce different sets of correction weights. I am already doing that in my analysis:
> 
> https://github.com/uhh-cms/hh2bbww/blob/master/hbw/production/weights.py#L117-L128

After discussion with marcel we decided to add it, since it is something every cms analysis should have (if they plan to use muon triggers) and it shows an example on how to do it if different derives are needed. (Actually, is it in the documentation at all at the moment? I can't remember) So it is in fact not a TODO anymore but part of the PR

columnflow/production/cms/electron.py
```diff
@@ -14,19 +16,40 @@
...
+        if self.working_point and self.hlt_path:
+            raise ValueError("only one of working_point or hlt_path must be set")
+
+    @classmethod
+    def new(
+        cls,
+        obj: ElectronSFConfig | tuple[str, str, str],
+    ) -> ElectronSFConfig:
+        # purely for backwards compatibility with the old tuple format
+        return obj if isinstance(obj, cls) else cls(*obj)
```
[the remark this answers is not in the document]
nprouvost: done, does this fit your needs?

columnflow/production/cms/muon.py
```diff
@@ -14,6 +16,27 @@
...
+    ) -> MuonSFConfig:
+        # purely for backwards compatibility with the old tuple format
+        if isinstance(obj, cls):
+            return obj
+        elif isinstance(obj, list) or isinstance(obj, tuple) or isinstance(obj, set):
+            return cls(*obj)
+        elif isinstance(obj, dict):
+            return cls(**obj)
+        else:
+            raise ValueError(f"cannot convert {obj} to MuonSFConfig")
```
riga, line 37: De-nesting. Also, unordered sets are not supported.

```suggestion
        if isinstance(obj, cls):
            return obj
        if isinstance(obj, str):
            return cls(obj)
        if isinstance(obj, (list, tuple)):
            return cls(*obj)
        if isinstance(obj, dict):
            return cls(**obj)
        raise ValueError(f"cannot convert {obj} to MuonSFConfig")
```

columnflow/production/cms/muon.py
```diff
@@ -14,6 +16,27 @@
 ak = maybe_import("awkward")
 
 
+@dataclass
+class MuonSFConfig:
+    correction: str
+    campaign: str
```
riga, line 22: Optional.

```suggestion
    campaign: str = ""
```

riga (changes requested):
Looks very good, thanks @nprouvost .

I made some inline coding suggestions and added them, nothing to complain about the overall logic and backwards compatibility :+1:

columnflow/production/cms/electron.py
```diff
@@ -14,6 +16,35 @@
...
+            return cls(*obj)
+        elif isinstance(obj, dict):
+            return cls(**obj)
+        else:
+            raise ValueError(f"cannot convert {obj} to ElectronSFConfig")
+
+
 @producer(
     uses={"Electron.{pt,eta,deltaEtaSC}"},
     produces={"electron_weight{,_up,_down}"},
```
riga, line 50:
```suggestion
    produces in the init
```

riga (changes requested):

columnflow/production/cms/electron.py
```diff
@@ -14,15 +16,46 @@
...
+        elif isinstance(obj, dict):
+            return cls(**obj)
+        else:
+            raise ValueError(f"cannot convert {obj} to ElectronSFConfig")
+
+
 @producer(
     uses={"Electron.{pt,eta,deltaEtaSC}"},
-    produces={"electron_weight{,_up,_down}"},
+    produces in the init
```
riga, line 50:
```suggestion
    # produces in the init
```

riga (changes requested):

riga (approved):

PR #561, opened 2024-11-08
Add fallback to nano file access via pre-emptive xrdcp

riga:
The tasks `{Calibrate,Select,Reduce}Events` are the only spots where initial input files / lfns are accessed and fetched. In case the file is remote, they use `gfal2` for copying them to a local location (either tmp or cached). The logic is encapsulated into the `iter_nano_files` method of the `GetDatasetLFNs` task, that is required by the three tasks mentioned before.

However, when a calibrator or selector requires a sandbox that is incompatible with ``gfal2``, such as all CMSSW-based ones, this method will not work. Therefore, this PR implements a way to fetch lfns via fallbacks. For now, I implemented a fallback using `xrdcp` for the case that the protocol is `root://`, but the method can be extended for other implementations if we need them at a later stage.

The fallback even makes use of local wlcg caches so that already fetched files (e.g. from an earlier tasks that *was* compatible with ``gfal2``) do need to be re-fetched.

### DIFF
--- a/columnflow/tasks/external.py
+++ b/columnflow/tasks/external.py
@@ -227,11 +227,10 @@ class GetDatasetLFNs(DatasetTask, law.tasks.TransferLocalFile):
 
                 # try an optional fallback to pre-emptively fetch the lfn if necessary
                 if not is_local and not skip_fallback:
-                    # TODO: improve tmp file location, maybe still use the wlcg fs' cache if set
-                    input_file = law.LocalFileTarget(is_tmp="root")
-                    input_file = self._fetch_lfn_fallback(lfn, selected_fs, input_file)
+                    input_file, input_stat, is_tmp = self._fetch_lfn_fallback(lfn, selected_fs)
                     if input_file:
-                        input_stat = input_file.stat()
+                        if is_tmp:
+                            input_file.is_tmp = True
                         task.publish_message(f"using fs {selected_fs} via pre-emptive fetch")
                         break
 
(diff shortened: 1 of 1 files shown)


riga:
@LuSchaller @jolange

columnflow/tasks/external.py
```diff
@@ -156,6 +156,7 @@ def iter_nano_files(
         fs: str | Sequence[str] | None = None,
         lfn_indices: list[int] | None = None,
         eager_lookup: bool | int = 1,
+        skip_fallback: bool = False,
```
pkausw, line 159: Can you add this parameter to the doc string?

columnflow/tasks/external.py
```diff
@@ -269,6 +279,73 @@ def iter_nano_files(
...
+        force: bool = False,
+    ) -> tuple[law.LocalFileTarget | None, os.stat_result | None, bool]:
+        """
+        Fetches an *lfn* via fallback mechanisms. Currently, only ``xrdcp`` for remote file systems
+        *selected_fs* with `root://` bases is supported. Unless *force* is *True*, no fallbacks are
+        performed in case they are not necessary in the first place (determined by the availability
+        of the ``gfal2`` package).
+
+        :param lfn: Logical file name to fetch.
+        :param selected_fs: Name of the file system to fetch the LFN from.
```
pkausw, line 295: It might make sense to explicitly state that the root protocol needs to be defined in the `selected_fs` in the law.cfg as either `base` or `base_filecopy`

columnflow/tasks/external.py
```diff
@@ -269,6 +279,73 @@ def iter_nano_files(
...
+            with wlcg_fs.cache.lock(lfn):
+                wlcg_fs.cache.allocate(stat.st_size)
+                clfn = law.LocalFileTarget(wlcg_fs.cache.cache_path(lfn))
+                destination.move_to_local(clfn)
+            return clfn, stat, False
+
+        # here, the destination will be temporary, but set its tmp flag to False to prevent its
+        # deletion when this method goes out of scope, and set the decision for later use instead
+        destination.is_tmp = False
+        return destination, stat, True
```
pkausw, line 347: Why is this necessary? In L233, you set `destination.is_tmp=True` in any case, so why set it `False` in the first place?

pkausw (changes requested):
Hi @riga , thanks a lot for this workaround! I think a lot of people can benefit from that. Just some minor (doc-related) comments and a question, nothing critical

columnflow/tasks/external.py
riga: Added.

columnflow/tasks/external.py
riga: Added.

PR #563, opened 2024-11-15
Enhance histogram task with selection handling

haddadanas:
This PR adds:

- the possibility to utilize the selection parameter int eh variable instances of order to apply a selection in the `CreateHistogram` task on the fly.
- short documentation of the functionality above :)

### DIFF
--- a/columnflow/tasks/histograms.py
+++ b/columnflow/tasks/histograms.py
@@ -5,8 +5,6 @@ Task to produce and merge histograms.
 """
 
 from __future__ import annotations
-from functools import reduce
-from operator import and_
 
 import luigi
 import law
@@ -228,16 +226,18 @@ class CreateHistograms(
                         # enable weights and store it
                         histograms[var_key] = h.Weight()
 
-                    # create event mask
-                    masks = []
+                    # mask events and weights when selection expressions are found
+                    masked_events = events
+                    masked_weights = weight
                     for variable_inst in variable_insts:
                         sel = variable_inst.selection
-                        if sel != "1":
-                            if not callable(sel):
-                                raise ValueError(f"invalid selection '{sel}', for now only callables are supported")
-                            masks.append(sel(events))
-                    mask = reduce(and_, masks, np.ones(len(events), dtype=bool))
-                    masked_events = events[mask]
+                        if sel == "1":
+                            continue
+                        if not callable(sel):
+                            raise ValueError(f"invalid selection '{sel}', for now only callables are supported")
+                        mask = sel(masked_events)
+                        masked_events = masked_events[mask]
+                        masked_weights = masked_weights[mask]
 
                     # merge category ids
                     category_ids = ak.concatenate(
@@ -250,7 +250,7 @@ class CreateHistograms(
                         "category": category_ids,
                         "process": masked_events.process_id,
                         "shift": np.ones(len(masked_events), dtype=np.int32) * self.global_shift_inst.id,
-                        "weight": weight[mask],
+                        "weight": masked_weights,
                     }
                     for variable_inst in variable_insts:
                         # prepare the expression
@@ -261,9 +261,8 @@ class CreateHistograms(
                                 if len(events) == 0 and not has_ak_column(events, route):
                                     return empty_f32
                                 return route.apply(events, null_value=variable_inst.null_value)
-                        arr = expr(masked_events)
                         # apply it
-                        fill_data[variable_inst.name] = arr
+                        fill_data[variable_inst.name] = expr(masked_events)
 
                     # fill it
                     fill_hist(


mafrahm:
Hi @haddadanas, I have not yet checked your implementation, I just wanted to mention that it might already be a way to add event selections via the `weight_producer`. Did you already try that? And is there some advantage to add your way of implementing selections to variables?

haddadanas:
> Hi @haddadanas, I have not yet checked your implementation, I just wanted to mention that it might already be a way to add event selections via the `weight_producer`. Did you already try that? And is there some advantage to add your way of implementing selections to variables?

Hi @mafrahm,

The idea was, that the order objects `variable` already provide a `selection` parameter which we don't use. I think the only advantage of this method is the convenience (espeacially for simple one-line selections), but I haven't used a selection with the `weight_producer` and therefore cannot compare those methods in an unbiased way 😅  (I assume define a weight producer, which sets all unwanted events to 0, so probably not a big difference).

columnflow/tasks/histograms.py
```diff
@@ -148,12 +148,16 @@ def run(self):
...
             )
-            for inp in (
+            for inp in ((
                 [variable_inst.expression]
                 if isinstance(variable_inst.expression, str)
                 # for variable_inst with custom expressions, read columns declared via aux key
                 else variable_inst.x("inputs", [])
-            )
+            ) + (
+                variable_inst.x("inputs", [])
```
riga, line 157: There is a case now where `variable_inst.x("inputs", [])` is used twice. Pretty sure this is not intentional.

columnflow/tasks/histograms.py
```diff
@@ -243,8 +247,17 @@ def expr(events, *args, **kwargs):
...
                                 return route.apply(events, null_value=variable_inst.null_value)
+                        arr = expr(events)
+                        # prepare the selection
+                        sel = variable_inst.selection
+                        if sel != "1":
+                            if callable(sel):
+                                mask = sel(events)
+                                arr = ak.where(mask, arr, variable_inst.null_value)
+                            else:
+                                raise ValueError(f"invalid selection: {sel}")
```
riga, line 258: Just some de-nesting.

```suggestion
                            if not callable(sel):
                                raise ValueError(f"invalid selection: {sel}")
                            mask = sel(events)
                            arr = ak.where(mask, arr, variable_inst.null_value)
```

docs/user_guide/plotting.md
```diff
@@ -363,3 +363,51 @@ An example on how to implement such a plotting function is shown in the followin
 :start-at: def my_plot1d_func(
 :end-at: return fig, (ax,)
 ```
+
+## Applying a selection to a variable
+
+In some cases, you might want to apply a selection to a variable before plotting it.
+Instead of creating a new column with the selection applied, columnflow provides the possibility to apply a selection to a variable directly when histograming it.
+For this purpose, the ```selection``` parameter can be added in the variable definition in the config.
```
riga, line 371: Single backticks are sufficient for md. Same for all instances below.

docs/user_guide/plotting.md
```diff
@@ -363,3 +363,51 @@ An example on how to implement such a plotting function is shown in the followin
...
+This may look as follows:
+
+```python
+
```
riga, line 375: Can you remove the leading and trailing empty lines in code blocks?

riga (changes requested):
That's really great! Some variables indeed have an implicit phase space selection and it's good that we can define this now right on the variable, rather than by having an implicit accompanying category.

I think that's a different use case than what you suggested @mafrahm?

mafrahm:
Hi @riga, adding a mask during the `weight_producer` or via this mechanism are basically fulfilling the same usecase. The main difference is that this mechanism couples the selection to a single variable, while the `weight_producer` can be used if you need to apply the same selections to a bunch of variables. In any case, both mechanisms should work independent of each other and of the categorization.

riga (approved):

columnflow/tasks/histograms.py
```diff
@@ -243,8 +261,9 @@ def expr(events, *args, **kwargs):
                                 if len(events) == 0 and not has_ak_column(events, route):
                                     return empty_f32
                                 return route.apply(events, null_value=variable_inst.null_value)
+                        arr = expr(masked_events)
                         # apply it
-                        fill_data[variable_inst.name] = expr(events)
+                        fill_data[variable_inst.name] = arr
```
riga, line 266: I guess this could be reverted to a single line again like before.

columnflow/tasks/histograms.py
```diff
@@ -221,18 +228,29 @@ def run(self):
...
+                    # create event mask
+                    masks = []
+                    for variable_inst in variable_insts:
+                        sel = variable_inst.selection
+                        if sel != "1":
+                            if not callable(sel):
+                                raise ValueError(f"invalid selection '{sel}', for now only callables are supported")
+                            masks.append(sel(events))
+                    mask = reduce(and_, masks, np.ones(len(events), dtype=bool))
+                    masked_events = events[mask]
```
riga, line 240: Could we reduce events here iteratively?

```suggestion
                    # mask events and weights when selection expressions are found
                    masked_events = events
                    masked_weights = weights
                    for variable_inst in variable_insts:
                        sel = variable_inst.selection
                        if sel == "1":
                            continue
                        if not callable(sel):
                            raise ValueError(f"invalid selection '{sel}', for now only callables are supported")
                        mask = sel(masked_events)
                        masked_events = masked_events[mask]
                        masked_weights = masked_weights[mask]
```

Not that this has an overly large positive impact on performance (although a small one for sure), but it would be good to consistently start using this pattern :)

If we decide for that, the imports above and the use of `weights` below needed to be adapted as well.

riga (changes requested):
Thanks @haddadanas !

I had one main comment about how the iterative mask application is done and made a code suggestion that, when accepted, needs two additional changes. It's not a critical thing though but rather a choice to use a more efficient pattern when dealing with iterative masking. This might be more beneficial in other locations but since we often take inspiration from existing code, I think it's helpful to consistently use that :)

riga (approved):
🎉 LGTM

PR #568, opened 2024-11-25
Handle missing selector steps in `CreateCutflowHistograms` and related plot tasks.

### LINKED ISSUE #567: `PlotCutflow` and related tasks cannot handle processes with different `selector_steps`
Selectors may define additional selector steps depending on the dataset/process they are running on. For example, selectors for `data` usually define a `json` selector step due to the application of the JSON filter, but MC processes do not.

Currently, adding these to the `--selector-steps` while specifying datasets/processes with different selector steps defined results in an error during `CreateCutflowHistograms`. As a workaround, one can run the `CreateCutflowHistograms` tasks separately for each processes, but one then runs into issues when calling e.g. `PlotCutflow`, where the task either fails or results in the *x*-axis labels being incorrect.

A possible fix would be to ignore the missing steps for `CreateCutflowHistograms` and ensure that the missing steps are properly handles in the plotting functions. One can also make this behavior configurable via a command-line parameter. I've started work on a proposed fix and will open a PR to address it.

dsavoiu:
This PR addresses an edge case in cutflow-related tasks where processes define different sets of `selector_steps` (cf. #567)

A new law config parameter `analysis.missing_selector_step_strategy` is introduced:
* `error` (default): an exception is raised if a requested selector step is not defined by any process
* `skip`: the selector step is ignored and the produced histogram will not contain a corresponding entry in the `step` axis)
* `dummy`: the selector step is ignored, but the produced histogram will contain a corresponding entry in the `step` axis that is identical to the previous step (i.e. the missing step is treated as not having any effect on the event selection).

On the plotting side, histograms with different `step` axes are pre-processed to insert any missing steps before being passed, ensuring that all `step` axes are identical. The values and variances for the inserted `step` bins are taken from the previous existing step in the sequence.

To facilitate the dynamic resolution of steps per process and merging of histograms for plotting, the `step` axis of all cutflow histograms is initialized with an empty list and `growth=True`.

Closes #567.

### DIFF
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -706,6 +706,8 @@ class PlotCutflowVariables1D(
                 step_hists = OrderedDict(
                     (process_inst.copy_shallow(), h[{"step": hist.loc(step)}])
                     for process_inst, h in hists.items()
+                    # skip missing steps
+                    if step in h.axes["step"]
                 )
 
                 # call the plot function
@@ -728,6 +730,8 @@ class PlotCutflowVariables1D(
                 process_hists = OrderedDict(
                     (step, h[{"step": hist.loc(step)}])
                     for step in self.chosen_steps
+                    # skip missing steps
+                    if step in h.axes["step"]
                 )
 
                 # call the plot function
--- a/columnflow/tasks/cutflow.py
+++ b/columnflow/tasks/cutflow.py
@@ -64,6 +64,18 @@ class CreateCutflowHistograms(
     # strategy for handling missing source columns when adding aliases on event chunks
     missing_column_alias_strategy = "original"
 
+    # strategy for handling selector steps not defined by selectors
+    missing_selector_step_strategy = luigi.ChoiceParameter(
+        significant=False,
+        default=law.config.get_default("analysis", "missing_selector_step_strategy", "error"),
+        choices=("error", "skip", "dummy"),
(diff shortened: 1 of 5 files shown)


columnflow/tasks/cutflow.py
```diff
@@ -217,11 +231,21 @@ def get_point(mask=Ellipsis):
...
+                        if self.missing_selector_step_strategy == "error":
+                            raise ValueError(
+                                f"step '{step}' is not defined by selector {self.selector}",
+                            )
+                        elif self.missing_selector_step_strategy == "skip":
+                            continue
+                        elif self.missing_selector_step_strategy == "dummy":
+                            pass
+                        else:
+                            assert False
```
riga, line 243: Just a suggestion for de-nesting. Also, I skipped the AssertionError since the choice parameter already does a value validation. I think we can rely on this and don't need to raise here.

```suggestion
                        if self.missing_selector_step_strategy == "error":
                            raise ValueError(
                                f"step '{step}' is not defined by selector {self.selector}",
                            )
                        if self.missing_selector_step_strategy == "skip":
                            continue
```

riga (changes requested):
This sounds good and very useful!

I just had one inline comment regarding some de-nesting and error handling, but besides that I'd say we can merge.

dsavoiu:
Thanks for the review @riga! I applied the patch you suggested. Please let me know if you have any other comments or we can go ahead with the merge.

riga (approved):
Thanks! LGTM!

PR #569, opened 2024-11-25
Update package versions in sandboxes.

riga:
Before working on finalizing #222, we should update versions of packages in our current sandboxes, which is done in this PR.

The versions of all array core packages were updated as well, leading to some speed ups during processing. as it is right now, coffea cannot be updated. I left a note in the `columnar.txt` file where we also should paste a link to a bug report once opened (I need to confirm this, but this should not hold back the PR).

Could you try this too @haddadanas ?

@mafrahm

### DIFF
--- a/sandboxes/cf.txt
+++ b/sandboxes/cf.txt
@@ -1,8 +1,8 @@
-# version 12
+# version 13
 
-tenacity!=8.4.0
-luigi~=3.5.1
+#tenacity!=8.4.0
+luigi~=3.5.2
 scinum~=2.2.0
 six~=1.16.0
-pyyaml~=6.0.1
+pyyaml~=6.0.2
 typing_extensions~=4.12.2
--- a/sandboxes/columnar.txt
+++ b/sandboxes/columnar.txt
@@ -1,12 +1,20 @@
-# version 14
+# version 16
 
-awkward==2.6.6
-uproot==5.3.9
-pyarrow==16.1.0
-dask-awkward==2024.6.0
+# exact versions for core array packages
+awkward==2.7.1
+uproot==5.5.1
+pyarrow==18.0.0
+dask-awkward==2024.9.0
+correctionlib==2.6.4
+# newer coffea versions use the vector library in a suboptimal way:
+# when loading only the pt column of a "vector" array, the field cannot be accessed
+# since a check is in place to ensure that the three other columns (eta,phi,mass)
+# are present as well; however, we might not need them yet still cannot access pt;
+# see BUGREPORT-TO-BE-ADDED
 coffea==2024.6.1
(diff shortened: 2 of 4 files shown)


pkausw (approved):
From the point of view of #222 this PR should be good to go :+1:

haddadanas:
All the venvs are working perfectly up to the Plotting tasks!

pfackeldey:
Hi @riga,

A note regarding the `vector` access of fields after some have been deleted:

```python
import vector
import awkward as ak

vector.register_awkward()

vec = ak.Array(
  [
    [{"x": 1, "y": 1.1, "z": 0.1, "t": 0.5}, {"x": 2, "y": 2.2, "z": 0.2, "t": 0.5}],
    [],
    [{"x": 3, "y": 3.3, "z": 0.3, "t": 0.5}],
    [
      {"x": 4, "y": 4.4, "z": 0.4, "t": 0.5},
      {"x": 5, "y": 5.5, "z": 0.5, "t": 0.5},
      {"x": 6, "y": 6.6, "z": 0.6, "t": 0.5},
    ],
  ],
  with_name="Momentum4D",
)

# remove one field
del vec["x"]

# doesn't work
vec.x
# >> ValueError: array does not have azimuthal coordinates (x/px, y/py or rho/pt, phi): y, z, t

# does work
vec["y"]
# >> <Array [[1.1, 2.2], [], [3.3], [4.4, 5.5, 6.6]] type='4 * var * float64'>
```

This is the difference between bare awkward-array field access via `__getitem__` and `vector` behavior mixing properties that are accessed via `__getattr__`.

Hope this clarification helps regarding the note of the coffea update and the vector behaviors.

I think that one can rely on `__getitem__` is getting overwritten in `vector`, all mixin utilities (e.g. `pt`, `eta`, ...) are implemented as properties. 
Can you confirm this policy @Saransh-cpp?

Best, Peter


PS: It is however not prohibited to do that in general, one can in principle overwrite `__getitem__` :eyes: - behavior mixins can modify awkward-arrays in more-or-less _any_ way:
```python
class GetItem(ak.Array):
  def __getitem__(self, k):
    print(k)
    
ak.behavior["*", "GetItem"] = GetItem

ak.with_name(vec, "GetItem")["y"]
# >> 'y'
```

riga:
Thanks for the hint @pfackeldey !

I'm going to check if this works for us 👏

Saransh-cpp:
> Can you confirm this policy @Saransh-cpp?

Sounds right! This is one of the differences that popped up once Coffea was migrated.

PS: not relevant to this PR, but some discussion over properties v/s fields in vector - https://github.com/scikit-hep/vector/issues/498

PR #570, opened 2024-11-25
Add --htcondor-disk parameter.

riga:
This PR adds a parameter `--htcondor-disk` which sets the RequestDisk ClassAd for htcondor submission files. The default is empty, meaning that no ClassAd is set. However, this is configurable through the `law.cfg`.


mafrahm (approved):
LGTM

PR #571, opened 2024-11-25
Make merging tasks remote workflows.

riga:
As discussed in our previous in-person dev meeting, we want the `MergeReductionStats` and `MergeSelectionStats` tasks to be workflows
- since they might take a while to process, and
- for them not to absorb workflow-related parameters and stop them from being passed upstream.

This PR contains the necessary changes to make them workflows. This is rather simple on its own, but a couple additional changes were required to connect them properly from within downstream tasks which are already workflows, but most likely with different branching granularity.

@TomPD


mafrahm (approved):
LGTM

PR #574, opened 2024-11-26
do not remove anything from plotting work tree

### LINKED ISSUE #520: broken --remove-output / --print-status
I noticed the output removal and status printing via the command line doesn't work properly anymore. The cf.SelectEvents task is sometimes indicated a incomplete while it isn't. That's also recognised by columnflow that doens't rerun the task despite the output of --print-status. When trying to remove the output of via --remove-output of the cf.SelectEvents task, nothing is removed.

I've tracked down the issue to the recent addition of optional histograms at the selection state. Since these are optional, columnflow considers the SelectEvents task finished if only the hists.pickle file is missing. This seems to work well when running the workflow, but I didn't check why that works. When printing the status, I found it `TargetCollection.status_text` gets called, which calls `TargetCollection.count`, which calls `TargetCollection._iter_state`. When removing files, `SiblingFileCollectionBase.remove` gets called, which calls `TargetCollection.iter_existing` which again calls  `TargetCollection._iter_state`. 

Now,   `TargetCollection._iter_state` iterates over all the groups of files that are complete: that is, if columns, stats, results, and hists are all present. It can take into account optional files, like "hists" but it only does so if one passes it `optional_existing=True`. One can pass this argument to `count` and `iter_existing`, but the problem is that this isn't done by neither `status_text` not `remove`.

pkausw:
Currently, we remove everything above the `MergeHistograms` level when requesting any plotting function. As discussed in the last in-person dev meeting, this decreases the time it takes to start the actual task itself. However, this has some negative interference with other convenience functionalities, such as printing the status within the task tree or removing outputs for levels above 1.

This PR changes this behavior again.

Closes #520

### DIFF
--- a/columnflow/tasks/plotting.py
+++ b/columnflow/tasks/plotting.py
@@ -212,11 +212,6 @@ class PlotVariablesBaseSingleShift(
 
     def workflow_requires(self):
         reqs = super().workflow_requires()
-
-        # no need to require merged histograms since each branch already requires them as a workflow
-        if self.workflow == "local":
-            reqs.pop("merged_hists", None)
-
         return reqs
 
     def requires(self):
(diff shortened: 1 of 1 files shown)


mafrahm (approved):
LGTM

PR #575, opened 2024-11-26
Add deterministic seeds to jer smearing application

Bogdan-Wiederspan:
This PR improves the seeding generation algorithm and adds a mechanism to apply these in JER smearing. 

specific features that are added:
- A new flag is added to activate the deterministic smearing, otherwise the previous default behavior is used as fallback. 
- This new mechanism applys JER independent of the used chunk size, which is useful when someone wants to syncronize.
- Improve seeding algorithm utilizes more event statistics and primes. 
The result are more unique seeds ( collision rate around 0% ). The algorithm is also more stable, due to the removal of redundant broadcasting.


columnflow/calibration/cms/jets.py
```diff
@@ -841,11 +843,21 @@ def jer_setup(self: Calibrator, reqs: dict, inputs: dict, reader_targets: Insert
...
+                np.random.Generator(bit_generator(_seed)).normal(_loc, _scale)
+                for _loc, _scale, _seed in zip(loc, scale, seed)
+            ])
+        self.deterministic_normal = deterministic_normal
 
 #
 # single calibrator for doing both JEC and JER smearing
 #
 
+
```
riga, line 860: Stray change.

```suggestion
```

columnflow/calibration/cms/jets.py
```diff
@@ -563,6 +563,8 @@ def get_jer_config(self) -> DotDict:
     propagate_met=True,
     # only run on mc
     mc_only=True,
+    # use deterministic seeds for random smearing
+    use_deterministic_seeds=False,
```
riga, line 567: I forgot...we need to have the index (e.g. the *i*th random number to use per seed) configurable, and we can use the same index to en/disable to use of deterministic seeds.

```suggestion
    # use deterministic seeds for random smearing and
    # take the "index"-th random number per seed when not -1
    deterministic_seed_index=-1,
```

columnflow/calibration/cms/jets.py
```diff
@@ -647,10 +645,14 @@ def jer(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
...
 
     # -- stochastic smearing
-
     # normally distributed random numbers according to JER
-    jer_random_normal = ak_random(0, jer, rand_func=rand_gen.normal)
-
+    jer_random_normal = (
+        ak_random(0, jer, events.Jet.deterministic_seed, rand_func=self.deterministic_normal)
+        if self.use_deterministic_seeds else
+        ak_random(0, jer, rand_func=np.random.Generator(
```
riga, line 652:
```suggestion
        ak_random(0, jer, events.Jet.deterministic_seed, rand_func=self.deterministic_normal)
        if self.deterministic_seed_index >= 0
        else ak_random(0, jer, rand_func=np.random.Generator(
```

columnflow/calibration/cms/jets.py
```diff
@@ -647,10 +645,14 @@ def jer(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
...
+            np.random.SFC64(events.event.to_list())).normal,
+        )
+    )
```
riga, line 655: For code style.

```suggestion
    )

```

columnflow/calibration/cms/jets.py
```diff
@@ -841,11 +843,21 @@ def jer_setup(self: Calibrator, reqs: dict, inputs: dict, reader_targets: Insert
         for name, key in jer_keys.items()
     }
 
+    # use deterministic seeds for random smearing if requested
+    if self.use_deterministic_seeds:
+        bit_generator = np.random.SFC64
+        def deterministic_normal(loc, scale, seed):
+            return np.asarray([
+                np.random.Generator(bit_generator(_seed)).normal(_loc, _scale)
```
riga, line 851:
```suggestion
    if self.deterministic_seed_index >= 0:
        idx = self.deterministic_seed_index
        bit_generator = np.random.SFC64
        def deterministic_normal(loc, scale, seed):
            return np.asarray([
                np.random.Generator(bit_generator(_seed)).normal(_loc, _scale, size=idx + 1)[-1]
```

riga (changes requested):
Thanks @Bogdan-Wiederspan !

I made a couple changes to make the position / index of the random number being used configurable. The *index* then also serves as the switch to en/disable the feature.

columnflow/calibration/cms/jets.py
pkausw: I think there is a typo in this suggestion - the names of the input arguments does not match the name of the variables used in L852 @riga

riga (approved):
Looks good, but let's get @dsavoiu in the loop as well since he authored the jer calibration.

columnflow/calibration/cms/jets.py
Bogdan-Wiederspan: > I think there is a typo in this suggestion - the names of the input arguments does not match the name of the variables used in L852 @riga

Yeah its a little hard to see, but this is intended.  I use the variables immediately in a list comprehensions, where the variables are then used with an _.

riga (approved):

dsavoiu:
Thanks a lot @Bogdan-Wiederspan for the PR and @riga the review and the ping! I think it's in general a good idea to use the determinstic seeds for JER :D I'll have a look in more detail right away.

dsavoiu (changes requested):
Hi @Bogdan-Wiederspan, thanks again for the PR! I think this is really useful for making the JER smearing more stable, especially for synchronization exercises. I didn't get into the math of actually calculating the seeds, but in general it seems well thought out and has the advantage of using more of the event content than before.

I mostly have minor comments, but there is also one important point I think we should discuss before merging. This concerns the hard requirements for specific input columns that the `deterministic_seeds` function would have in the proposed implementation, since I think this could break current analysis workflows (more info below).

columnflow/production/cms/seeds.py
```diff
@@ -23,12 +18,15 @@
...
-        optional("Jet.nConstituents"), optional("Jet.nElectrons"), optional("Jet.nMuons"),
+        "Pileup.nPU",
+        # columns needed to extract object counts only
+        "Photon.pt", "SV.pt", "FatJet.pt", "SubJet.pt",
         optional("GenJet.pt"), optional("GenPart.pt"),
+        # per-object columns for further hashing
+        "Electron.jetIdx", "Electron.seediPhiOriY",
+        "Muon.jetIdx", "Muon.nStations",
+        "Tau.jetIdx", "Tau.decayMode",
+        "Jet.nConstituents", "Jet.nElectrons", "Jet.nMuons",
```
dsavoiu, line 29: Why is the `optional` flag being dropped here? Unfortunately, this will definitely introduce breaking changes for analyses that rely on running `determinisic_seeds` later in the workflow, e.g. during `ProduceColumns`, since in that case not all columns defined in the original NanoAOD would be available.

I think we should keep this implementation as flexible as possible and not impose any hard requirements on input columns. Can this be adapted to use the `optional` flag for all columns except the mandatory `run`, `luminosityBlock` and `event`?

dsavoiu (changes requested):
Seems like my main inline comment was not added to the review, so I'm resubmitting is here. Sorry for the spam!

columnflow/production/cms/seeds.py
riga: Yes, we can make them optional again. The motivation for dropping it was to be sensitive the possible broken inputs where certain columns should exist, but are in fact missing, leading to different seeds without knowing.

But in order not to break existing code, we could introduce a `missing_column_behavior`, that is `"ignore"` by default (old behavior), but can also be set to `"warning"` or `"raise"`.

columnflow/production/cms/seeds.py
```diff
@@ -138,6 +151,17 @@ def deterministic_jet_seeds(self: Producer, events: ak.Array, **kwargs) -> ak.Ar
...
+@deterministic_jet_seeds.setup
+def deterministic_jet_seeds_setup(
+    self: Producer,
+    reqs: dict,
+    inputs: dict,
+    reader_targets: InsertableDict,
+) -> None:
+    # store primes in array
+    self.primes = np.array(primes, dtype=np.uint64)
+
```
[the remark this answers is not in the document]
riga: Since the `init` is (and needs to be) called by every task (and every branch of every workflow) in the entire dependency tree below the producer / calibrator using it, I deemed `setup` to be a better place, which is only called once and in this way, sets up things that are only needed when subsequently calling the actual function payload.

columnflow/production/cms/seeds.py
```diff
@@ -119,15 +132,15 @@ def deterministic_jet_seeds(self: Producer, events: ak.Array, **kwargs) -> ak.Ar
     events = self[deterministic_event_seeds](events, **kwargs)
 
     # create the per jet seeds
-    prime_offset = 18
+    primes = self.primes[events.deterministic_seed % len(self.primes)]
     jet_seed = events.deterministic_seed + (
-        primes[prime_offset] * ak.values_astype(ak.local_index(events.Jet, axis=1), np.uint64)
+        primes * ak.values_astype(ak.local_index(events.Jet, axis=1) + 1, np.uint64)
```
riga, line 137: @Bogdan-Wiederspan Thinking about this again, we use the event seed to infer jet seeds using the local index. However, if we use the same idea for other objects (eg. leptons), per event we end up with the same seed for the objects (jet 0 will have the same seed as lepton 0, etc).

We should introduce an additional offset here and we should use a different one for future objects. I'd suggest the 50th prime rather than just `1`.

```suggestion
        primes * ak.values_astype(ak.local_index(events.Jet, axis=1) + self.primes[50], np.uint64)
```

columnflow/production/cms/seeds.py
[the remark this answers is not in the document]
dsavoiu: I see, that makes sense! Let's keep `setup` then 👍

columnflow/production/cms/seeds.py
dsavoiu: Right, that's also a valid point. The `missing_column_behavior` sounds like a reasonable way to make this more flexible, but I'm thinking whether we should implement that separately or extend the existing mechanisms to make sure unexpectedly missing columns don't go unnoticed.

We could in principle do this via the `check_used_columns` mechanism in the base class. At the moment that only supports either ignoring the columns or raising an exception, depending on whether the column is wrapped by the `optional` modifier. 

But maybe we can refine this behavior to issue a `warning` if certain optional columns are missing? This could be controlled by another aux flag à la `skip` or `optional`. And as a user interface we could add an `optional_warn` column wrapper that sets the flag. What do you think, @riga?

At any rate, the implementation would still need to check if these columns actually exist before using them to calculate the seeds.

riga: I like the idea of having a variant of the `optional` modifier that issues a warning in case columns are not existing. With that, all three use cases would be covered.

I would prepare this for another PR though if you agree.

> At any rate, the implementation would still need to check if these columns actually exist before using them to calculate the seeds.

In *this* PR, we should do exactly that imho. Would you be fine if we just default to printing warnings when columns are missing?

columnflow/production/cms/seeds.py
```diff
@@ -45,41 +43,54 @@ def deterministic_event_seeds(self: Producer, events: ak.Array, **kwargs) -> ak.
     """
     # start with a seed of the event number itself as the most sensitive integer
     # and define the offset of the first prime to use
```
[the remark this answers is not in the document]
riga: Changed in 406930d0.

columnflow/production/cms/seeds.py
riga: In 406930d0, we made the columns optional again and added the warnings :)

dsavoiu: OK, great! Yes, let's do the `optional` variant in another PR. For this PR it would be fine to just ignore the columns and directly print the warnings, as you suggest.

riga:
Thanks for the review @dsavoiu.

PR #577, opened 2024-11-27
Adapt JEC/JER calibrators to work with user-defined jet collections.

dsavoiu:
This PR implements JEC/JER for jet collections other than the standard AK4 jets.

A parameter `jet_name` is introduced to the `jec` and `jer` calibrators, allowing users to specify the input collection name.

To allow configuring JEC/JER for multiple input collections, the canonical form of the `jec` and `jer` configs is changed to include an additional layer specifying the input collection in NanoAOD. For example, the canonical config to configure JEC for AK4 jets changed as follows:
```diff
-cfg.jec = {
-    "campaign": "Summer19UL17",
-    ...,
-}
+cfg.jec = {
+    "Jet": {
+        "campaign": "Summer19UL17",
+        ...,
+    },
+}
```

Calibrators for custom jet collections may be created by deriving from the generic calibrators, e.g.:
```
my_jec = jec.derive("my_jec", cls_dict={"jet_name": "MyJetCollection"})
```

Explicit calibrators for the two standard NanoAOD jet collections are provided centrally (`je{c,r}_ak4` for `Jet`, `jec{c,r}_ak8` for `FatJet`). MET propagation is only performed for AK4 jets by default.

Implementation notes
------------

- replaced hard-coding of `Jet` in `jec` and `jer` calibrators; input collection now specified via new class attribute `jet_name` (and `gen_jet_name` for gen-jets used for JER)
- changes are fully backwards compatible, equivalent to using `Jet` as the default/only input collection
- a deprecation warning is issued if the `jet_name` is the default `Jet`, but config does not have the additional collection name layer
- the external file key of the input JSONs is determined automatically for the standard jet collections (`Jet` -> `jet_jerc`, `FatJet` -> `fatJet_jerc`)
- a new, optional `external_file_key` config entry can be set to override the default external file key (useful for `FatJet` JER, which are sometimes not in the same file as the JECs and/or have a different `jet_type`)
- the option to pass a custom `get_je{c,r}_file` method is unaffected by the above


dsavoiu:
@riga @Bogdan-Wiederspan This will likely have conflicts with #575 😅 But at first glance it seems to be a relatively straightforward merge.

riga:
With #575 merged, I don't see conflicts, so I'd just to ahead and pull the master back into this branch in order to start the review.

columnflow/calibration/cms/jets.py
```diff
@@ -111,58 +115,130 @@ def ak_evaluate(evaluator: correctionlib.highlevel.Correction, *args) -> float:
...
     """
-    Function to obtain external jec files.
+    Function to obtain external correction files for JEC and/or JER.
 
     By default, this function extracts the location of the jec correction
-    files from the current config instance *config_inst*:
+    files from the current config instance *config_inst*. The key of the
+    external file depends on the jet collection. For ``Jet`` (AK4 jets), this
+    resolves to ``jet_jerc``, and for ``FatJet`` it is resolved to
+    ``fatJet_jerc``.
```
riga, line 126: I think an all-lowercase `fatjet_jerc` key would be more consistent.
Would you mind renaming? (same for the occurrences below)

columnflow/calibration/cms/jets.py
```diff
@@ -111,58 +115,130 @@ def ak_evaluate(evaluator: correctionlib.highlevel.Correction, *args) -> float:
...
+    :return: Dictionary containing configuration for jet energy calibration
     """
-    return self.config_inst.x.jec
+    jec_cfg = self.config_inst.x.jec
+
+    # check for old-style config
+    if self.jet_name not in jec_cfg:
+        # if jet collection is `Jet`, issue deprecation warning
+        if self.jet_name == "Jet":
+            logger.warning_once(
```
riga, line 207: `warning_once` accepts two arguments with the first one being an identifier to check if this type of message should be skipped when logged again. E.g. `f"{id(self)}_depr_jec_config"`.

columnflow/calibration/cms/jets.py
```diff
@@ -111,58 +115,130 @@ def ak_evaluate(evaluator: correctionlib.highlevel.Correction, *args) -> float:
...
+
+
+# default external file keys for known jet collections
+get_jerc_file_default.map_jet_name_file_key = {
+    "Jet": "jet_jerc",
+    "FatJet": "fatJet_jerc",
+}
 
 
 def get_jec_config_default(self) -> DotDict:
```
riga, line 182: Could you add a type hint `self: Calibrator` here? Just to reflect to which object this function will be bound.

columnflow/calibration/cms/jets.py
```diff
@@ -503,40 +614,57 @@ def make_jme_keys(names, jec=jec_cfg, is_data=self.dataset_inst.is_data):
...
-    By default, this is extracted from the current *config_inst*:
-
-    .. code-block:: python
+    jer_cfg = self.config_inst.x.jer
+
+    # check for old-style config
+    if self.jet_name not in jer_cfg:
+        # if jet collection is `Jet`, issue deprecation warning
+        if self.jet_name == "Jet":
+            logger.warning_once(
```
riga, line 649: Same as above, the first argument could be `f"{id(self)_depr_jec_config}"`.

dsavoiu: Added as `f"{id(self)_depr_jer_config}"`.

columnflow/calibration/cms/jets.py
dsavoiu: Added as suggested.

columnflow/calibration/cms/jets.py
dsavoiu: Renamed to `fat_jet_jerc` as discussed privately.

dsavoiu:
Thanks a lot @riga for the review! I added a commit implementing your suggestions. It was also necessary to change one line of code as a result of PR #575 being merged, which still used hard-coded name for the jet collection when applying the JER smearing.

riga (approved):
Looks good! Let's merge

PR #579, opened 2024-12-01
Integrate chapter about systematic uncertainties into docs

pkausw:
This PR implements a chapter on systematic uncertainties and how they are implemented in columnflow in the docs.
In particular, the following things are discussed

- Simple rate uncertainties (example: Lumi)
- weight-based shape uncertainties (example: muon scale factors)
- shape uncertainties that impact the selected analysis phase space (example: jet energy corrections)
- shape uncertainties with dedicated datasets (example: tune uncertainty)

Each kind of uncertainty is explained by giving a motivation/use case, and the already existing code examples from the CMS analysis template are used explicitely.

Apart from this, this PR implements some minor changes:

- minor bug fix: the `columnar_utils` module is currently empty in the docs because the `__all__` list is set empty. Fixed by commenting this behavior out for now
- the `WeightProducer` modules are integrated in the api docs
- the jet selection now implements the jec as a shift

To have a look at the updated docs, you can do the following (from your analysis directory)

```bash
# checkout this branch
cd modules/columnflow
git fetch
git checkout origin/docs/systematics
cd ../..
# run your setup
source setup.sh YOUR_SETUP_NAME
# build the docs
./modules/columnflow/tests/run_docs
# host a local server
cd modules/columnflow/docs/_build/html
python3 -m http.server 8088
```

You can then look around in the docs in your browser under the address `http://localhost:8088`

### DIFF
--- a/docs/user_guide/shifts.md
+++ b/docs/user_guide/shifts.md
@@ -3,7 +3,7 @@
 
 Many aspects of the work of analyzers revolve around systematic uncertainties on the modeling of the simulation or in the conditions during data taking.
 columnflow is designed to support various types of systematic uncertainties, also referred to as *shifts*.
-We distinguish two general classes of uncertainties: 
+We distinguish two general classes of uncertainties:  
 
 - **rate uncertainties** only modify the overall yield of a distribution.
 - **shape uncertainties** can modify both the yield and the shape of distributions.
@@ -99,7 +99,6 @@ This way, the workflow is now fully aware of the shift.
 You can test your implementation e.g. by running the {py:class}`~columnflow.tasks.plotting.PlotShiftedVariables1D` task.
 For more information, see the [plotting overview](../task_overview/plotting_tasks.md).
 
-
 To include this shift in a statistical inference model, we need to add it as a parameter (see CMS analysis example):
 
 ```{literalinclude} ../../analysis_templates/cms_minimal/__cf_module_name__/inference/example.py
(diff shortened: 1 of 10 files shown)


docs/user_guide/shifts.md
```diff
@@ -0,0 +1,203 @@
...
+:dedent:
+:language: python
+:start-at: '# add column aliases for shift mu'
+:end-before: '# external files'
+```
+
+Note that this function makes some assumptions on the naming of the varied quantities and the shifts.
+
+Finally, we can now configure the modules within the workflow to use this shift.
+As mentioned above, this class of systematic uncertainties does not change any selection efficiencies.
```
mafrahm, line 81: The selection efficiency can be modified via event weights, but we can still reuse SelectEvents because the acceptance of any individual event is not changed.

```suggestion
As mentioned above, this class of systematic uncertainties does not affect the selection acceptance.
```

docs/user_guide/shifts.md
```diff
@@ -0,0 +1,203 @@
...
+:end-before: '# jet energy correction uncertainty'
+```
+
+In particular, note that this type of uncertainty needs to define the keyword argument `config_shift_source` in the {py:func}`~columnflow.inference.InferenceModel.add_parameter` function to properly resolve the necessary shifts to run through the workflow.
+
+## Uncertainties that modify the selection efficiency
+
+A more complex type of shape uncertainty arises from effects that can modify vital quantities for analysis decisions.
+One example for this class of uncertainties are calibrations of objects such as jets, which modify the four-momenta content of these objects.
+Since these calibrations are generally subject to uncertainties, the corresponding variations can propagate throughout the analysis workflow and modify for example selection efficiencies in the analysis phase space.
```
mafrahm, line 117:
```suggestion
Since these calibrations are generally subject to uncertainties, the corresponding variations can propagate throughout the analysis workflow and modify event and object selection acceptances.
```

docs/user_guide/shifts.md
```diff
@@ -0,0 +1,203 @@
...
+:end-before: '# external files'
+```
+
+Note that this function makes some assumptions on the naming of the varied quantities and the shifts.
+
+Finally, we can now configure the modules within the workflow to use this shift.
+As mentioned above, this class of systematic uncertainties does not change any selection efficiencies.
+Therefore, it is sufficient to consider the varied quantities in the {py:class}`~columnflow.tasks.histograms.CreateHistograms` task.
+
+The calculation of the final event weight is handles by the {py:class}`~columnflow.weight.WeightProducer` instance you specify at command-line level.
```
Lara813, line 84:
```suggestion
The calculation of the final event weight is handled by the {py:class}`~columnflow.weight.WeightProducer` instance you specify at command-line level.
```

docs/user_guide/shifts.md
```diff
@@ -0,0 +1,203 @@
...
+:dedent:
+:language: python
+:start-at: "# add column aliases for shift jec"
+:end-before: '# event weights due to muon scale factors'
+```
+
+Note that as opposed to the previous example, we use the `name` of the shift to construct the final name of the column containing the varied values.
+
+Finally, we need to make the relevant modules aware of the shift.
+As explained above, the jet energy calibration and its variations can have a non-trivial effect on the selection efficiency in the final analysis phase space.
```
mafrahm, line 149:
```suggestion
As explained above, the jet energy calibration and its variations can have a non-trivial effect on the selection acceptance in the final analysis phase space.
```

mafrahm (approved):
One final inline comment.

LGTM otherwise

PR #581, opened 2024-12-04
Make input columns for deterministic seeds configurable.

riga:
As a follow-up to #575, this PR makes the columns used for the deterministic seed determination configurable. The configs distinguish between three types of columns:

- `event_columns`: Columns that lead to existing event-level integer fields.
- `object_count_columns`: Columns to load that allow querying for the number of objects in a collection.
- `object_columns`: Columns that lead to object-level integer fields. Per unique collection (`Jet.`, `Muon.`, ...), one of these columns should be placed in `object_count_columns` instead of any other non-integer column (e.g. `Jet.pt`), in order to save IO.

As before, the default columns are optional to not break existing code.

Also, the code and comments switched between "columns", "fields", and "routes". I removed all occurrences of "field" for consistency.

### DIFF
--- a/columnflow/production/cms/seeds.py
+++ b/columnflow/production/cms/seeds.py
@@ -170,6 +170,7 @@ def deterministic_event_seeds_setup(
 
 
 @producer(
+    uses={"Jet.pt"},
     produces={"Jet.deterministic_seed"},
 )
 def deterministic_jet_seeds(self: Producer, events: ak.Array, **kwargs) -> ak.Array:
@@ -179,11 +180,14 @@ def deterministic_jet_seeds(self: Producer, events: ak.Array, **kwargs) -> ak.Ar
     :py:func:`deterministic_event_seeds` which is not called by this producer for the purpose of
     of modularity. The strategy for producing seeds is identical.
     """
-    # create the per jet seeds
+    # get sorting indices for jets using their pt as a local index is used for the hashing
+    sorting_indices = ak.argsort(-events.Jet.pt, axis=1)
+
+    # create the seeds
     primes = self.primes[events.deterministic_seed % len(self.primes)]
     jet_seed = events.deterministic_seed + (
         primes * ak.values_astype(ak.local_index(events.Jet, axis=1) + self.primes[50], np.uint64)
-    )
+    )[sorting_indices]
     np_jet_seed = np.asarray(ak.flatten(jet_seed))
     np_jet_seed[:] = create_seed_vec(np_jet_seed)
 
--- a/columnflow/production/cms/seeds.py
+++ b/columnflow/production/cms/seeds.py
@@ -21,6 +21,17 @@ ak = maybe_import("awkward")
 logger = law.logger.get_logger(__name__)
 
 
+def create_seed(val: int, n_hex: int = 16) -> int:
+    """
+    Create a seed from an integer value by hashing it and returning the trailing 64 bit integer.
+    """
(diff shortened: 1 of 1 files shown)


columnflow/production/cms/seeds.py
```diff
@@ -25,19 +25,23 @@
...
-        }),
+        # all other columns to be loaded are defined through the lists below
     },
     produces={"deterministic_seed"},
+    # columns of event integer fields
+    event_columns=[optional("Pileup.nPU")],
+    # columns to be loaded to infer counts of objects
+    object_count_columns=list(map(optional, [
+        "Jet.nConstituents", "FatJet.pt", "SubJet.pt", "Photon.pt", "Muon.jetIdx",
+        "Electron.jetIdx", "Tau.jetIdx", "SV.pt", "GenJet.pt", "GenPart.pt",
```
pkausw, line 36: Why don't we use one observable consistently, e.g. pt? Is this just for demonstration purposes or is there a deeper reason?

columnflow/production/cms/seeds.py
```diff
@@ -117,6 +118,17 @@ def deterministic_event_seeds(self: Producer, events: ak.Array, **kwargs) -> ak.
...
 
+@deterministic_event_seeds.init
+def deterministic_event_seeds_init(self: Producer) -> None:
+    """
+    Producer initialization that adds columns to the set of *used* columns based on the
+    *event_columns*, *object_count_columns*, and *object_columns* lists.
+    """
+    # add used columns
+    for column in self.event_columns + self.object_count_columns + self.object_columns:
+        self.uses.add(Route(column))
```
pkausw, line 129: This can be simplified

```suggestion
    self.uses.update(self.event_columns, self.object_count_columns, self.object_columns)
```

pkausw (changes requested):
Thanks for this update @riga ! Just minor comments

columnflow/production/cms/seeds.py
riga: Because, e.g., we already need `Tau.jetIdx` as an object column, and upon loading, we can infer the object counts from there, so there's no need to load an additional pt column per tau.

columnflow/production/cms/seeds.py
riga: This misses the `Route()` cast.

pkausw (approved):

columnflow/production/cms/seeds.py
```diff
@@ -125,53 +148,44 @@ def deterministic_event_seeds_setup(
...
             )
             return None
 
-    self.apply_field = apply_field
+    self.apply_route = apply_route
 
 
 @producer(
-    uses={deterministic_event_seeds},
     produces={"Jet.deterministic_seed"},
```
dsavoiu, line 173: I understand the advantages of removing the `deterministic_event_seeds` producer from the requirements, but this implementation still accesses the `"deterministic_seed"` column, so we should at least add that to the `uses`.

dsavoiu (changes requested):
Thanks for the PR! This looks good to me, I just have a minor comment on the input columns for `deterministic_jet_seeds` below.

columnflow/production/cms/seeds.py
riga: I think this would fail if the input file does not already contain a column `deterministic_seed`.

Now that I'm thinking about that, this check should actually be changed to consider columns that are produced along the way. We have all information necessary for detecting that, but I guess this would be another PR (we can talk about this offline).

PR #586, opened 2024-12-10
Preserve original parquet file structure after merging.

### LINKED ISSUE #583: IO parallelization
Our IO is designed to load chunks of events in parallel using a thread pool, effectively bridging IO wait gaps. This worked well for some time, but as recently noticed by @mafrahm, our memory footprint suddenly started scaling with the size of the thread pool.

The reason for that is a change in `dask_awkward` which causes all partitions of an input parquet file to be considered as a single, large partition. As a result, each thread loads the full file into memory, independent of the slice that selects the chunk.

We can go back to the old, "non-merging" behavior by setting `split_row_groups=True` when loading the parquet file. However, after debugging this with @pfackeldey, we found an issue in either `dask_awkward` or `awkward` which occurs when materializing parts of a file with multiple, preserved partitions.

I opened an issue here: https://github.com/scikit-hep/awkward/issues/3331

Once resolved, we should use row group splitting and go back to our default `MaterializationStrategy.SLICES` strategy in the chunked IO reader.

riga:
This PR updates the merging of parquet files after chunked processing.

Our processing happens in three steps.

1. Chunks of events are loaded and processed by *task array functions*.
2. The resulting *new* columns are saved in parquet files via `ak.to_parquet`. **Here, we rely on awkward to pass specific settings to the underlying parquet writer**.
3. After all chunks are processed, we merge these parquet files. **This merging is happening without awkward, using a bare `pyarrow.parquet.ParquetWriter` instance**. However, for full consistency with the format expected by awkward and dask_awkward when reading these files again in a downstream task, a setting `use_compliant_nested_type=False` was missing. This setting makes sure that the internal node / tree structure is consistent w.r.t. the input files and not changed as described [here](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html#pyarrow-parquet-parquetwriter).

Memory is stable now over multiple processes. I removed previous workarounds.

@mafrahm Would you mind running your previous tests to confirm this fix?

Fixes #583.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -6,10 +6,9 @@ Helpers and utilities for working with columnar libraries.
 
 from __future__ import annotations
 
-# __all__ = []
+__all__ = []
 
 import os
-import gc
 import re
 import math
 import time
@@ -2674,6 +2673,7 @@ class DaskArrayReader(object):
         # backwards compatibility: when row group splitting is enbaled (the default), detect whether
         # the input file was written with support for that; a file does not support splitting in
         # case nested nodes separated by "*.list.element.*" (rather than "*.list.item.*") are found
+        # (to be removed in the future)
         if open_options.get("split_row_groups"):
             nodes = ak.ak_from_parquet.metadata(path)[0]
             cre = re.compile(r"^.+\.list\.element(|\..+)$")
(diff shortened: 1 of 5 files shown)


mafrahm:
Hi @riga, I had to merge this PR into the status from #566 for testing. However, there should be no technical differences regarding column processing.

First of all, I got errors when using my already existing outputs both in SelectEvents and in CreateHistograms. Therefore, I suppose I'd need to reprocess everything from scratch with this PR.

During SelectEvents, I don't see significant memory differences, however these differences are quite difficult to track.
During CreateHistograms, however, the previous implementation still performs better compared to this one

Screenshot 2024-12-11 at 09 07 19

riga:
Solved after private communication. Resource consumption is now down to a reasonable level when using multithreaded reading.

riga:
Thanks for checking :+1:
I added a last commit that removes manual garbage collection calls, plus a critical law update, i.e., nothing that changes the overall idea of this PR.
Merging now.

PR #587, opened 2024-12-13
Consistent units for htcondor_memory and htcondor_disk

riga:
This is a small PR that makes the units of the `htcondor_memory` and `htcondor_disk` settings consistent. They were MB and GB before, and now are both GB.

I also dropped the implicit conversion in the job file, since htcondor supports providing explicit units.


mafrahm (approved):

pkausw (approved):
LGTM, thanks!

PR #588, opened 2024-12-13
Various smaller CMS run 3 changes

riga:
This PR adds a couple smaller updates to adapt to changes in correctionlib files for run 3 (e.g. new electron SFs, jet veto map, etc.)

In particular, this PR:

- Adds a mechanism to the central jet veto map selector to account for additional veto map queries as required by the [CMS 2023BPix recipe](https://cms-jerc.web.cern.ch/Recommendations/#jet-veto-maps). For 2023BPix, the veto maps should be `&`'ed with the negation of a second map. This was rather straight forward, using the same corrector.
- Generalize `btag_weights` to configure a custom weight name
- Add jet phi to variables for JEC.
- Add electron phi to variables for SFs.
- Smaller "compactifications" of used/produced columns.

### DIFF
--- a/columnflow/production/cms/electron.py
+++ b/columnflow/production/cms/electron.py
@@ -168,8 +168,7 @@ def electron_weights_setup(
 
 @electron_weights.init
 def electron_weights_init(self: Producer, **kwargs) -> None:
-    weight_name = self.weight_name
-    self.produces |= {weight_name, f"{weight_name}_up", f"{weight_name}_down"}
+    self.produces.add(f"{self.weight_name}{{,_up,_down}}")
 
 
 # custom electron weight that runs trigger SFs
--- a/columnflow/production/cms/electron.py
+++ b/columnflow/production/cms/electron.py
@@ -46,7 +46,7 @@ class ElectronSFConfig:
 
 
 @producer(
-    uses={"Electron.{pt,eta,deltaEtaSC}"},
+    uses={"Electron.{pt,eta,phi,deltaEtaSC}"},
     # produces in the init
     # only run on mc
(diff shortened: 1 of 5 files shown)


columnflow/production/cms/electron.py
```diff
@@ -168,8 +168,7 @@ def electron_weights_setup(
 
 @electron_weights.init
 def electron_weights_init(self: Producer, **kwargs) -> None:
-    weight_name = self.weight_name
-    self.produces |= {weight_name, f"{weight_name}_up", f"{weight_name}_down"}
+    self.produces.add(f"{self.weight_name}{{,_up,_down}}")
```
mafrahm, line 171: I think it's great that we can add variables with this notation but it is not very verbose (especially the comma to indicate an empy string). Maybe we want to add comments to clarify that we add three columns to the uses here (nominal+up/down)?

mafrahm (approved):
I left one very minor comment, looks good to me otherwise.

Did you cross-check that the outputs are consistent with the previous implementation (e.g. by running a single `cf_diff` command)?

columnflow/production/cms/electron.py
riga: Indeed. I think it takes some time to get used to it. I added comments though to make life easier.

riga:
> I left one very minor comment, looks good to me otherwise.

Thanks for the swift review (this one is indeed a showstopper for us right now).

> Did you cross-check that the outputs are consistent with the previous implementation (e.g. by running a single `cf_diff` command)?

Yes, I run our entire chain for test samples and all looks good.

PR #590, opened 2024-12-18
Control workflow resources via law.cfg

riga:
This PR adds a feature to remote workflows that allows them to read resource values from the law.cfg file.

Example section in law.cfg:

```ini
[resources]
run3_2023__cf.CalibrateEvents: htcondor_memory=5GB, htcondor_disk=20GB
f.CalibrateEvents: htcondor_memory=2GB
```

The **keys** are interpreted identically to those in `[outputs] and [versions]`, that we use to pinpoint task output locations and default versions. Thus, the order of options matters as the resolution is performed in that order and stops once a match is found.

**Values** can be comma-separated values, with each one corresponding to parameter=value pairs. Parameters must be names of actual parameters, and their values are parsed using the underlying parameter instances.

In the scenario above, `cf.CalibrateEvents` tasks running with a config named `run3_2023` have increased htcondor memory on disk resources, while for all other configs, only a specific, lower memory constraint is in place.

As for versions, the value is only accepted by the task if no other value was specified by the user (e.g. via `CalibrateEvents.req(self, htcondor_memory=...)` in requirements or `--htcondor-memory ...` on the cli).

### DIFF
--- a/analysis_templates/cms_minimal/law.cfg
+++ b/analysis_templates/cms_minimal/law.cfg
@@ -54,8 +54,10 @@ default_selection_hists_optional: True
 skip_ensure_proxy: False
 
 # some remote workflow parameter defaults
+# (resources like memory and disk can also be set in [resources] with more granularity)
 htcondor_flavor: $CF_HTCONDOR_FLAVOR
 htcondor_share_software: False
+htcondor_memory: -1
 htcondor_disk: -1
 slurm_flavor: $CF_SLURM_FLAVOR
 slurm_partition: $CF_SLURM_PARTITION
@@ -119,6 +121,19 @@ lfn_sources: wlcg_fs_infn_redirector, wlcg_fs_global_redirector
 # example:
 ; run3_2023__cf.CalibrateEvents__nomin*: prod1
 ; cf.CalibrateEvents: prod2
+
+
+[resources]
+
+# default sources of remote workflows
+# keys can have the same format as described above in [versions] to pinpoint specific tasks
+# values should be comma-separated strings in the form "RESOURCE=VALUE", where RESOURCE should refer
+# to a valid task parameter (e.g. max_runtime, htcondor_memory, etc.) so that VALUE can be parsed
+# by the respective parameter instance at runtime
+# same as for [versions], the order of options is important as it defines the resolution order
+# example:
+; run3_2023__cf.CalibrateEvents__nomin*: htcondor_memory=5GB
+; run3_2023__cf.CalibrateEvents: htcondor_memory=2GB
 
 
 [job]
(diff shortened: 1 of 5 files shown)


columnflow/tasks/framework/base.py
```diff
@@ -219,6 +222,36 @@ def _get_cfg_versions_dict(cls):
...
+                        continue
+                    if "=" not in part:
+                        logger.warning_once(
+                            f"invalid_resource_{key}",
+                            f"resource for key {key} contains invalid instruction {part}, skipping",
+                        )
+                        continue
+                    param, value = (s.strip() for s in part.split("=", 1))
+                    params.append((param, value))
+                return key, params
```
mafrahm, line 243: this parse function is rather general (parsing into dictionary by splitting at "=" and ",") and independent of this particular function/class. Might this be reused at other points? Might it be better to separate it from the class?

analysis_templates/cms_minimal/law.cfg
```diff
@@ -121,6 +123,19 @@ lfn_sources: wlcg_fs_infn_redirector, wlcg_fs_global_redirector
...
+# default sources of remote workflows
+# keys can have the same format as described above in [versions] to pinpoint specific tasks
+# values should be comma-separated strings in the form "RESOURCE=VALUE", where RESOURCE should refer
+# to a valid task parameter (e.g. max_runtime, htcondor_memory, etc.) so that VALUE can be parsed
+# by the respective parameter instance at runtime
+# same as for [versions], the order of options is important as it defines the resolution order
+# example:
+; run3_2023__cf.CalibrateEvents__nomin*: htcondor_memory=5GB
+; run3_2023__cf.CalibrateEvents: htcondor_memory=2GB
+
```
mafrahm, line 137: Just for clarification, this should also be possible, right?

```suggestion
; run3_2023__cf.CalibrateEvents__nomin*: htcondor_memory=5GB
; run3_2023__cf.CalibrateEvents: htcondor_memory=2GB
; run3_2023__cf.MLTraining: htcondor_memory=4GB, htcondor_gpus=1

```

mafrahm (approved):
LGTM, just two minor inline comments

analysis_templates/cms_minimal/law.cfg
mafrahm: ah I could have just read your example in the PR description

columnflow/tasks/framework/base.py
riga: So far, there is no other place where this kind of parsing is used, but I agree we should move it once this approach can be used more generally.

analysis_templates/cms_minimal/law.cfg
riga: Yes, that's possible.

riga:
Thanks for reviewing!

PR #592, opened 2024-12-19
Add producer for recalculating softdrop mass

dsavoiu:
This PR adds a `Producer` for recalculating the softdrop mass of large-radius jets based on their subjets. This makes it possible to propagate any calibrations applied to the subjets (typically JEC/JER) to the softdrop mass (including systematic variations).

The input collections and output column for the recalcualted softdrop mass can be customized via class variables (defaults: `FatJet` for large-radius jets, `SubJet` for their subjets, `FatJet.msoftdrop` for the output).


riga (changes requested):
Thanks @dsavoiu !

It's nice to have these very generic, CMS-specific producers collected in one place and in the same format :+1:

Just one minor comment that I couldn't add as an inline-comment: Could you add this module to the `[analysis] production_modules` option in the analysis template?

Otherwise everything looks good to me.

dsavoiu:
Thanks for the quick review, @riga! I added the module to the `law.cfg`, as you suggested.

riga (approved):

PR #593, opened 2024-12-20
fix missing shift bins in plotting task

mafrahm:
This PR implements handling of missing shift bins in different configs/datasets corresponding to one process.
We can review+merge this directly into #566 and continue the review of the main branch, or we merge #566 first and create a separate PR afterwards.

The logic is as follows:
- per process, all shifts that are defined in one of the datasets corresponding to this processes are gathered via upstream task dependencies
- if a shift is requested for a process and part of the plot_shifts, but the shift is missing for a dataset corresponding to this process, we fill the missing shift bins with the content of the nominal bin

### DIFF
--- a/columnflow/tasks/plotting.py
+++ b/columnflow/tasks/plotting.py
@@ -190,7 +190,7 @@ class PlotVariablesBase(
                         }]
                         h = h[{"process": sum}]
 
-                        # create expexted shift bins and fill them with the nominal histogram
+                        # create expected shift bins and fill them with the nominal histogram
                         expected_shifts = set(plot_shifts) & process_shift_map[process_inst.name]
                         add_missing_shifts(h, expected_shifts, str_axis="shift", nominal_bin="nominal")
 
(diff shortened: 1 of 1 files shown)


columnflow/tasks/plotting.py
```diff
@@ -88,64 +155,58 @@ def run(self):
...
-                                for p in sub_process_insts[process_inst]
-                                if p.id in h.axes["process"]
+                                hist.loc(proc_id)
+                                for proc_id in process_info["dataset_procid_map"][dataset_inst]
+                                if proc_id in h.axes["process"]
                             ],
                         }]
                         h = h[{"process": sum}]
 
+                        # create expexted shift bins and fill them with the nominal histogram
```
pkausw, line 193:
```suggestion
                        # create expected shift bins and fill them with the nominal histogram
```

pkausw (commented):
LGTM. Let's wait with the merge until we have decided on the order of the PR resolution

PR #594, opened 2024-12-20
Utility functions and producers for generic delta-R matching

dsavoiu:
This PR adds functionality for performing $\Delta R$-based matching between arrays containing Lorentz vectors.

Three sets of functions are implemented:

* a set of utility functions in `production/util.py` for convenient operation on arrays of Lorentz vectors (LVs):
  - functions `lv_xyzt`, `lv_mass`, `lv_energy` for easy construction of behavior-aware LV arrays from any suitable *AwkwardArray*
  - functions `delta_r_match` and `delta_r_match_multiple` for performing $\Delta R$-based matching between arrays containing LVs and nested lists of LVs, respectively
* a base implementation in `production/matching.py` of a generic producer for $\Delta R$-based matching between physics object collections, taking the following parameters:
  - `src_name`: name of the input column to use as a source array for the matching
  - `dst_name`: name of the input column to use as a destination array for the matching
  - `output_idx_column`: name of the output column, which will be have the same structure as the *source* array and contain a list of indices that can be applied to the *destination* array to obtain the corresponding matched objects, or `-1` if no valid match was found
  - `max_dr`: the largest possible value of $\Delta R$ between source and destination objects that can be considered a valid match; if set to ``None``, no limit is imposed
* a series of derived producer classes in `production/cms/matching.py` with preset parameters for CMS-specific use cases


pkausw:
Why is this needed in the first place? delta R matching is directly implemented in the coffea behavior, does that have some specific short coming?

dsavoiu:
Hi @pkausw, *coffea* provides the behavior needed to compute the $\Delta R$ between object arrays via the `delta_r` and `metric_table` methods (and this PR relies on those), but there is no generic high-level functionality to actually perform the matching.

This implementation provides a convenient way to do this for arbitrary arrays of four-vectors, without needing to write the matching code from scratch or caring about coffea-specific things like behaviors.

It is sometimes also helpful to store the result of the matching in a persistent way, for example by storing the indices to the matched objects from another array, which can then be read in by later processing steps to retrieve the matched objects. This functionality is also provided in this PR both as a generic `Producer` and specializations for common matching tasks that can run on NanoAOD-like inputs.

columnflow/production/util.py
```diff
@@ -39,3 +42,151 @@ def attach_coffea_behavior(
...
+    # don't use `reduce` or list comprehensions
+    # to keep memory use as low as possible
+    tmp_lv_sum = None
+    for lv in lv_arrays:
+        if tmp_lv_sum is None:
+            tmp_lv_sum = lv
+        else:
+            tmp_lv_sum = tmp_lv_sum + lv
+
+    return tmp_lv_sum
```
riga, line 98: Is this function doing something identical to

```python
ak.concatenate([arr[..., None] * 1 for arr in lv_arrays], axis=-1).sum(axis=-1)
```
?

columnflow/production/util.py
```diff
@@ -39,3 +42,151 @@ def attach_coffea_behavior(
...
+    dst_lvs_filtered = ak.where(ak.is_none(dst_lvs_filtered, axis=0), [[]], dst_lvs_filtered)
+
+    # return either index or four-vector or best match
+    best_match = best_match_dst_idx if as_index else ak.firsts(dst_lvs[best_match_dst_idx])
+    return best_match, dst_lvs_filtered
+
+
+def delta_r_match_multiple(
+    dst_lvs: ak.Array,
+    src_lvs: ak.Array,
```
riga, line 152: Same as above :)

columnflow/production/matching.py
```diff
@@ -0,0 +1,97 @@
...
+def delta_r_matcher_init(self: Producer, **kwargs) -> None:
+    """
+    Dynamically add `uses` and `produces`
+    """
+    # input columns
+    self.uses |= {
+        f"{collection}.{var}"
+        for collection in (self.src_name, self.dst_name)
+        for var in ("pt", "eta", "phi", "mass")
+    }
```
riga, line 94:
```suggestion
    self.uses.add(f"{{{self.src_name},{self.dst_name}}}.{{pt,eta,phi,mass}}")
```

... although the mix of formatting and brace expansion isn't necessarily more readable. Up to you ;)

riga (changes requested):
Hey @dsavoiu ,

this is certainly helpful, especially when it comes to standard CMS-specific matching settings, like collection specific ∆R values.

I only have minor code / typehint comments below.

columnflow/production/util.py
riga, line 98: Just some code polishing.

```suggestion
    if not lv_arrays:
        raise ValueError("sum of empty input lorentz vector arrays impossible")
    return sum(lv_arrays[1:], lv_arrays[0])
```

columnflow/production/matching.py
dsavoiu: Thanks for the suggestion! I'm a bit torn on this due to the triple-braces tbh. It's super useful to have this auto-expansion feature when writing out multiple sub-columns manually, but when computing the column set programmatically, I think I ultimately would prefer to keep the loops explicit and the format string as short as possible, for readability reasons. Does that sound reasonable @riga?

columnflow/production/util.py
dsavoiu: This is a good suggestion and originally I actually had something similar, however, a big disadvantage is that it does not work if `lv_arrays` is a generator expression. We have this use case in the `mttbar` analysis where the entries in `lv_arrays` take up a large amount of memory, and after some profiling tests it turned out that looping over the generator explicitly and performing the addition in-place performed better in terms of the memory footprint. (The downside of this is unfortunately more verbose code.)

Despite the above, I actually noticed that this function is not needed for this PR. I copied it over from the `mttbar` analysis along with the others without checking to see if it was actually used. Given this, we can probably remove this from the PR, unless you think it would be useful to keep it in as a sort of "memory-efficient" implementation of LV addition?

columnflow/production/util.py
```diff
@@ -39,3 +42,151 @@ def attach_coffea_behavior(
...
+    return tmp_lv_sum
+
+
+#
+# functions for matching between collections of Lorentz vectors
+#
+
+def delta_r_match(
+    dst_lvs: ak.Array,
+    src_lv: ak.Array,
```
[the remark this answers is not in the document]
dsavoiu: Fair point! The reason I had for putting `dst` first was to serve as an _aide-mémoire_ about which array the matching result actually refers to. Since the definition of "source" and "destination" array is a bit arbitrary in this context, I often found myself wondering which was which myself while implementing this :D

But I agree this is subjective, and having `dst` before `src` could be confusing to users. Ultimately, the information contained in the docstring is hopefully clear enough, so that we can revert this to the more natural ordering.

columnflow/production/util.py
```diff
@@ -39,3 +42,151 @@ def attach_coffea_behavior(
...
+    Build an array containing only certain `fields` of an input array `arr`,
+    preserving behaviors.
+    """
+    # reattach behavior
+    if "behavior" not in kwargs:
+        kwargs["behavior"] = arr.behavior
+
+    return ak.zip(
+        {
+            field: getattr(arr, field)
```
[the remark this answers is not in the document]
dsavoiu: That's a good point. I didn't test this with nested fields since in probably all cases I would think to use this the fields would all be direct subfields of the object array. There might be a way to make this more generic, but I would probably leave this as-is until we have a concrete use case for that.

I'll add the typehints ;)

columnflow/production/matching.py
riga: Yes, fine by me :+1:

riga (approved):
Looks good! Let's merge it.

PR #597, opened 2025-01-09
Store all pdf weights in pdf_weights producer

riga:
This PR updates our CMS-related PDF weight producer with an option that allows it to store all (usually 100) weights per event.

Difference to the original `LHEPdfWeight`:

- Only the first 100 weights are stored, not 102 (which are usually alpha_s variations)
- The weights are stored normalized w.r.t. the nominal weight.

---

I also fixed a small bug in the determination of the combined variations (using only the first 100, instead of 102 weights), and updated the outlier log.

### DIFF
--- a/columnflow/production/cms/pdf.py
+++ b/columnflow/production/cms/pdf.py
@@ -4,13 +4,15 @@
 Column production methods related to the PDF weights.
 """
 
+from __future__ import annotations
+
 import functools
 
 import law
 
 from columnflow.production import Producer, producer
 from columnflow.util import maybe_import
-from columnflow.columnar_util import set_ak_column, full_like
+from columnflow.columnar_util import set_ak_column, full_like, fill_at
 
 np = maybe_import("numpy")
 ak = maybe_import("awkward")
@@ -18,8 +20,9 @@ ak = maybe_import("awkward")
 
 logger = law.logger.get_logger(__name__)
 
-# helper
+# helpers
 set_ak_column_f32 = functools.partial(set_ak_column, value_type=np.float32)
+fill_at_f32 = functools.partial(fill_at, value_type=np.float32)
 
 
 @producer(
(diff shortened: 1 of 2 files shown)


columnflow/production/cms/pdf.py
```diff
@@ -108,27 +119,36 @@ def pdf_weights(
             "weight is already included in the LHEWeight.",
         )
 
-    # normalize all weights by the nominal one, assumed to be the first value
-    pdf_weights = events.LHEPdfWeight[~bad_mask, 1:] / pdf_weight_nominal
+    # normalize all 100 weights by the nominal one, assumed to be the first value
+    # (for datasets with 103 weights, the last two weights are related to alpha_s variations)
+    pdf_weights = ak.without_parameters(events.LHEPdfWeight[~bad_mask, 1:101]) / pdf_weight_nominal
```
mafrahm, line 124: not really part of this PR, but I think that applying this `~bad_mask` does not work as intended, because as soon as some events are masked, the shape of the `pdf_weights` changes and we'll get an error message when trying to add this column to the events (similar issue is also in `scale.py`).
However, I think we want to have the option to raise an exception when there is an invalid number of pdf (scale) weights and maybe provide an option how to handle such events (e.g. setting nominal/up/down to 1)

mafrahm (approved):
LGTM

columnflow/production/cms/pdf.py
riga: Good point. Let's include this here though.

riga: Added.

columnflow/production/cms/pdf.py
```diff
@@ -61,44 +78,48 @@ def pdf_weights(
...
-            f"unknown outlier_action '{outlier_action}', "
-            f"known values are {','.join(known_actions)}",
-        )
-    known_log_modes = ("none", "info", "debug", "warning")
-    if outlier_log_mode not in known_log_modes:
-        raise ValueError(
-            f"unknown outlier_log_mode '{outlier_log_mode}', "
-            f"known values are {','.join(known_log_modes)}",
-        )
+    _raise_unknown_action("invalid_weights_action", invalid_weights_action, ("ignore", "raise"))
```
mafrahm, line 81: where is this coming from?

riga: It's at the end of the file.

riga:
Thanks for the reviews!

PR #598, opened 2025-01-13
Prefer true / false over 1 / 0 in shell scripts.

riga:
This PR affects only setup scripts. So far, we *sometimes* used integers 0 and 1 for flags that should actually be stored as real booleans. This lead to ambiguities in cases where an actual boolean value was expected, e.g. when using a variable in the law.cfg for an option that expects "true" or "false" and that cannot interpret "1".

This PR changes that in a backwards compatible way. All flags that are expected throughout cf are converted at the earliest instance possible.

Things not only become consistent but also far more readable and expressive.


pkausw (approved):
LGTM, I agree that this improves readability by alot!

mafrahm (approved):

PR #599, opened 2025-01-13
Remove forest merge interface.

riga:
This PR adjusts `MergeReducedEvents` and `MergeSelectionStats`. They were inheriting from the `ForestMerge` interface which allowed them to perform all `N -> M` merges in a cascaded way.

However, this is barely necessary as the merging of json (parquet) files usually takes less than 2s (10-15s) per `n -> 1` step and causes overly large task trees that accumulate runtimes exceeding that of the actual task processing.

I removed the interface and made these two tasks normal merging workflows. Now, even with a large dataset, the merging is faster and even beats the tree building time with the forest merge interface.

### DIFF
--- a/columnflow/tasks/reduction.py
+++ b/columnflow/tasks/reduction.py
@@ -416,7 +416,7 @@ class MergeReducedEvents(
     SelectorStepsMixin,
     CalibratorsMixin,
     DatasetTask,
-    law.tasks.ForestMerge,
+    law.LocalWorkflow,
     RemoteWorkflow,
 ):
 
@@ -427,8 +427,6 @@ class MergeReducedEvents(
         f"removed after successful merging; default: {default_keep_reduced_events}",
     )
 
-    max_merge_factor = 50
-
     sandbox = dev_sandbox(law.config.get("analysis", "default_columnar_sandbox"))
 
     # upstream requirements
@@ -437,22 +435,6 @@ class MergeReducedEvents(
         MergeReductionStats=MergeReductionStats,
         ReduceEvents=ReduceEvents,
     )
-
-    @property
-    def merge_factor(self) -> int:
-        """
-        Defines the number of inputs to be merged per output at any point in the merging forest.
-        Required by law.tasks.ForestMerge.
-        """
-        # return as many inputs as leafs are present to create the output of this tree, capped at 50
-        return min(self.file_merging, self.max_merge_factor)
-
-    def is_sandboxed(self):
-        # when the task is a merge forest, consider it sandboxed
-        if self.is_forest():
-            return True
-
-        return super().is_sandboxed()
 
     @law.workflow_property(setter=True, cache=True, empty_value=0)
     def file_merging(self):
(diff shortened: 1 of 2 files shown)


columnflow/tasks/reduction.py
```diff
@@ -467,54 +449,40 @@ def workflow_condition(self):
...
 
-    def merge_workflow_requires(self):
-        return {
-            "stats": self.reqs.MergeReductionStats.req_different_branching(self),
-            "events": self.reqs.ReduceEvents.req_different_branching(self, branches=((0, -1),)),
-        }
+    def workflow_requires(self):
+        reqs = super().workflow_requires()
+        reqs["stats"] = self.reqs.MergeReductionStats.req_different_branching(self)
+        reqs["events"] = self.reqs.ReduceEvents.req_different_branching(self)
```
pkausw, line 458: Are we sure that we don't need the `branches` parameters here? I seem to remember that there was a minor issue with the resolution of specific branches - I think @mafrahm found that some time ago, maybe he can comment

pkausw (changes requested):
Good idea in principle, just have a couple of questions

columnflow/tasks/reduction.py
```diff
@@ -467,54 +449,40 @@ def workflow_condition(self):
...
-    def merge_output(self):
-        # use the branch_map defined in DatasetTask to compute the number of files after merging
-        n_merged = len(DatasetTask.create_branch_map(self))
-        return law.SiblingFileCollection([
-            {"events": self.target(f"events_{i}.parquet")}
-            for i in range(n_merged)
-        ])
+        return {
+            "events": self.target(f"events_{self.branch}.parquet"),
+        }
```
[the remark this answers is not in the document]
riga: The creation of a SiblingFileCollection is done by anyways by the workflow, wrapping all outputs of its branches. The SFC we had before on branch level didn't make too much sense anyway.

columnflow/tasks/reduction.py
riga: That's what the `req_different_branching` is actually doing. It prevents parameters from being passed that would imply an equal branch granularity.

columnflow/tasks/selection.py
```diff
@@ -295,57 +295,46 @@ class MergeSelectionStats(
...
 
-    def merge_requires(self, start_branch, end_branch):
-        return self.reqs.SelectEvents.req_different_branching(
-            self,
-            branches=((start_branch, end_branch),),
-            workflow="local",
-            _exclude={"branch"},
-        )
+    def requires(self):
+        return self.reqs.SelectEvents.req_different_branching(self, workflow="local", branch=-1)
```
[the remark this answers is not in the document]
riga: `requires` is on branch-level already, so everything but "local" could cause jobs submitting jobs. In `workflow_requires`, the workflow type is properly propagated.

columnflow/tasks/reduction.py
mafrahm: The issue @pkausw mentions was solved with this PR:  https://github.com/columnflow/columnflow/pull/544
and as far as I remember, setting this `branches` parameter was necessary to not submit certain jobs

riga: Indeed, this was required for the `merge_workflow_requires`, since this got called multiple times by workflows in different tree heights (in the "old" forest merge approach). This should not apply anymore with the forest merge removed :+1:

Thanks for the check!

pkausw (approved):

mafrahm:
Hi @riga @pkausw, I just tested the changes from this PR and saw that the `cf.MergeReducedEvents` only merges N-1 files per branch when I have a `merge_factor` N, meaning that some files will be omitted during the merging. I also checked with one dataset where all files are supposed to be merged into one and the selected events from the last branch are missing. Can you reproduce this issue?

columnflow/tasks/reduction.py
```diff
@@ -467,54 +449,40 @@ def workflow_condition(self):
...
 
-    def merge_requires(self, start_branch, end_branch):
+    def requires(self):
         return {
             "stats": self.reqs.MergeReductionStats.req_different_branching(self, branch=0),
             "events": self.reqs.ReduceEvents.req_different_branching(
                 self,
-                branches=((start_branch, end_branch),),
                 workflow="local",
+                branches=((min(self.branch_data), max(self.branch_data)),),
```
mafrahm, line 467: I think this needs to be `max(self.branch_data) + 1`

PR #601, opened 2025-01-13
Make attach_coffea_behavior functional.

riga:
This PR adds three changes:

- Separate functional part of the `attach_coffea_behavior` into a function in our columnar utils. This way, it can be called as a producer (as before) and also in different contexts.
- Enables the MET and PuppiMET behaviors again, as the reason why it was disabled is obsolete.
- Calls `attach_coffea_behavior` right before histogramming so that functional variable expressions can make use of it.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -47,14 +47,58 @@ hist = maybe_import("hist")
 logger = law.logger.get_logger(__name__)
 logger_perf = law.logger.get_logger(f"{__name__}-perf")
 
-#: Columns that are always required when opening a nano file with coffea.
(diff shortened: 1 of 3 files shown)


mafrahm (approved):
Nice! Just one minor question, is there some simple/convenient way to overwrite the `default_coffea_collections` from user side to allow including behavior of custom collections in histogramming?

riga:
Not yet. At the moment it only adds what we know exists in the standard nano scheme. However, I think we should make this (as well as what we call "default") somehow configurable in the future as this is not only analysis- but also experiment-specific.

PR #606, opened 2025-01-20
Remove ForestMerge from MergeMLStats

riga:
This PR removes the ForestMerge interface from MergeMLStats.

The complexity of the cascaded merging is unnecessary for the merging of (essentially) json files and removing it actually speeds up the entire process.

### DIFF
--- a/columnflow/tasks/ml.py
+++ b/columnflow/tasks/ml.py
@@ -303,9 +303,6 @@ class MergeMLStats(
     DatasetTask,
     law.tasks.ForestMerge,
 ):
-    # recursively merge 20 files into one
-    merge_factor = 20
-
     # skip receiving some parameters via req
     exclude_params_req_get = {"workflow"}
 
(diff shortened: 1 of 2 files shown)


columnflow/tasks/ml.py
```diff
@@ -303,9 +303,6 @@ class MergeMLStats(
     DatasetTask,
     law.tasks.ForestMerge,
 ):
-    # recursively merge 20 files into one
-    merge_factor = 20
-
     # skip receiving some parameters via req
     exclude_params_req_get = {"workflow"}
```
mafrahm, line 307: should we also remove this attribute?
```suggestion
```

columnflow/tasks/ml.py
```diff
@@ -303,9 +303,6 @@ class MergeMLStats(
     DatasetTask,
     law.tasks.ForestMerge,
```
mafrahm, line 304:
```suggestion
    law.LocalWorkflow,
    RemoteWorkflow,
```

mafrahm (changes requested):
two small but important suggestions. LGTM otherwise

riga:
Thanks for the review @mafrahm , I added both suggestions.

mafrahm (approved):

PR #607, opened 2025-01-20
Flip stack plot order.

riga:
This PR adjust the order of processes in stack plots.

As per recommendation, the process that is lowest in the legend should also be lowest in the stack plot. This was not the case yet but comes with this PR. To revert to the old behavior, there is a `flip_stack` option now.

~~Would you need this legacy option @mafrahm @dsavoiu ? If not, we can also drop it and remove the code necessary for flipping legend handles.~~

---

This PR also includes a couple smaller plot updates, mostly on the configurability of the `grid_spec` (of the axes) itself, plus some code cleanup.


riga:
After discussing privately with @dsavoiu it is clear now that we don't need to `flip_stack` option. I removed it in 7443290.

PR #608, opened 2025-01-20
Option to skip selection hists

riga:
The `SelectionEvents` tasks creates multiple outputs:

- selection masks
- additional columns
- json files with statistics
- histograms with statistics

The last one is optional, however, it's always created.

This PR adds an option to disable the writing of histogram files altogether. Independent of whether an analysis needs them, they should only be created if desired. Otherwise, all status checks in a larger workflow will include these O(10k) files, which can be skipped.

### DIFF
--- a/analysis_templates/cms_minimal/law.cfg
+++ b/analysis_templates/cms_minimal/law.cfg
@@ -49,8 +49,8 @@ default_keep_reduced_events: True
 # slightly to the left to avoid them being excluded from the last bin; None leads to automatic mode
 default_histogram_last_edge_inclusive: None
 
-# boolean flag that, if True, sets the *hists* output of cf.SelectEvents and cf.MergeSelectionStats to optional
-default_selection_hists_optional: True
+# boolean flag that, if True, configures cf.SelectEvents to create statistics histograms
+default_create_selection_hists: False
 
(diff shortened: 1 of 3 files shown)


analysis_templates/cms_minimal/law.cfg
```diff
@@ -49,8 +49,8 @@ default_keep_reduced_events: True
 # slightly to the left to avoid them being excluded from the last bin; None leads to automatic mode
 default_histogram_last_edge_inclusive: None
 
-# boolean flag that, if True, sets the *hists* output of cf.SelectEvents and cf.MergeSelectionStats to optional
-default_selection_hists_optional: True
+# boolean flag that, if True, configures cf.SelectEvents to create statistics histograms
+default_create_selection_hists: False
```
mafrahm, line 53: I do not like setting this parameter to False per default, because this means that histograms will suddenly not be produced anymore, even if users might already be expected to have them produced. Not sure, how many users already use them, if my analysis is the only one, then I would not mind

riga: Right, the default should be True :+1:

riga:
Thanks for the review @mafrahm , I changed the default.

mafrahm (approved):
perfect, you also identified the correct points where to change the defaults

PR #609, opened 2025-01-20
Fix task inheritance.

riga:
This PR fixes the inheritance order of various classes, which in its current state, leads to a wrong parameter resolution order (e.g. shifts other than nominal are allowed for data).

The problem can be described with the `CalibrateEvents` task.

Currently, the inheritance order is

```python
class CalibrateEvents(
    CalibratorMixin,
    ChunkedIOMixin,
    DatasetTask,
    law.LocalWorkflow,
    RemoteWorkflow,
)
```

The "main" task (not a mixin) is DatasetTask. Since it is not the first class in this order, **and** in inherited methods, most tasks do their `super()...` call in the beginning, this leads to a situation where `CalibratorMixin` is evaluated before `DatasetTask`, even though it's potentially more specialized (since it's a mixin).

The actual order should be

```python
class CalibrateEvents(
    DatasetTask,
    ChunkedIOMixin,
    CalibratorMixin,
    law.LocalWorkflow,
    RemoteWorkflow,
)
```

Now, from top to bottom, mixins become more specialized and resolution orders are consistent.

Another example is `ReduceEvents`, whose order is now

```python
class ReduceEvents(
    DatasetTask,
    ChunkedIOMixin,
    CalibratorsMixin,
    SelectorStepsMixin,
    law.LocalWorkflow,
    RemoteWorkflow,
)
```

Again, mixins become more specialized from top to bottom.

To cope with the wrong order, but still end up with correct output paths, the current code used a trick in `store_parts` (always adding new fragment before the version parameter when existing). I adjusted all `store_parts` to make paths consistent and backwards compatible (adding new fragments behind the `config` entry, since all mixins are (at least for now) `ConfigTask`'s; for the multi-config approach, this would only need be be changed to (e.g.) task_family).


columnflow/tasks/framework/mixins.py
```diff
@@ -201,19 +201,12 @@ def calibrator_repr(self):
...
-        This method calls :py:meth:`store_parts` of the ``super`` class and inserts
-        `{"calibrator": "calib__{self.calibrator}"}` before keyword ``version``.
-        For more information, see e.g. :py:meth:`~columnflow.tasks.framework.base.ConfigTask.store_parts`.
-
-        :return: Updated parts to create output path to store intermediary results.
+        :return: Dictionary with parts that will be translated into an output directory path.
         """
         parts = super().store_parts()
-        parts.insert_before("version", "calibrator", f"calib__{self.calibrator_repr}")
+        parts.insert_after("config", "calibrator", f"calib__{self.calibrator_repr}")
```
mafrahm, line 209: Can you add the "config" store_parts name as classatribute to the `ConfigTask`? In that way, extending to the `MultiConfigTask` will get much easier
```suggestion
        parts.insert_after(self.configs_part, "calibrator", f"calib__{self.calibrator_repr}")
```

mafrahm (changes requested):
One inline comment. Before merging, I'd like to check if changing the task inheritance changes any behavior in my analysis.

columnflow/tasks/framework/mixins.py
riga: Not a bad idea in general. But would analyses typically change this value, or is it rather inheriting tasks that should overwrite it?

Also, I'm trying to see if there are already situations where this value is defined dynamically. If so, it would make things a bit complicated.

riga:
>  Before merging, I'd like to check if changing the task inheritance changes any behavior in my analysis.

Indeed, and highly appreciated. This is quite a core change and knowing how dynamic your analysis code is, a second test would be mandatory I think.

columnflow/tasks/framework/mixins.py
mafrahm: it's just to make the mixins depend on the choice of `ConfigTask` vs `MultiConfigTask`, so I would not expect to change this from analysis side


(we could also just name the part "config" (or "configs") independently of single config vs multi config)

riga: Added in 5c6a1482.

In some custom tasks this was not generically possible, but these are perfect examples of when the inherited behavior has to be overwritten.

columnflow/tasks/framework/base.py
```diff
@@ -990,6 +990,9 @@ class ConfigTask(AnalysisTask):
         description=f"name of the analysis config to use; default: '{default_config}'",
     )
 
+    # the field in the store parts behind which the new part is inserted
+    store_part_anchor = "task_family"
```
mafrahm, line 994: mhh, this would have been the point where I'd differentiate between "config" and "configs" (In `MultiConfigTask`), but now there is no unique definition of the `store_part_anchor. I'd prefer to keep this simple and only distinguish between "config" and "configs" using the classattribute

Also how does it work with class inheritance and having multiple definitions of `store_part_anchor`? Do we not just end up defining it as whatever is set by the first class in our inheritance? Is that what is intended?

riga:
Closed in favor of https://github.com/columnflow/columnflow/pull/628.

PR #610, opened 2025-01-23
Energy calibrations for electrons and photons

pkausw:
This PR implements the energy scale and resolution calibrations for electrons and photons. The procedure follows the recommendations as described in the [EGamma TWiki](https://twiki.cern.ch/twiki/bin/view/CMS/EgammSFandSSRun3#Scale_And_Smearings_Example). The methodology is the same for both electrons and photons, which is why there is a base class for the scale and resolution calibrations, respectively. The calibrations for the individual objects are derived from the base classes.

Since the resolution calibration is also based on random numbers, it is advisable to implement a procedure that uses deterministic seeds. This implementation is similar to the implementation for JER. Since this requires object-specific seeds, the `deterministic_jet_seeds` class is first abstracted to generate seeds for arbitrary objects. The object-specific modules are then derived from the generic one.

Additionally, this PR includes new producers to calculate the supercluster eta for photons (for nanoAOD version < 14) and electrons, which are necessary for the EGamma calibrations.

Note that unittests and the documentation strings will be updated in a separate PR to make these new modules usable on a shorter time scale.


columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+class egamma_scale_corrector(Calibrator):
+
+    with_uncertainties = True
+    """Switch to control whether uncertainties are calculated."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @property
+    @abc.abstractmethod
```
riga, line 47: There's also `abc.abstractproperty` for this. Also applies to a couple places below (e.g. in egamma_resolution_corrector).

columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+    ) -> ak.Array:
+        """
+        Calibrator for photon and electron energy scales.
+        Requires an external file in the config under ``electronSS`` or ``photon_SS``,
+        e.g.
+
+        .. code-block:: python
+
+            cfg.x.external_files = DotDict.wrap({
+                "tau_sf": "/afs/cern.ch/work/m/mrieger/public/mirrors/jsonpog-integration-6ce37404/POG/TAU/2017_UL/tau.json.gz",  # noqa
```
riga, line 79: still referring to tau sfs.

columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+        """
+        ...
+
+    @abc.abstractmethod
+    def get_scale_config(self) -> EGammaCorrectionConfig:
+        """Function to retrieve the configuration for the photon energy correction."""
+        ...
+
+    def call_func(
+        self: Calibrator,
```
riga, line 67: Same comment as for the seed producers below, we usually don't need to have type hints for `self` (same below)

columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+            # nano columns
+            f"{self.source_field}.{{seedGain,pt,superclusterEta,r9}}",
+            "run",
+            optional_column(f"{self.source_field}.rawPt"),
+        }
+        self.produces |= {
+            f"{self.source_field}.pt",
+            optional_column(f"{self.source_field}.rawPt"),
+        }
+        self.scale_config = self.get_scale_config()
```
riga, line 181: Can you move this assignment to the setup()? I guess it's only needed within the run() function.

columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+        # photon scale _uncertainties_ are only available for MC
+        if self.with_uncertainties and self.dataset_inst.is_mc:
+            src_fields = [f"{self.source_field}.pt"]
+            self.produces |= {
+                f"{field}_scale_{direction}"
+                for field, direction in itertools.product(
+                    src_fields,
+                    ["up", "down"],
+                )
+            }
```
riga, line 197:
```suggestion
            self.produces.add(f"{self.source_field}.pt_scale_{{up,down}}")
```

columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+        """Function to retrieve the configuration for the photon energy correction."""
+        ...
+
+    def call_func(
+        self: Calibrator,
+        events: ak.Array,
+        **kwargs,
+    ) -> ak.Array:
+        """
+        Calibrator for tau energy. Requires an external file in the config under ``tau_sf``, e.g.
```
riga, line 262: Docs still tau related.

columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+            # nano columns
+            f"{self.source_field}.{{pt,superclusterEta,r9}}",
+            optional_column(f"{self.source_field}.rawPt"),
+        }
+        self.produces |= {
+            f"{self.source_field}.pt",
+            optional_column(f"{self.source_field}.rawPt"),
+        }
+
+        self.resolution_config = self.get_resolution_config()
```
riga, line 390: Can be moved to setup().

columnflow/calibration/cms/egamma.py
```diff
@@ -0,0 +1,581 @@
...
+            # also check if met propagation is enabled
+            src_fields = [f"{self.source_field}.pt"]
+
+            self.produces |= {
+                f"{field}_res_{direction}"
+                for field, direction in itertools.product(
+                    src_fields,
+                    ["up", "down"],
+                )
+            }
```
riga, line 404:
```suggestion
            self.produces.add(f"{self.source_field}.pt_res_{{up,down}}")
```

riga (changes requested):
Indeed a very important addition to our cms-specific recipes, thanks!

Besides a couple inline comments below, I have one general question: Are changes to electrons / photons usually propagated to MET? Also adding @dsavoiu here.

columnflow/production/cms/supercluster_eta.py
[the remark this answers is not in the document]
pkausw: I would say no - I removed the module for photons for now, but I would still argue that it's better to keep the one for electrons for modularity. Open for discussion though

columnflow/production/cms/seeds.py
```diff
@@ -172,67 +173,110 @@ def apply_route(ak_array: ak.Array, route: Route) -> ak.Array | None:
...
+        ...
 
-    # uniqueness test across all jets in the chunk for debugging
-    # n_jets = ak.sum(ak.num(events.Jet, axis=1))
-    # n_seeds = len(set(np_jet_seed))
-    # match_text = "yes" if n_jets == n_seeds else "NO !!!"
-    # print(f"jets: {n_jets}, unique seeds: {n_seeds}, match: {match_text}")
+    def call_func(self: Producer, events: ak.Array, **kwargs) -> ak.Array:  # type: ignore
+        """
+        Produces deterministic seeds for each jet and stores them in *events* which is also returned.
```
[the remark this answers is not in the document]
pkausw: I thought we decided to postpone the update of the docstrings into another PR... I'll do it now

columnflow/production/cms/supercluster_eta.py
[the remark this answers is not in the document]
riga: Ah right, I was neglecting the electron sceta.

columnflow/calibration/cms/egamma.py
pkausw: Good catch, I copied this from the jets calibrators, who actually need the configs in their init. Fixed :+1:

PR #613, opened 2025-01-28
correct path only_missing looks for in MergeHistograms

nprouvost:
MergeHistograms fails as of now when using the only_missing argument due to an error in the lookup in the dictionary structure. This PR fixes that.


pkausw (approved):
LGTM, thanks @nprouvost !

PR #614, opened 2025-01-29
Hotfix: properly forward and use `deps_kwargs`.

dsavoiu:
This PR introduces a hotfix for the proper handling of `deps_kwargs`. There are three main changes, addressing distinct but related issues:

1. The `instance_cache` used by `ArrayFunction.create_dependencies` uses only the dependency class as the key, but does not consider any class attributes, which may be overridden by user-defined `init` functions. The dependency instance is thus not recreated by the second call to `ArrayFunction.create_dependencies`. Similarly, the `only_update` flag causes the recreation to be skipped if the dependency already exists in `self.deps`. The result of this is that dependencies are always initialized with their default `kwargs`, and any `deps_kwargs` supplied via the user-defined `init` functions are ignored. While a full fix would require significant refactoring, this PR introduces a workaround by fully recreating the dependencies, bypassing the `instance_cache` and `only_update` mechanisms.
2. The `ArrayFunction.__init__` method ignores any user-defined `**kwargs` passed to it, including those set via `deps_kwargs`. This PR adds `setattr` calls to propagate these to the resulting object as instance attributes.
3. The `deps_kwargs` are initialized with a `DotDict()`, but the dependency classes are not automatically added as keys (as done for the dependency lookup via `__getitem__`), leading to unexpected `KeyErrors` in user-defined `init` functions. This PR uses a `defaultdict(dict)` for the initialization as a workaround.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -19,7 +19,7 @@ import threading
 import multiprocessing
 import multiprocessing.pool
 from functools import partial
-from collections import namedtuple, OrderedDict, deque
+from collections import namedtuple, OrderedDict, deque, defaultdict
 
 import law
 import order as od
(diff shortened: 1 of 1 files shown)


riga:
Looks good to me!

In any case, the refactoring that @mafrahm and I have in mind should likely also clean up the task array function initialization.

riga (approved):

PR #615, opened 2025-02-06
Update input files lepton SFs

nprouvost:
update lepton SFs producers such that json files can also be used as inputs for the correctionlib


riga (approved):
:+1:

Issue #616, opened 2025-02-06
priority style_configs default plotting functions


maadcoen:
I wrote my own plotting task, in which I reused a columnflow plotting function. However, I wanted to change the default style configuration to change some labeling. So I wrote something in my plotting code like thi

```python
fig, _ = self.call_plot_func(
    self.plot_function,
    ...
    style_config=my_default_style_config,
    **self.get_plot_parameters(),
)
```

This worked, but I noticed that afterwards, I could not change the specifications in  `my_default_style_config` via the command line argument `--custom-style-config`. I found out that this is happens because of how the following function is coded:

```python
def update_plot_kwargs(self, kwargs: dict) -> dict:
    """
    Hook to update keyword arguments *kwargs* used for plotting in :py:meth:`call_plot_func`.
    """
    ...

    # update style_config
    style_config = kwargs.get("style_config", {})
    if isinstance(custom_style_config, dict) and isinstance(style_config, dict):
        style_config = law.util.merge_dicts(custom_style_config, style_config)
        kwargs["style_config"] = style_config
    # update defaults after application of general_settings
    kwargs.setdefault("cms_label", "pw")
    
    ...

    return kwargs
```

so basically, the line `law.util.merge_dicts(custom_style_config, style_config)` overwrites `custom_style_config` with the options given in `style_config`, which in my case would be `my_default_style_config`. To me this seems undesirable, so I wondered if it wouldn't make more sense to reverse this to `law.util.merge_dicts(style_config, custom_style_config)` such that style options specified via the command line always have precedence.

mafrahm:
Hi @maadcoen, I am not sure if there was a reason to implement the style resolving in that order. Right now, within columnflow tasks/plot functions, the order of style resolving seems to be:
task style > custom_style_config > default_custom_style_config > plot-function style

Your proposal is to change (by switching the arguments in the `merge_dicts`call) the order to:
custom_style_config > default_custom_style_config > task style > plot-function style

Is that correct? I feel like your proposed order is preferable because it gives more control to the user and changing this should not break anything in the existing plot functions / tasks.

maadcoen:
That is indeed my proposal! I made a pull request to do the change

mafrahm:
thank you! closed via #617

PR #617, opened 2025-02-06
prioritise custom style config via command line

maadcoen:
Style configs hard coded in the tasks were given priority over command line specified style configs (defined via the custom style config groups). Reversed this situation to give more control via command line.

see also issue #616


mafrahm (approved):
LGTM, thank you for opening the PR!

PR #620, opened 2025-02-10
Correct input keys for electron Efficiencies

nprouvost:
add correct keys for Efficiencies correction sets for electrons


riga (approved):
Thanks, and good catch!

PR #622, opened 2025-02-11
adding dy weights producer

aalvesan:
This merge adds:

- a new producer `gen_dilepton` that reconstructs the (visible) di-lepton pair from generator level info. It also provides the four-momenta of all leptons (including neutrinos) from the hard process. It produces `gen_dilepton_pt`, `gen_dilepton_vis.{pt,phi}`, `gen_dilepton_all.{pt,phi}"`.
- a new producer `dy_weights` that creates columns with dy weights per systematic shift. The number os up/down shifts if dependent on the order of the dy datasets and is calculated on the fly. 
- a new producer `recoil_corrections`
- a hotfix for coffea behaviour.


columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,189 @@
...
+    # electrons and muons need to have status == 1, so find them
+    ele_mu_mask = (pdg_id == 11) | (pdg_id == 13)
+    final_ele_mu_mask = ele_mu_mask & (status == 1)
+
+    # taus need to have status == 2,
+    tau_mask = pdg_id == 15
+    final_tau_mask = tau_mask & (status == 2)
+
+    # combine the masks
+    lepton_mask = final_ele_mu_mask | final_tau_mask
```
riga, line 63: I think we can just combine the masks without storing partial ones.

```suggestion
    # electrons and muons need to have status 1, taus status 2
    lepton_mask = (
        # e, mu
        (((pdg_id == 11) | (pdg_id == 13)) & (status == 1)) |
        # tau
        ((pdg_id == 15) & (status == 2))
    )

```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,189 @@
...
+from columnflow.columnar_util import set_ak_column
+
+np = maybe_import("numpy")
+ak = maybe_import("awkward")
+vector = maybe_import("vector")
+
+
+# helper
+set_ak_column_f32 = functools.partial(set_ak_column, value_type=np.float32)
+
```
riga, line 25: Not used.

```suggestion
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,189 @@
+# coding: utf-8
+
+"""
+Column production methods related to Drell-Yan reweighting.
+"""
+
+from __future__ import annotations
+import functools
```
riga, line 8: Not used.

```suggestion
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # fill the mask with False if it is None and extract the gen leptons
+    mask_gen = ak.fill_none(mask_gen, False)
+    lepton_pairs = events.GenPart[mask_gen]
+
+    # some up the four momenta of the leptons
+    lepton_pair_momenta = lepton_pairs.sum(axis=-1)
+
+    # finally, save the pt of the lepton pair on generator level
+    events = set_ak_column(events, "gen_dilepton.pt", lepton_pair_momenta.pt)
```
riga, line 75: Since we only store one feature here, we should perhaps just call it "gen_dilepton_pt" to avoid creating a nested structure. This would require a corresponding change above and below.

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,189 @@
...
+    return events
+
+
+@producer(
+    uses={get_gen_dilepton.PRODUCES},
```
riga, line 80: Since the gen lepton producer is running during the selection, and the weight producer most likely during the ProduceColumns stage later, there shouldn't be a direct dependence, but we can just add the requested column name here.

```suggestion
    uses={"gen_dilepton_pt"},
```

riga (changes requested):
Hey @aalvesan ! Thanks, this looks very good. I only had a small number of inline code comments, nothing structural.

columnflow/production/cms/dy.py
philippgadow: I plan to modify this in a future PR to also store the phi of the dilepton, so I think the nested structure makes sense.

riga: Ok :+1:

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,501 @@
...
+    if dy_n_unc != self.n_unc:
+        raise ValueError(
+            f"Expected {self.n_unc} uncertainties, got {dy_n_unc}",
+        )
+
+
+#
+# Helper functions for kinematics.
+#
+def GetXYfromPTPHI(pt, phi):
```
riga, line 254: I just wanted to start reviewing this, but I stumbled over the four-vector logic here and below. Even though they look good actually, I think we should stick to the existing methods available through the nano behavior. Or was there an issue with the behavior we should solve?

philippgadow: I'll change these today. This was my first contact point with implemeting columnar analysis. Now I have some more experience from BTV workflows in coffea and can update these to nano behaviour.

philippgadow: I could not push to this branch (on the columnflow project), so I made a [MR targeting this branch](https://github.com/columnflow/columnflow/compare/tmp_dy...philippgadow:columnflow:tmp_dy?expand=1) on my project.

philippgadow: I now pushed directly to this branch.

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,458 @@
...
+"""
+
+from __future__ import annotations
+
+import law
+
+from dataclasses import dataclass
+
+from columnflow.production import Producer, producer
+from columnflow.util import maybe_import, InsertableDict
```
mafrahm, line 14: Just a single comment because I just tested this: the `columnflow.util.InsertableDict` has been removed at some point, apparently
```suggestion
from columnflow.util import InsertableDict
from law.util import InsertableDict
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+                "Incomplete dy_weight_config: missing era, order, correction or unc_correction.",
+            )  # noqa
+
+
+@producer(
+    uses={"GenPart.*"},
+    produces={
+        "gen_dilepton_pt",
+        "gen_dilepton_vis.{pt,phi}",
+        "gen_dilepton_all.{pt,phi}",
```
riga, line 49: Just to visually simplify this a bit.

```suggestion
        "gen_dilepton_pt", "gen_dilepton_{vis,all}.{pt,phi}",
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+    lepton_pair_momenta = lepton_pairs.sum(axis=-1)
+    lepton_pair_momenta_vis = lepton_pairs_vis.sum(axis=-1)
+    lepton_pair_momenta_all = lepton_pairs_all.sum(axis=-1)
+
+    # finally, save the pt and phi of the lepton pair on generator level
+    events = set_ak_column(events, "gen_dilepton_pt", lepton_pair_momenta.pt)
+    events = set_ak_column(events, "gen_dilepton_vis.pt", lepton_pair_momenta_vis.pt)
+    events = set_ak_column(events, "gen_dilepton_vis.phi", lepton_pair_momenta_vis.phi)
+    events = set_ak_column(events, "gen_dilepton_all.pt", lepton_pair_momenta_all.pt)
+    events = set_ak_column(events, "gen_dilepton_all.phi", lepton_pair_momenta_all.phi)
```
riga, line 108: Since this already stores pt and phi, would it make sense to also add eta and mass here? This would make this producer a bit more general and the objects could be casted to actual LorentzVectorArrays when read again.

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+def dy_weights_requires(self: Producer, reqs: dict) -> None:
+    """
+    Adds the requirements needed the underlying task to derive the Drell-Yan weights into *reqs*.
+    """
+    if "external_files" in reqs:
+        return
+
+    from columnflow.tasks.external import BundleExternalFiles
+
+    reqs["external_files"] = BundleExternalFiles.req(self.task)
```
riga, line 207:
```suggestion
def dy_weights_requires(self: Producer, task: law.Task, reqs: dict) -> None:
    """
    Adds the requirements needed the underlying task to derive the Drell-Yan weights into *reqs*.
    """
    if "external_files" in reqs:
        return

    from columnflow.tasks.external import BundleExternalFiles

    reqs["external_files"] = BundleExternalFiles.req(task)
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+@dy_weights.setup
+def dy_weights_setup(
+    self: Producer,
+    reqs: dict,
+    inputs: dict,
+    reader_targets: InsertableDict,
+) -> None:
```
riga, line 216: Just adding the task argument here.

```suggestion
def dy_weights_setup(
    self: Producer,
    task: law.Task,
    reqs: dict,
    inputs: dict,
    reader_targets: law.util.InsertableDict,
) -> None:
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+
+    # import all correctors from the external file
+    import correctionlib
+
+    correctionlib.highlevel.Correction.__call__ = (
+        correctionlib.highlevel.Correction.evaluate
+    )
+    correction_set = correctionlib.CorrectionSet.from_string(
+        self.get_dy_weight_file(bundle.files).load(formatter="gzip").decode("utf-8"),
+    )
```
riga, line 233: Here we can use the new `load_correction_set` helper.

```suggestion
    correction_set = load_correction_set(self.get_dy_weight_file(bundle.files))
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+
+from __future__ import annotations
+
+import law
+
+from dataclasses import dataclass
+
+from columnflow.production import Producer, producer
+from columnflow.util import maybe_import 
+from law.util import InsertableDict
```
riga, line 15:
```suggestion
from columnflow.util import maybe_import, load_correction_set
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+@recoil_corrections.requires
+def recoil_corrections_requires(self: Producer, reqs: dict) -> None:
+    # Ensure that external files are bundled.
+    if "external_files" in reqs:
+        return
+    from columnflow.tasks.external import BundleExternalFiles
+
+    reqs["external_files"] = BundleExternalFiles.req(self.task)
```
riga, line 428: Adding the task argument here.

```suggestion
def recoil_corrections_requires(self: Producer, task: law.Task, reqs: dict) -> None:
    # Ensure that external files are bundled.
    if "external_files" in reqs:
        return

    from columnflow.tasks.external import BundleExternalFiles
    reqs["external_files"] = BundleExternalFiles.req(task)
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+    # --- Systematic variations ---
+    # Derive H from the nominal corrected MET: H = - (MET_corr + vis)
+    Hx = -met_corr_x - vis.x
+    Hy = -met_corr_y - vis.y
+    H_pt = np.sqrt(Hx**2 + Hy**2)
+    H_phi = np.arctan2(Hy, Hx)
+    # Project H into the full boson coordinate system
+    H_para = H_pt * np.cos(H_phi - full.phi)
+    H_perp = H_pt * np.sin(H_phi - full.phi)
+
```
riga, line 368: Could you go for lower case variables + snake-case here?

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+    if "external_files" in reqs:
+        return
+    from columnflow.tasks.external import BundleExternalFiles
+
+    reqs["external_files"] = BundleExternalFiles.req(self.task)
+
+
+@recoil_corrections.setup
+def recoil_corrections_setup(
+    self: Producer, reqs: dict, inputs: dict, reader_targets: InsertableDict,
```
riga, line 433: Adding the task argument here.

```suggestion
    self: Producer,
    task: law.Task,
    reqs: dict,
    inputs: dict,
    reader_targets: law.util.InsertableDict,
```

columnflow/production/cms/dy.py
```diff
@@ -0,0 +1,459 @@
...
+    import correctionlib
+
+    correctionlib.highlevel.Correction.__call__ = (
+        correctionlib.highlevel.Correction.evaluate
+    )
+
+    # Load the correction set from the external file.
+    correction_set = correctionlib.CorrectionSet.from_string(
+        self.get_dy_recoil_file(bundle.files).load(formatter="gzip").decode("utf-8"),
+    )
```
riga, line 450: Using the new `load_correction_set` helper.

```suggestion
    correction_set = load_correction_set(self.get_dy_recoil_file(bundle.files))
```

riga (changes requested):
Hey @aalvesan and @philippgadow !

This looks pretty nice and very streamlined, and I reckon this will be very useful for many analyses to consider. Thanks for that.

I only have a few inline comments that should help with transitioning to the [new task array function interface](https://columnflow.readthedocs.io/en/refactor-taf_init/user_guide/task_array_functions.html) and to add couple helpers that were introduced during the refactoring.

Looking forward to merge this one.

riga:
@all-contributors please add @philippgadow for code.

columnflow/production/cms/dy.py
philippgadow: Thanks, good suggestion. I changed also the `produces`.

columnflow/production/cms/dy.py
philippgadow: Fixed.

riga:
Thanks @philippgadow for the latest updates!

The code looks good to me, I'd just have two quick questions on the physics / the method (and sorry for coming up with this only know).

- Are `gen_dilepton_pt` and `gen_dilepton_all.pt` potentially the same?
- `lepton_all_mask` matches all direct children of Z/gamma's, correct? If so, then `lepton_vis_mask` matches the same particles except for neutrinos. But with that, `gen_dilepton_vis` would contain 1 or 2 neutrinos of the subsequent tau decay. Is that intended in terms of what the DY corrections expect? Maybe one or two additional lines in the docstring could be helpful.

PR #628, opened 2025-02-15
Refactor task array function interface

riga:
This PR refactors the task array function (calibrators, selectors, producers, etc) interface. The refactoring is necessary due to a flaw in the handling between tasks and array functions, causing a variety of performance drawbacks and ambiguous behavior, such as

- an undesired feedback loop / interdependence between task array functions and shifts they are allowed to "yield" up to tasks (currently causing "real data" sets to accept systematic shifts),
- an exceedingly large repetition of initialization calls,
- an unclear definition of when these init calls are invoked and which task attributes they have access to, and
- a bug introduced (but not initially caused) by https://github.com/columnflow/columnflow/pull/614 that leads to duplication of dependencies.

Details are given in [this presentation](https://indico.desy.de/event/48268/contributions/183329/attachments/95592/130878/2025-02-14_cf_array_function_interface.pdf).

The new interface consists of the following hooks:

hooks

With that, the changes in this PR can be subdivided into 4 groups:

<details><summary>1. Changes to TaskArrayFunction</summary>
- The hooks have been adapted in columnar_util.
- Tasks making direct use of them were adapted as well.
- Attributes passed from a task to the task array function init are fixed to always contain `{analysis,config,dataset}_inst`. They are known as members at all times and they can be considered constant (the references to them, not their content).
</details>

<details><summary>2. Splitting of mixins into class & instance specific ones</summary>
- Tasks that continue to need direct access to task array function **instances** are mostly unchanged.
- They have a new yet *hidden* parameter now for the actual instance, allowing them to pass it upstream to task dependencies without the need of re-initializing them.
- Other tasks (like plotting ones) only need to access the name and class of task array function. They now inherit from different mixins (e.g. CalibratorClassMixin), to make this distinction clear.
</details>

<details><summary>3. Adaptation of all shipped task array functions</summary>
- All calibrators, selectors, producers, and weight producers were adjusted. However, these changes mostly aimed to using the new interface (e.g. moving from `init` to `pre_init` or `post_init`).
- These changes make up most of this PR.
</details>

<details><summary>4. Additional changes</summary>
- This PR is based on https://github.com/columnflow/columnflow/pull/609 to start from a corrected inheritance pattern. The PR can be closed once *this* one is merged.
- Small issues that occurred during testing were fixed right away, e.g. htcondor submission and sandbox related things.
- Various doc strings were outdated or missing. I reduced them where I saw fit, and added new ones including type hints where I spotted them. At some point, we surely need another pass though.
</details>

**Note:** Since the ML pipeline is strongly connected to the multi-config efforts, corresponding tasks are untouched and should be adapted in a later stage (see below).

---

With that, the next steps are:

- [ ] Review changes.
- [ ] Port features of https://github.com/columnflow/columnflow/pull/566 and https://github.com/columnflow/columnflow/pull/593.
- [ ] Adjust ML mixins and tasks in the ML pipeline. They are connected to changes regarding multi-config tasks.
- [ ] Adjust documentation, especially provide user_guid/task_array_functions with the full interface description.

Since potentially breaking changes are introduced, we should release 0.2.5 before merging this PR, and 0.3.0 right after.


mafrahm:
With these changes to the `TaskArrayFunctions`, what are the expected changes on user side? I'll try to summarize

- `inits` need to be split into `pre_init` (only for deps_kwargs, so not used by that many CSPs), `init`, and `post_init` (e.g. for resolving columns after shifts are resolved)
- shift dependencies needs to be removed from the main `init`
- `TaskArrayFunctions` no longer carry the `task` as attribute -> remove `self.task` dependencies in `pre_init`, `init` and `skip`, pass (and use) `task` as additional input argument to other hooks (`post_init`, `requires`, `setup`) and in the call method itself

columnflow/tasks/cutflow.py
```diff
@@ -12,10 +12,10 @@
 import order as od
 
 from columnflow.tasks.framework.base import (
-    Requirements, AnalysisTask, DatasetTask, ShiftTask, wrapper_factory, RESOLVE_DEFAULT,
+    Requirements, AnalysisTask, wrapper_factory, RESOLVE_DEFAULT,
 )
 from columnflow.tasks.framework.mixins import (
-    CalibratorsMixin, SelectorStepsMixin, VariablesMixin, CategoriesMixin, ChunkedIOMixin,
```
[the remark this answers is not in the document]
mafrahm: Ah I see that you merged the `SelectorMixin` and `SelectorStepsMixin`. Does that mean we can require selector steps on command line in `SelectEvents` now or did you take care to remove this possibility?

columnflow/tasks/cms/inference.py
```diff
@@ -21,13 +21,13 @@
 
 
 class CreateDatacards(
-    HistHookMixin,
+    CalibratorClassesMixin,
+    SelectorClassMixin,
+    ProducerClassesMixin,
```
mafrahm, line 26: I remember that changing this order of inheritance messed up the order of how CSP init functions were being run. Is this fixed with this PR?

columnflow/tasks/production.py
```diff
@@ -84,8 +91,13 @@ def run(self):
...
         # run the producer setup
-        producer_reqs = self.producer_inst.run_requires()
-        reader_targets = self.producer_inst.run_setup(producer_reqs, luigi.task.getpaths(producer_reqs))
+        self._array_function_post_init()
+        producer_reqs = self.producer_inst.run_requires(task=self)
+        reader_targets = self.producer_inst.run_setup(
+            task=self,
+            reqs=producer_reqs,
+            inputs=luigi.task.getpaths(producer_reqs),
+        )
```
mafrahm, line 100: are the `post_init` only being run directly before the requires func or are they also invoked outside of a run method sometimes?
I am thinking a bit on the use cases like using the `btag_weights.PRODUCES` in the `uses` of another CSP, does that still work?
Example: https://github.com/uhh-cms/hh2bbww/blob/c88c9bca0e72d7e6fd2b9b4da155624ff168bfd0/hbw/production/normalized_btag.py#L21

columnflow/tasks/framework/mixins.py
```diff
@@ -1852,34 +1767,44 @@ def resolve_param_values(cls, params: dict[str, Any]) -> dict[str, Any]:
...
+        :return: Dictionary with parts that will be translated into an output directory path.
+        """
         parts = super().store_parts()
         if self.inference_model != law.NO_STR:
-            parts.insert_before("version", "inf_model", f"inf__{self.inference_model_repr}")
+            parts.insert_after(self.config_store_anchor, "inf_model", f"inf__{self.inference_model_repr}")
         return parts
 
 
+class InferenceModelMixin(InferenceModelClassMixin):
```
mafrahm, line 1791: does this mixin not need a `inference_model_inst` parameter?

mafrahm (commented):
Changes make sense to me overall, although I'm sure I did not yet understand all the details. Especially in `mixins.py`, it's quite hard to read the diffs, although the final results looks rather clear now.

I left some inline comments, mostly about details/functionality where I am unsure if this PR might influence certain behaviours. I suppose many things will clear up when merging with #566 and directly testing functionality.

columnflow/columnar_util.py
```diff
@@ -2063,19 +2085,25 @@ def _get_used_columns(self, _cache: set | None = None) -> set[Route]:
 
     @property
     def used_columns(self) -> set[Route]:
-        return self._get_used_columns()
+        try:
+            return self._get_used_columns()
+        except AttributeError as e:
+            raise Exception(str(e)) from e
```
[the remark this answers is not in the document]
riga: Python allows properties to "transparently" raise AttributeErrors in which case "hasattr" would just return False. I redirected the exception type to avoid that and really raise.

columnflow/production/cms/btag.py
```diff
@@ -275,20 +276,21 @@ def add_weight(syst_name, syst_direction, column_name):
     return events
 
 
-@btag_weights.init
-def btag_weights_init(self: Producer) -> None:
+@btag_weights.post_init
+def btag_weights_post_init(self: Producer, task: law.Task, **kwargs) -> None:
```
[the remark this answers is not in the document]
riga: At the moment, yes. There are two ways to avoid that:

- we can add a flag denoting if a producer will eventually have a non-empty `produces` set, independent of whether it's empty after pre_init and init, or
- we call post_init in the mixin's `__init__`'s, with the `self` as the task reference.

I am in favor of the latter.

columnflow/tasks/calibration.py
```diff
@@ -16,13 +16,18 @@
 ak = maybe_import("awkward")
 
 
-class CalibrateEvents(
+class _CalibrateEvents(
```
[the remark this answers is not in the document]
riga: In case a user wants to create an analysis-specific variant of some tasks - say, CalibrateEvents - it's way easier to just inherit from the same original classes like that. Future updates from our cf side will transparently be included in user tasks.

columnflow/tasks/framework/mixins.py
```diff
@@ -25,203 +29,186 @@
...
     """
+
     calibrator = luigi.Parameter(
         default=RESOLVE_DEFAULT,
         description="the name of the calibrator to be applied; default: value of the "
         "'default_calibrator' config",
     )
-    calibrator.__annotations__ = " ".join("""
-        the name of the calibrator to be applied; default: value of the
-        'default_calibrator' config""".split())
```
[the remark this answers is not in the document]
riga: I'll add it back as

```python
#: the description
calibrator = ...
```

which is accepted by sphinx.

columnflow/tasks/cms/inference.py
riga: Yup :+1:

columnflow/tasks/production.py
riga: Yes, that still works.

columnflow/tasks/framework/mixins.py
riga: Indeed. I generalized the `TaskArrayFunction(s)Parameter` to `DerivableInst(s)Parameter` to be able to capture Derivable subclasses as well and added it :+1:

maadcoen:
Hi! I noted that the ML tasks that use the preparation producer aren't calling the post_init function before setup. Is this meant to be like this?

riga:
> Hi! I noted that the ML tasks that use the preparation producer aren't calling the post_init function before setup. Is this meant to be like this?

I think you are right @maadcoen. I added this in 9d5283b8, could you check if that works for you? Also pinging @mafrahm here again as he is perhaps more aware of the intricacies of the ML mixins.

mafrahm (approved):
LGTM

riga:
Let's go!

PR #632, opened 2025-02-25
correct eta calculation trigger sf

nprouvost:
ID SF need supercluster eta, trigger SF need Electron eta, this option is added in this PR

### DIFF
--- a/columnflow/production/cms/electron.py
+++ b/columnflow/production/cms/electron.py
@@ -53,7 +53,7 @@ class ElectronSFConfig:
     get_electron_file=(lambda self, external_files: external_files.electron_sf),
     # function to determine the electron weight config
     get_electron_config=(lambda self: ElectronSFConfig.new(self.config_inst.x.electron_sf_names)),
-    # choose if the eta variable should be the electron eta (triggerSF) or the super cluster eta (IDSF)
+    # choose if the eta variable should be the electron eta or the super cluster eta
     use_supercluster_eta=True,
     weight_name="electron_weight",
     supported_versions=(1, 2, 3),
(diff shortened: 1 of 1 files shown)


columnflow/production/cms/electron.py
```diff
@@ -53,6 +53,8 @@ def new(
     get_electron_file=(lambda self, external_files: external_files.electron_sf),
     # function to determine the electron weight config
     get_electron_config=(lambda self: ElectronSFConfig.new(self.config_inst.x.electron_sf_names)),
+    # choose if the eta variable should be the electron eta (triggerSF) or the super cluster eta (IDSF)
```
riga, line 56:
```suggestion
    # choose if the eta variable should be the electron eta or the super cluster eta
```

columnflow/production/cms/electron.py
```diff
@@ -92,11 +94,14 @@ def electron_weights(
...
     based only on a subset of electrons.
     """
-    # flat super cluster eta and pt views
-    sc_eta = flat_np_view((
-        events.Electron.eta[electron_mask] +
-        events.Electron.deltaEtaSC[electron_mask]
-    ), axis=1)
+    # flat super cluster eta/flat eta and pt views
+    if self.use_supercluster_eta:
+        sc_eta = flat_np_view((
```
riga, line 99: We should name this eta then, also below.

riga (changes requested):
Thanks @nprouvost.

Just two super minor inline comments, before we can merge.

columnflow/production/cms/electron.py
nprouvost: done

columnflow/production/cms/electron.py
nprouvost: done

riga (approved):
:+1:

PR #638, opened 2025-03-04
Add DY recoil corrections to DY ptll correction producer

philippgadow:
This PR targets Ana's branch which adds the DY ptll corrections and adds on top the DY recoil corrections, so that both corrections can be merged with a single PR to the column flow master branch.

### DIFF
--- a/columnflow/production/cms/dy.py
+++ b/columnflow/production/cms/dy.py
@@ -33,26 +33,42 @@ class DrellYanConfig:
             raise ValueError("Incomplete dy_weight_config: missing era, order, correction or unc_correction.")  # noqa
 
 
+@dataclass
+class DrellYanConfig:
+    era: str
+    order: str
+    correction: str
+    unc_correction: str
+
+    def __post_init__(self) -> None:
+        if not self.era or not self.order or not self.correction or not self.unc_correction:
+            raise ValueError("Incomplete dy_weight_config: missing era, order, correction or unc_correction.")  # noqa
+
+
 @producer(
     uses={"GenPart.*"},
-    produces={"gen_dilepton.pt"},
+    produces={
+        "gen_dilepton_pt",
+        "gen_dilepton_vis.{pt,phi}",
+        "gen_dilepton_all.{pt,phi}"
+    },
 )
 def gen_dilepton(self, events: ak.Array, **kwargs) -> ak.Array:
     """
-    Reconstruct the di-lepton pair from generator level info.
+    Reconstruct the di-lepton pair from generator level info. This considers only visible final-state particles. In addition it provides the four-momenta of all leptons (including neutrinos) from the hard process.
     """
 
     # get the absolute pdg id (to account for anti-particles) and status of the particles
     pdg_id = abs(events.GenPart.pdgId)
     status = events.GenPart.status
 
-    # electrons and muons need to have status == 1, so find them
+    # lepton masks for DY ptll reweighting corrections
+    # -> https://indico.cern.ch/event/1495537/contributions/6359516/attachments/3014424/5315938/HLepRare_25.02.14.pdf
     ele_mu_mask = (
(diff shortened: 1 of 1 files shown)


columnflow/production/cms/dy.py
```diff
@@ -32,48 +32,74 @@ def __post_init__(self) -> None:
...
         (pdg_id == 15) &
         (status == 2) &
         events.GenPart.hasFlags("fromHardProcess")
     )
 
+    # lepton masks for recoil corrections
+    # -> https://indico.cern.ch/event/1495537/contributions/6359516/attachments/3014424/5315938/HLepRare_25.02.14.pdf
+    lepton_vis_mask = (
+        # e, mu, taus, neutrinos
+        (((pdg_id == 11) | (pdg_id <= 13)| (pdg_id <= 15)) & (status == 1) & events.GenPart.hasFlags("fromHardProcess")) |
```
mafrahm, line 70: currently, this mask considers all particles with pdgid <= 15
```suggestion
        (((pdg_id == 11) | (pdg_id == 13)| (pdg_id == 15)) & (status == 1) & events.GenPart.hasFlags("fromHardProcess")) |
```

Also maybe more of a question regarding tau's: you are both trying to select the tau directly or all the decay product of the tau (`events.GenPart.hasFlags("isDirectHardProcessTauDecayProduct")`). Is it possible that you might select both the tau and it's decay products at the same time? I suppose this is prevented because tau's should always have `status == 2`, right?

mafrahm: I kind of suspect that you are missing some brackets. I think at the moment, this mask returns true for all gen particles that have `events.GenPart.hasFlags("isDirectHardProcessTauDecayProduct") = True`, so you would also select the electron/muon neutrinos from tau decays
```suggestion
        (
            ((pdg_id == 11) | (pdg_id == 13)| (pdg_id == 15)) & 
            (((status == 1) & events.GenPart.hasFlags("fromHardProcess")) | events.GenPart.hasFlags("isDirectHardProcessTauDecayProduct"))
        )
```

philippgadow: Hi @mafrahm , thank you for spotting the issue with the lepton vis mask!
I found in [the slides by Dennis Roy](https://indico.cern.ch/event/1495537/contributions/6359516/attachments/3014424/5315938/HLepRare_25.02.14.pdf) a snipped of what definition was used in deriving the corrections. 
image

I think for the `lepton_vis_mask` the correct and simplest approach is to take the `lepton_all_mask` and hide all particles with pdgID in [12,14,16] to remove neutrinos.

mafrahm: Yes you're probably right. My previous suggestion would also remove non-lepton tau decay products, which (I suspect) would also be wrong.

```python
lepton_vis_mask = lepton_all_mask & (pdg_id != 12) & (pdg_id != 14) & (pdg_id != 16)
```

philippgadow: Yes, that does the job. Thank you :)

aalvesan (approved):

columnflow/production/cms/dy.py
```diff
@@ -29,51 +29,88 @@ class DrellYanConfig:
...
-        events.GenPart.hasFlags("fromHardProcess")
+        (pdg_id == 15) & (status == 2) & events.GenPart.hasFlags("fromHardProcess")
+    )
+
+    # lepton masks for recoil corrections
+    # -> https://indico.cern.ch/event/1495537/contributions/6359516/attachments/3014424/5315938/HLepRare_25.02.14.pdf
+    lepton_all_mask = (
+        # e, mu, taus, neutrinos
+        (
+            (pdg_id >= 11)
```
Lara813, line 78: I think an 'or' is missing at the end of the line

PR #639, opened 2025-03-05
Update shift resolution and shift sources validation.

riga:
- Adds shift resolution based on an "upstream" `shift_resolution_task_cls` to `ShiftTask`s with multi-config support. It's named differently compared to `shift_validation_task_cls` of the `ShiftSourcesMixin` as it does something slightly different (not just saying `True`/`False` on "Does this shift exist?" but it performs the resolution itself, optionally looping through other parameters).
- Streamlines the shift validation in `ShiftSourcesMixin` to accept an actual shift name, rather than a source for which previously, only the `*_up` shift was checked.

### DIFF
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -1460,15 +1460,6 @@ class ShiftTask(ConfigTask):
                     params["local_shift_insts"] = {config_inst: params["local_shift_inst"]}
 
         elif (config_insts := params.get("config_insts")):
-            # a shift validatio task supporting a single config must be set
-            if cls.shift_validation_task_cls is None:
-                raise AttributeError(
-                    f"shift resolution of multi-config task '{cls}' requires 'shift_validation_task_cls' to be set",
-                )
-            if not cls.shift_validation_task_cls.has_single_config():
-                raise ValueError(
-                    f"shift_validation_task_cls '{cls.shift_validation_task_cls}' must have a single config",
-                )
             configs_repr = lambda: ", ".join(map(repr, config_insts))
             missing_pair = (None, None)
             empty_pair = (law.NO_STR, law.NO_STR)
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -1403,6 +1403,7 @@ class ShiftTask(ConfigTask):
     exclude_params_remote_workflow = {"local_shift"}
 
     allow_empty_shift = False
+    shift_resolution_task_cls = None
 
     @classmethod
     def modify_param_values(cls, params: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


columnflow/tasks/framework/base.py
```diff
@@ -1434,96 +1459,95 @@ def resolve_shifts(cls, params: dict[str, Any]) -> dict:
...
+        if not cls.shift_resolution_task_cls.has_single_config():
+            raise ValueError(
+                f"shift_resolution_task_cls {cls.shift_resolution_task_cls} must support a single config",
+            )
+        # run the parameter resolution and return obtained shifts
+        _params = params | {
+            "config": config_inst.name,
+            "config_inst": config_inst,
+        }
+        cls.shift_resolution_task_cls.resolve_shifts(_params, allow_missing=True)
```
mafrahm, line 1546: 1.) the `cls.shift_resolution_task_cls.resolve_shifts` will never be called when `cls` is a single config task
2.) at which stage do we call the `cls.shift_resolution_task_cls.resolve_shifts`per dataset? To me it seems that this is only being done for the `DatasetsProcessesShiftSourcesMixin`. Do we also need a `DatasetsProcessesShiftMixin` doing something similar?

mafrahm (changes requested):
I feel like the `ShiftTask.resolve_shifts` function should be more or less the same for single config and multi config, the main difference should lay in handling tasks with and without a `cls.shift_resolution_task_cls`

columnflow/tasks/framework/base.py
```diff
@@ -1434,96 +1459,95 @@ def resolve_shifts(cls, params: dict[str, Any]) -> dict:
...
+        elif (config_insts := params.get("config_insts")):
+            # a shift validatio task supporting a single config must be set
+            if cls.shift_validation_task_cls is None:
+                raise AttributeError(
+                    f"shift resolution of multi-config task '{cls}' requires 'shift_validation_task_cls' to be set",
+                )
+            if not cls.shift_validation_task_cls.has_single_config():
+                raise ValueError(
+                    f"shift_validation_task_cls '{cls.shift_validation_task_cls}' must have a single config",
+                )
```
mafrahm, line 1471: Code duplication (this is already part of `cls._resolve_upstream_shifts`
```suggestion
```

riga: True, I forgot I moved it.

columnflow/tasks/framework/base.py
riga: 1) I'm going to add this right now :+1: Regarding the similarity between the sc and mc handling, I wanted to keep them separate for now, and streamline it after we managed to make it behave the way we want.
2) Indeed, this is to be done by subclasses just like we do with shift sources. `DatasetsProcessesShiftMixin` sounds reasonable and would fall right into place. I can create a shallow task in *this* PR, but I would leave it to the next one to use it in the single shift plot base.

PR #644, opened 2025-03-19
Improve known_shifts caching between workflow and branches.

riga:


mafrahm:
This test case seems to fail now: when passing `RESOLVE_DEFAULT` (single value, not a tuple) to the `resolve_config_default`, we expect to return only the first element of the resolved param, but now the whole list seems to be returned.
```python
        resolved_producer = AnalysisTask.resolve_config_default(
            param=RESOLVE_DEFAULT,
            task_params=self.base_params,
            container=self.analysis_inst,
            default_str="default_producer",
            multi_strategy="first",
        )
        self.assertEqual(resolved_producer, "A")
 ```

mafrahm (approved):
LGTM

PR #648, opened 2025-03-21
Add reducer interface.

riga:


mafrahm (approved):
LGTM. Looks like backwards compatibility can be achieved (if necessary) with a rather simple `store_parts_modifier`:

```python
    def pre_reducer_parts(task, store_parts):
        store_parts.pop("reducer")
        store_parts["weight_producer"] = store_parts["weight_producer"].replace("weight", "hist")
        return store_parts
    
    analysis_inst.x.store_parts_modifiers = {
        "pre_reducer": pre_reducer_parts,
    }
```

PR #649, opened 2025-03-21
Systematic shift plotting

riga:
This PR adds all the necessary bits and pieces to plot systematic errors due to shift variations. Compared to statistical errors that are already shown, it adds a differently hatched band to denote systematics.

The following changes were made:

- Added the `draw_syst_error_bands` function, and renamed the existing `draw_error_bands` to `draw_stat_error_bands` to distinguish them. The syst band plotting required a couple helpers for combining shifts.
- Overhauled some plot functionality (mostly regarding plot dimensions and label handling in legends) and slowly started forwarding additional objects to plot functions (e.g. `shift_insts`) to eventually phase-out the necessity to pass the entire `config_inst` or even several of them.
- Added a new task that triggers the plot. As discussed a while back, the name of the already existing `PlotShiftedVariables1D` task actually fits best for that type of plot, so I renamed the existing one to `PlotShiftedVariablesPerShift1D` to use `PlotShiftedVariables1D` for the new task. This is in line with other plot tasks that create a plot for a specific entity (like PlotVariablesPerProcess1D that creates a single plot per process). In a similar fashion, the existing task now produces multiple plots with one shift applied for each of them. I updated the [graph in our wiki](https://github.com/columnflow/columnflow/wiki) (orange box) to visually check if the distinction makes sense for the task structure.


columnflow/plotting/plot_util.py
```diff
@@ -379,13 +356,10 @@ def use_flow_bins(
...
 
 
-def apply_density_to_hists(hists: dict, density: bool | None = False) -> dict:
+def apply_density(hists: dict) -> dict:
     """
     Scales number of histogram entries to bin widths.
     """
-    if not density:
-        return hists
-
```
mafrahm, line 388: I don't understand this change. I much prefer to keep this in here such that all the `apply_<parameter>` functions are just one-liners such that resulting plot functions are as compact as possible. Especially since I already have quite a lot of custom plot functions which I'd now need to fix accordingly.

columnflow/plotting/plot_util.py
```diff
@@ -396,22 +370,49 @@ def apply_density_to_hists(hists: dict, density: bool | None = False) -> dict:
     return hists
 
 
-def remove_residual_axis(hists: dict, ax_name: str, max_bins: int = 1) -> dict:
+def _remove_residual_axis(
```
mafrahm, line 373: Maybe just a bit more verbosity to understand the differences between the two functions, e.g.
```suggestion
def remove_residual_axis_single_hist(
```

columnflow/plotting/plot_util.py
```diff
@@ -486,31 +488,31 @@ def prepare_stack_plot_config(
...
-                line_hists.append(h)
-                line_colors.append(process_inst.color1)
-                line_labels.append(process_inst.label)
-                line_hide_errors.append(proc_hide_errors)
-            else:
-                mc_hists.append(h)
-                mc_colors.append(process_inst.color1)
-                mc_edgecolors.append(process_inst.color2)
-                mc_labels.append(process_inst.label)
+        elif getattr(process_inst, "unstack", False):
```
mafrahm, line 504: As mentioned above
```suggestion
        elif getattr(process_inst, "unstack", False) or process_inst.x("unstack", False):
```

columnflow/plotting/plot_all.py
```diff
@@ -8,51 +8,154 @@
...
+    syst_error_up = []
+    syst_error_down = []
+    for b in range(h.axes[0].size):
+        up_diffs = []
+        down_diffs = []
+        for source, (up_shift, down_shift) in shift_groups.items():
+            up_diff = shift_stacks[up_shift].values()[b] - h.values()[b]
+            down_diff = shift_stacks[down_shift].values()[b] - h.values()[b]
+            up_diffs.append(max(up_diff, down_diff, 0))
+            down_diffs.append(min(up_diff, down_diff, 0))
```
mafrahm, line 122: maybe rename variable names for more verbosity? e.g.
`up_diff` --> `shift_up_diff`
`up_diffs` --> `value_up_diffs`

columnflow/plotting/plot_all.py
```diff
@@ -8,51 +8,154 @@
...
+
+    # loop over bins, subtract nominal yields from stacked yields and merge differences into
+    # a systematic error per bin using the given method (quadratic sum vs. evelope)
+    # note 1: if the up/down variations of the same shift source point in the same direction, a
+    #         statistical combination is pointless and their minimum/maximum is selected instead
+    # note 2: relative signs are consumed into the meaning of "up" and "down" here as they already
+    #         are combinations evaluated for a specific direction
+    syst_error_up = []
+    syst_error_down = []
+    for b in range(h.axes[0].size):
```
mafrahm, line 115: axis 0 is the axis defined by the variable, correct? Can we pick the axis from it's name instead to be independent from axis ordering?

columnflow/plotting/plot_all.py
```diff
@@ -8,51 +8,154 @@
...
+        "bottom": baseline * (1 - rel_syst_error_down),
+        "height": baseline * (rel_syst_error_up + rel_syst_error_down),
+        **kwargs,
+        "width": h.axes[0].edges[1:] - h.axes[0].edges[:-1],
+        "hatch": "\\\\\\",
+        "linewidth": 0,
+        "color": "none",
+        "edgecolor": "#30c300",
+        "alpha": 1.0,
+    }
```
mafrahm, line 157: To give priority to values defined in `**kwargs`
```suggestion
    bar_kwargs = {
        "x": h.axes[0].centers,
        "bottom": baseline * (1 - rel_syst_error_down),
        "height": baseline * (rel_syst_error_up + rel_syst_error_down),
        "width": h.axes[0].edges[1:] - h.axes[0].edges[:-1],
        "hatch": "\\\\\\",
        "linewidth": 0,
        "color": "none",
        "edgecolor": "#30c300",
        "alpha": 1.0,
        **kwargs,
    }
```

mafrahm (changes requested):
This look nice, thank you! Just a few minor comments that should be easy to address.

columnflow/tasks/plotting.py
```diff
@@ -461,30 +466,34 @@ class PlotVariablesBaseMultiShifts(
...
+            keys.append("shift_source")
+        return [DotDict(zip(keys, vals)) for vals in itertools.product(*seqs)]
 
     def requires(self):
-        req = {}
+        req_cls = lambda dataset_name: (
+            self.reqs.MergeShiftedHistograms
+            if self.config_inst.get_dataset(dataset_name).is_mc
+            else self.reqs.MergeHistograms
+        )
```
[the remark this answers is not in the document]
riga: Yeah, this also bothered me a bit, but I think running an extra task that is not really required is a bigger drawback.

But seeing the point, how about law receives the feature like

```
law run TASK --remove-output cf.CreateHistograms,a
```

meaning "remove everyhing up to (and including) cf.CreateHistograms". I think this would also be convenient in other places.

columnflow/plotting/plot_util.py
riga: Imho, calling a method even when it's clear it should not have an effect and thus, not be called in the first place is not a pattern that I would generally use.

However, I would include it again for backwards compatibility to not break existing user code as you pointed out.

columnflow/plotting/plot_util.py
```diff
@@ -233,16 +189,20 @@ def apply_process_settings(
...
 
     def get_stack_integral() -> float:
         nonlocal stack_integral
         if stack_integral is None:
             stack_integral = sum(
-                proc_h.sum().value
+                _remove_residual_axis(proc_h, "shift", select_value="nominal").sum().value
                 for proc, proc_h in hists.items()
-                if not hasattr(proc, "unstack") and not proc.is_data
+                if proc.is_mc and not getattr(proc, "unstack", False)
```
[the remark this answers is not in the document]
riga: I added a helper `get_attr_or_aux` that is used now in all spots where attribute or aux data access is needed as a fallback.

columnflow/plotting/plot_all.py
riga: Syst error band plotting is only supported for 1-D cases anyway, so 0 should always be ok in this implementation. In case histograms passed to this function are more than 1-D, the code will anyways require adjustments.

columnflow/plotting/plot_all.py
riga: I'm not sure if I'd understand the difference between `shift_` and `value_` here.

riga:
Thanks for reviewing! I made adjustments in 92f70d1 to include most of your feedback. I left three conversations unresolved to iterate on some decisions.

columnflow/plotting/plot_all.py
mafrahm: Ah sure, I thought about other axes aswell (e.g. `shift`), but they are also already separated into individual histograms at this point.

columnflow/tasks/plotting.py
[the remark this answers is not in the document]
mafrahm: I do not see a particularly large drawback in running an unnecessary task that does not really require much computation time or storage space. Nevertheless, these changes are still fine with me.

And this feature would be really nice to have! I suppose there could be some caveats that might be difficult to resolve. 
e.g. what happens when running 
```
law run TASK --remove-output cf.ProduceColumns,a
```
since there are some tasks that never require this particular task. I suppose one then needs to resolve the full task tree. 

In any case, this discussion is not really part of this PR, so this comment can be resolved :+1:

columnflow/plotting/plot_all.py
mafrahm: I agree with you, I just didn't come up with anything better yet.
I'd appreciate to somehow separate the "shift diffs" (`up` can both increase and decrease) and "value diffs" (`up` always means that the value increases).

riga: Yes, thinking about it, I agree.

In 580d0d79 I added comments and renamed `{up,down}_diff` to `shift_{up,down}_diff`.

mafrahm (approved):

PR #653, opened 2025-03-26
add hist-hooks-short-repr parameter

nprouvost:
add parameter to hash hist hook names to avoid too long output paths

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -2447,6 +2447,13 @@ class HistHookMixin(ConfigTask):
         "default: empty",
     )
 
+    hist_hooks_short_repr = luigi.BoolParameter(
+        default=False,
+        description="when True, all hist hooks names are shortened to a single hash in the path,"
+        "when False, only the first 5 hist hooks names are used to create the store part, the rest "
+        "is hashed; default: False",
+    )
+
     def invoke_hist_hooks(self, hists: dict) -> dict:
         """
         Invoke hooks to update histograms before plotting.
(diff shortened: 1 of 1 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -2479,10 +2486,13 @@ def hist_hooks_repr(self) -> str:
         """
         Return a string representation of the hist hooks.
         """
-        hooks = [hook for hook in self.hist_hooks if hook not in (None, "", law.NO_STR)]
+        if self.hist_hooks_short_repr:
+            hooks_repr = law.util.create_hash(self.hist_hooks)
```
mafrahm, line 2490: I'd suggest to add the number of used hist-hooks to the output path because it can help immensely to distinguish output paths of different combinations of hist hooks.
```suggestion
            hooks_repr = f"{len(self.hist_hooks))}__{law.util.create_hash(self.hist_hooks)}"
```

mafrahm (commented):
To me it seems a bit overkill to add a command-line parameter for this type of customization - I don't really see a world where you want switch back and forth between the two configurations. I'd prefer to customize this via a simple classattribute, which can be overwritten from user-side via patches (btw. you can also just overwrite the complete `hist_hooks_repr` via patches).

nprouvost:
> To me it seems a bit overkill to add a command-line parameter for this type of customization - I don't really see a world where you want switch back and forth between the two configurations. I'd prefer to customize this via a simple classattribute, which can be overwritten from user-side via patches (btw. you can also just overwrite the complete `hist_hooks_repr` via patches).

Sure, if you think there is a better option, I don't mind going with that, I just wanted to propose the feature, since I already needed this (or something similar) for a while now. And I'm actually using both configurations at the moment, since the default configuration gives the name of the hooks in the path, which is convenient to find your files when needed, without needing to open them or relying on --print-output, while I use this argument only when the path becomes too long and the default cannot be used anymore. But again, If there is a better way to reach a similar result, I would be happy using that too!

columnflow/tasks/framework/mixins.py
nprouvost: sounds great, I also added independency on the order of the hooks at the same time.

riga:
Maybe, for hist hooks, we can define an attribute on the hook function itself that is used for the store path.

We should pick up this topic in the framework meeting to find a general approach (or to identify if one of the existing methods can already do the job).

riga:
As discussed earlier, let's close this one on favor of https://github.com/columnflow/columnflow/pull/658.

PR #654, opened 2025-03-27
Fix pu weight extraction.

riga:
This PR fixes the way that PU weights are extraction in the correctlib based implementation.


mafrahm (approved):
Good catch, thank you!

PR #655, opened 2025-03-28
typo on inference task

JulesVandenbroeck:
There is a typo in the `cf.CreateDatacards` task that causes the task not to work.

Fix:
- replace ´proc_obj_name´ with ´proc_obj.name´ in L307 & L310


riga (approved):
Thanks @JulesVandenbroeck , this looks good to me.

PR #656, opened 2025-03-28
Code harmonization

riga:
This PR does not change any behavior, but just does the following things:

- Harmonized alignment in mixins and base.py
- Added some missing type hints
- Removed comments left during development
- Renamed `resolution_task_class` -> `resolution_task_cls` for consistency with other places where task classes are referenced.

### DIFF
--- a/columnflow/tasks/framework/mixins.py
+++ b/columnflow/tasks/framework/mixins.py
@@ -24,8 +24,9 @@ from columnflow.production import Producer
 from columnflow.histograming import HistProducer
 from columnflow.ml import MLModel
 from columnflow.inference import InferenceModel
-from columnflow.columnar_util import Route, ColumnCollection, ChunkedIOHandler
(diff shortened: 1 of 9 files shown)


columnflow/tasks/framework/mixins.py
```diff
@@ -36,9 +37,9 @@
 
 class ArrayFunctionClassMixin(ConfigTask):
 
-    def array_function_cls_repr(self, array_function) -> str:
+    def array_function_cls_repr(self, array_function: Type[TaskArrayFunction]) -> str:
```
mafrahm, line 40:
```suggestion
    def array_function_cls_repr(self, array_function: str) -> str:
```

mafrahm (changes requested):
LGTM, just one inline comment where the type hint was wrong

columnflow/tasks/framework/mixins.py
riga: Added in 0baa5feb (had to remove an import as well)

PR #657, opened 2025-03-28
Use process names in hist axes.

riga:
This PR applies the same id-to-name conversion to the "process" axis if histograms that is already in place for "category" and "shift".

I updated the docs accordingly, and also adapted the `get_config_process_map` lookup method.


mafrahm:
I think there are some more locations where we pick process ids out of histograms that would need to be updated (yields, cutflow, framework/histograms, inference)

riga:
Added in last commit.

mafrahm (approved):
LGTM

PR #661, opened 2025-03-31
CMS jet id producer

riga:
This PR adds a producer that recalculates the CMS jet id, which is [required in all recent nano versions](https://cms-talk.web.cern.ch/t/bug-in-the-jetid-flag-in-nanov12-and-nanov13/108135).

The same producer can be used to create the id for every jet type. The centrally provided correctionlib file contains corrections for all jet types and all selection flags. For parity with the original `jetId` column, the `JetIdConfig` could be set to

```python
JetIdConfig(corrections={
    "AK4PUPPI_Tight": 2,
    "AK4PUPPI_TightLeptonVeto": 3,
})
```

for standard AK4 jets.

### DIFF
--- a/columnflow/production/cms/jet.py
+++ b/columnflow/production/cms/jet.py
@@ -1,18 +1,176 @@
 # coding: utf-8
 
 """
-Jet-related quantities
+Jet-related producers.
 """
 
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


mafrahm (approved):
LGTM

PR #662, opened 2025-03-31
Eager taf teardown when call function fails.

riga:
This PR is straight forward and updates all tasks with task array function involvement to run the teardown method eagerly as soon as the run method fails. This is implemented through a `on_failure` decorator and a new method in all relevant mixins to trigger the teardown with correct parameters.

This allows resources to be freed up in case of issues of any kind.

---

Background: our ML evaluation of externally trained models is encapsulated into subprocesses for better resource control. However, in case a different producer fails, we need to trigger the subprocess termination, since it would otherwise block the main process as it never receives a stop signal. This already lead to remote jobs that are left in pending state.


mafrahm (approved):
LGTM

PR #663, opened 2025-03-31
Allow subpaths in external files.

riga:
This PR adds more capabilities to the `BundleExternalFiles` task.

In our analysis, we are evaluating several externally trained ML models that are stored in remote locations and repositories. So far, we added the archives of these repositories as external files, and then unpacked them within `setup` methods of our producers. **However**, one of the repositories in question is exceedingly large, whereas we only need some subdirectories of files (too many files to add them separately as external files though).

This PR introduces a container object `ExternalFile` which is understood by `BundleExternalFiles`. In our example, we would configure an entry like 

```python
config.x.external_files.ml_files = ExternalFile(
    location="https://gitlab.cern.ch/.../-/archive/master/repo-master.tar.gz",
    subpaths={"even": "path/to/even/model/dir", "odd": "path/to/odd/model/dir"},
    version="v3",
)
```

`BundleExternalFiles` will fetch the archive, unpack it, and then only consider given subpaths for bundling. Subpath keys allow convenient access later on (see below). Note that `ExternalFile` is backward compatible to the current single-string or tuple structure.

Then, in our `setup()` we can simply do

```python
bundle = reqs["external_files"]
even_model = load_model(bundle.files.ml_files["even"])
odd_model = load_model(bundle.files.ml_files["odd"])
```

### DIFF
--- a/columnflow/tasks/external.py
+++ b/columnflow/tasks/external.py
@@ -10,15 +10,16 @@ import os
 import time
 import shutil
 import subprocess
+from dataclasses import dataclass, field
 
 import luigi
 import law
 import order as od
 
-from columnflow.types import Sequence
 from columnflow.tasks.framework.base import AnalysisTask, ConfigTask, DatasetTask, wrapper_factory
 from columnflow.tasks.framework.parameters import user_parameter_inst
 from columnflow.util import wget, DotDict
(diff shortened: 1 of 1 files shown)


mafrahm (approved):
LGTM

PR #668, opened 2025-04-04
Make lumi in normalization weight producer configurable.

riga:
This is a small PR that allows making the lumi value used by the norm weight producer configurable.

Some analyses prefer to normalize simulation with 1/fb and then scale with the final lumi value only at plot level. This was not possible before.


mafrahm (approved):
LGTM

PR #669, opened 2025-04-04
Fix config lookup for taf classes mixins.

riga:
The mechanism that matches specific config settings to task parameter defaults, such as

```ini
cf.{SelectEvents,ReduceEvents}__sel_default: htcondor_memory=6.25GB
```

currently does not resolve keys for parameters that are present multiple times like `producers` and `calibrators`.

I added the responsible `get_config_lookup_keys` methods to the two relevant mixins and enabled sequences in the `_dfs_key_lookup`.

In addition, the `pilot` flag is now part of the lookup keys for tasks that support it, i.e., key fragments like `...__pilot_True__...` can be used now to trigger specific default conditions in case the task is in pilot mode (or not).

### DIFF
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -280,7 +280,7 @@ class AnalysisTask(BaseTask, law.SandboxTask):
             items = [
                 parse(key, value)
                 for key, value in law.config.items("resources")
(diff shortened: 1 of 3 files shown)


mafrahm (approved):
LGTM

PR #671, opened 2025-04-07
minor fixes and streamlining

mafrahm:
This PR fixes two minor issues:

- during Plotting, there was an issue with picking the "nominal" shift bin
- missing ReducerClassMixin for the MLModelTrainingMixin + fixing order of mixins

Additionally, I also streamlined the reading of inputs from variables. Right now, the "inputs" aux is ignored when passing an expression via string, however, there are usecases, where one still wants to read custom inputs, e.g. when a collection is updated/defined as part of the HistProducer:

```python
config.add_variable(
    name=f"lepton0_pt",
    expression="Lepton[:, 0].pt",
    aux=dict(
        inputs={"{Electron,Muon}.{pt,eta,phi,mass}"},
    ),
    binning=(40, 0., 400.),
    unit="GeV",
    null_value=EMPTY_FLOAT,
    x_title=f"Lepton 0 $p_{{T}}$",
)
 ```

### DIFF
--- a/columnflow/plotting/plot_functions_1d.py
+++ b/columnflow/plotting/plot_functions_1d.py
@@ -75,7 +75,7 @@ def plot_variable_stack(
         for proc_inst, h in hists.items()
         if proc_inst.is_mc and getattr(proc_inst, "unstack", False)
     }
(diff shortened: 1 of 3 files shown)


riga (approved):
Looks good! Let's merge.

PR #672, opened 2025-04-10
fix plotting with single varied shift

mafrahm:


columnflow/tasks/histograms.py
```diff
@@ -363,7 +363,7 @@ class MergeHistograms(_MergeHistograms):
         CreateHistograms=CreateHistograms,
     )
 
-    invokes_hist_producer = True
+    invokes_hist_producer = False
```
riga, line 367: We can remove the entry altogether.

```suggestion
```

riga (approved):
LGTM! Just one inline change.

PR #674, opened 2025-04-10
Apply new recommendation for egamma calibration

Bogdan-Wiederspan:
This PR enables the application of egamma corrections for run2 and run3, as recommended in [Twiki of JSON Pog](https://twiki.cern.ch/twiki/bin/view/CMS/EgammSFandSSRun3#2022_and_2023_Scale_and_Smearing).

- Due to changes in the correctionlib json more information needed to be given to the config. For this reason multiple conditions are added
- The calculation of up and down smearing was generalized to a more robust version.

### DIFF
--- a/columnflow/calibration/cms/egamma.py
+++ b/columnflow/calibration/cms/egamma.py
@@ -62,13 +62,30 @@ class egamma_scale_corrector(Calibrator):
 
     def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
         """
-        Apply energy corrections to EGamma objects in the events array.
-
-        There are two types of implementations: standard and Et dependent. For Run2 the standard
-        implementation is used, while for Run3 the Et dependent is recommended by the EGammaPog:
-        https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammSFandSSRun3?rev=41 The Et dependendent
-        recipe follows the example given in:
+        Apply energy corrections to EGamma objects in the events array. There are two types of implementations: standard
+        and Et dependent.
+        For Run2 the standard implementation is used, while for Run3 the Et dependent is recommended by the EGammaPog:
+        https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammSFandSSRun3?rev=41
+        The Et dependendent recipe follows the example given in:
         https://gitlab.cern.ch/cms-nanoAOD/jsonpog-integration/-/blob/66f581d0549e8d67fc55420d8bba15c9369fff7c/examples/egmScaleAndSmearingExample.py
+
+        Requires an external file in the config under ``electron_ss``. Example:
+        .. code-block:: python
+            cfg.x.external_files = DotDict.wrap({
+                "electron_ss": "/afs/cern.ch/work/m/mrieger/public/mirrors/jsonpog-integration-120c4271/POG/EGM/2022_Summer22//electronSS_EtDependent.json.gz",  # noqa
+            })
+
+        The pairs of correction set, value and uncertainty type names,  and if a compound method is used should be configured using the :py:class:`EGammaCorrectionConfig` as an
+        auxiliary entry in the config:
+
+        .. code-block:: python
+
+            cfg.x.eec = EGammaCorrectionConfig(
+                correction_set="EGMScale_Compound_Ele_2022preEE",
+                compound=True,
+                value_type="scale",
+                uncertainty_type="escale",
+            )
 
         Derivatives of this base class require additional member variables and functions:
 
@@ -276,6 +293,24 @@ class egamma_resolution_corrector(Calibrator):
         https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammSFandSSRun3?rev=41 The Et dependendent
         recipe follows the example given in:
         https://gitlab.cern.ch/cms-nanoAOD/jsonpog-integration/-/blob/66f581d0549e8d67fc55420d8bba15c9369fff7c/examples/egmScaleAndSmearingExample.py
+
+        Requires an external file in the config under ``electron_ss``. Example:
+        .. code-block:: python
+            cfg.x.external_files = DotDict.wrap({
+                "electron_ss": "/afs/cern.ch/work/m/mrieger/public/mirrors/jsonpog-integration-120c4271/POG/EGM/2022_Summer22/electronSS_EtDependent.json.gz",  # noqa
+            })
+
+        The pairs of correction set, value and uncertainty type names, and if a compound method is used should be configured using the :py:class:`EGammaCorrectionConfig` as an
+        auxiliary entry in the config:
+
+        .. code-block:: python
+
+            cfg.x.eec = EGammaCorrectionConfig(
+                correction_set="EGMSmearAndSyst_ElePTsplit_2022preEE",
+                compound=False,
+                value_type="smear",
+                uncertainty_type="esmear",
+            )
 
         Derivatives of this base class require additional member variables and functions:
 
--- a/columnflow/calibration/cms/egamma.py
+++ b/columnflow/calibration/cms/egamma.py
@@ -63,21 +63,26 @@ class egamma_scale_corrector(Calibrator):
         """
         Apply energy corrections to EGamma objects in the events array.
 
-        This implementation follows the recommendations from the EGamma POG:
-        https://twiki.cern.ch/twiki/bin/view/CMS/EgammSFandSSRun3#Scale_And_Smearings_Example
+        There are two types of implementations: standard and Et dependent. For Run2 the standard
+        implementation is used, while for Run3 the Et dependent is recommended by the EGammaPog:
+        https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammSFandSSRun3?rev=41 The Et dependendent
+        recipe follows the example given in:
+        https://gitlab.cern.ch/cms-nanoAOD/jsonpog-integration/-/blob/66f581d0549e8d67fc55420d8bba15c9369fff7c/examples/egmScaleAndSmearingExample.py
 
-        Derivatives of this base class require additional member variables and
-        functions:
+        Derivatives of this base class require additional member variables and functions:
 
-        - *source_field*: The field name of the EGamma objects in the events array (i.e. `Electron` or `Photon`).
-        - *get_correction_file*: Function to retrieve the correction file, e.g.from the list ,of external files in the current `config_inst`.
+        - *source_field*: The field name of the EGamma objects in the events array (i.e. `Electron`
+            or `Photon`).
+        - *get_correction_file*: Function to retrieve the correction file, e.g.from
+            the list, of external files in the current `config_inst`.
         - *get_scale_config*: Function to retrieve the configuration for the energy correction.
-            This config must be an instance of :py:class:`~columnflow.calibration.cms.egamma.EGammaCorrectionConfig`.
+            This config must be an instance of
+            :py:class:`~columnflow.calibration.cms.egamma.EGammaCorrectionConfig`.
 
-        If no raw pt (i.e., pt before any corrections) is available, use the nominal pt.
-        The correction tool only supports flat arrays, so inputs are converted to a flat numpy view first.
-        Corrections are always applied to the raw pt, which is important if more than one correction is applied in a
-        row. The final corrections must be applied to the current pt.
+        If no raw pt (i.e., pt before any corrections) is available, use the nominal pt. The
+        correction tool only supports flat arrays, so inputs are converted to a flat numpy view
+        first. Corrections are always applied to the raw pt, which is important if more than one
+        correction is applied in a row. The final corrections must be applied to the current pt.
 
         If :py:attr:`with_uncertainties` is set to `True`, the scale uncertainties are calculated.
         The scale uncertainties are only available for simulated data.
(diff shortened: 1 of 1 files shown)


columnflow/calibration/cms/egamma.py
```diff
@@ -29,6 +29,8 @@
 @dataclass
 class EGammaCorrectionConfig:
     correction_set: str = "Scale"
+    compound: bool = False
+    corrector_strings: dict[str] = field(default_factory=dict)
     corrector_kwargs: dict[str, Any] = field(default_factory=dict)
```
riga, line 34: I think the existing `corrector_kwargs` can already handle what `corrector_strings` are supposed to do. Also, it would be good to have a default here.

```suggestion
    compound: bool = False
    corrector_kwargs: dict[str, Any] = field(default_factory=lambda: {"type": ...})
```

columnflow/calibration/cms/egamma.py
```diff
@@ -68,8 +70,8 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
         functions:
 
         - *source_field*: The field name of the EGamma objects in the events array (i.e. `Electron` or `Photon`).
-        - *get_correction_file*: Function to retrieve the correction file, e.g.
-            from the list of external files in the current `config_inst`.
+        - *get_correction_file*: Function to retrieve the correction file, e.g.from the list ,
+        of external files in the current `config_inst`.
```
riga, line 74: Please revert.

columnflow/calibration/cms/egamma.py
```diff
@@ -119,6 +121,10 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
             "gain": gain,
             "r9": r9,
             "run": run,
+            "seedGain": gain,
+            "pt": pt_eval,
+            "AbsScEta": np.abs(sceta),
+            "ScEta": sceta,
```
riga, line 127: Just out of interest:

- Do the new corrections require both AbsScEta *and* ScEta? Or is it different sets that require different inputs?
- If so, are you sure that ScEta is meant to be the not-absolute value? Just to be cautious that there wasn't just a naming issue and they actually meant to use the abs value.
-

columnflow/calibration/cms/egamma.py
```diff
@@ -128,7 +134,7 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
 
         # varied corrections are only applied to MC
         if self.with_uncertainties and self.dataset_inst.is_mc:
-            scale_uncertainties = self.scale_corrector("total_uncertainty", *args)
+            scale_uncertainties = self.scale_corrector.evaluate(self.scale_config.corrector_strings["type"], *args)
```
riga, line 137: With the above suggestion to use `corrector_kwargs`, the first argument would not be needed, but please check :)

columnflow/calibration/cms/egamma.py
```diff
@@ -214,13 +220,17 @@ def setup_func(
...
+        corr_set = load_correction_set(corr_file)
+        if self.scale_config.compound:
+            corr_set = corr_set.compound
+        self.scale_corrector = corr_set[self.scale_config.correction_set]
 
-        # check versions
-        assert self.scale_corrector.version in [0, 1, 2]
+        if not self.scale_config.compound:
+            # check version, does not exist for compound
+            assert self.scale_corrector.version in [0, 1, 2]
```
riga, line 233: Tbh, I'd be fine with removing the version check here altogether.

columnflow/calibration/cms/egamma.py
```diff
@@ -300,52 +310,67 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
...
 
         # the correction tool only supports flat arrays, so convert inputs to flat np view first
-
         sceta = flat_np_view(events[self.source_field].superclusterEta, axis=1)
         r9 = flat_np_view(events[self.source_field].r9, axis=1)
         flat_seeds = flat_np_view(events[self.source_field].deterministic_seed, axis=1)
 
-        # prepare arguments
-        # we use pt as et since there depends in linear (following the recoomendations)
+        # we use pt as et since there depends in linear (following the recommendations)
```
riga, line 317: I don't know what this is supposed to mean. Is the recommendation to use pt? If yes, we should remove the comment

columnflow/calibration/cms/egamma.py
```diff
@@ -300,52 +310,67 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
...
             "eta": sceta,
             "r9": r9,
+            "pt": pt,
             **self.resolution_cfg.corrector_kwargs,
         }
+
         args = tuple(
-            variable_map[inp.name] for inp in self.resolution_corrector.inputs
+            variable_map[inp.name]
+            for inp in self.resolution_corrector.inputs
```
riga, line 332: If you make this change here, then please also do it in the other call func's in this file.

columnflow/calibration/cms/egamma.py
```diff
@@ -300,52 +310,67 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
...
             if inp.name in variable_map
         )
 
         # calculate the smearing scale
-        rho = self.resolution_corrector("rho", *args)
+        # available option are for Et: smear, esmear or escale, e = error
+
+        # TODO : ask Marcel/philip about this
```
riga, line 339: Remove.

columnflow/calibration/cms/egamma.py
```diff
@@ -300,52 +310,67 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
...
+        # NOTE: the presence of the scale uncertainties ("escale") among the "EGMSmearAndSyst",
+        # as mentioned in the example above, allows us to apply them directly to the MC simulation.
```
riga, line 341: "escale" is not necessarily used here, and we already were puzzled by "allows us to apply them directly to the MC simulation" when we read it in the Twiki, so please consider removing this here, or write it more precisely.

columnflow/calibration/cms/egamma.py
```diff
@@ -436,13 +461,16 @@ def setup_func(
...
-        # check versions
-        assert self.resolution_corrector.version in [0, 1, 2]
+        corr_set = load_correction_set(corr_file)
+        if self.resolution_cfg.compound:
+            corr_set = corr_set.compound
+        self.resolution_corrector = corr_set[self.resolution_cfg.correction_set]
+
+        if not self.resolution_cfg.compound:
+            # check versions, does not exist for compound version
+            assert self.resolution_corrector.version in [0, 1, 2]
```
riga, line 473: Same as above.

riga (changes requested):
Thanks for the generalization to eT-dependent calibrations @Bogdan-Wiederspan !

I just had a couple inline comments that should be straight forward to apply.

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: At first glance this would be true, since we use `corrector_kwargs` as dynamic extension of `variable_map`, and the correction and its uncertainty share the same keyword within the corrector I splitted this up. 

But right when I wrote this reply I thought of a way to apply this.

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: - The variable map currently is a superset of old and new method. The old version did not use AbsScEta, the new one does.  We could think about having 2 dictionaries and combine them for the sake of readability. 
- Im pretty sure that ScEta is the normal supercluster eta and Abs is the absolute. I took double check by compare the correctors requires inputs and with the given [example implementation](https://gitlab.cern.ch/cms-nanoAOD/jsonpog-integration/-/blob/master/examples/egmScaleAndSmearingExample.py#L59)

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: I have no strong opinion. Fact is no version check is currently possible with the compound path.
I would maybe leave this, but only because it may come a version check for the compound method.
But im also not sure what "version" in the context of correctionlib actually mean.

columnflow/calibration/cms/egamma.py
```diff
@@ -300,52 +310,67 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
...
+            rho_unc = self.resolution_corrector(self.resolution_cfg.corrector_strings["unc_type"], *args)
+            random_normal_number = functools.partial(ak_random, 0, 1)
+            smearing_func = lambda rng_array, variation: rng_array * variation + 1
+
             smearing_up = (
-                ak_random(
-                    1, rho + rho_unc, flat_seeds,
-                    rand_func=self.deterministic_normal_up,
+                smearing_func(
+                    random_normal_number(flat_seeds, rand_func=self.deterministic_normal),
```
pkausw, line 355: I think this should be the dedicated random function for the up variation to avoid unwanted correlations between the random numbers for the nominal and shifted corrections:

```suggestion
                    random_normal_number(flat_seeds, rand_func=self.deterministic_normal_up),
```

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: We can leave this out, it was just a hint in the twiki why we can use pt instead of et
I cite: 
```
Note that we deal with pt instead of energy in the code below. ... . Scales and smearings are equal for pt and energy as they are directly linearly proportional.
```

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: After sawing the more detailed example, im pretty sure we don't want to combine scale and smear into one step. Is removed.

columnflow/calibration/cms/egamma.py
```diff
@@ -300,52 +310,67 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
...
+                    rho + rho_unc,
                 )
             )
 
             smearing_down = (
-                ak_random(
-                    1, rho - rho_unc, flat_seeds,
-                    rand_func=self.deterministic_normal_down,
+                smearing_func(
+                    random_normal_number(flat_seeds, rand_func=self.deterministic_normal),
```
pkausw, line 367: similar to the up variation, this should be the dedicated gauss for the down variation

```suggestion
                    random_normal_number(flat_seeds, rand_func=self.deterministic_normal_down),
```

pkausw (changes requested):
Thanks for this PR @Bogdan-Wiederspan ! Apart from the comments @riga already gave, I also have two additional requests - should be easy to implement.

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: I removed this line

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: I looked this up. The descriptor says that eta in the standard method json is actually the super clustereta...
How this will definitly break in near future *whistle*

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: As discussed offline - we added concrete arguments instead of using corrector_kwargs. This is more explicit and forces the user to hand these values.

columnflow/calibration/cms/egamma.py
Bogdan-Wiederspan: same as above

columnflow/calibration/cms/egamma.py
```diff
@@ -59,24 +62,45 @@ def get_scale_config(self) -> EGammaCorrectionConfig:
...
+        The pairs of correction set, value and uncertainty type names,  and if a compound method is used should be configured using the :py:class:`EGammaCorrectionConfig` as an
+        auxiliary entry in the config:
+
+        .. code-block:: python
+
+            cfg.x.eec = EGammaCorrectionConfig(
+                correction_set="EGMScale_Compound_Ele_2022preEE",
+                compound=True,
+                value_type="scale",
+                uncertainty_type="escale",
```
riga, line 87:
```suggestion
                value_type="scale",
                uncertainty_type="escale",
                compound=True,
```

columnflow/calibration/cms/egamma.py
```diff
@@ -261,11 +288,31 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
...
+            })
+
+        The pairs of correction set, value and uncertainty type names, and if a compound method is used should be configured using the :py:class:`EGammaCorrectionConfig` as an
+        auxiliary entry in the config:
+
+        .. code-block:: python
+
+            cfg.x.eec = EGammaCorrectionConfig(
+                correction_set="EGMSmearAndSyst_ElePTsplit_2022preEE",
+                compound=False,
```
riga, line 310:
```suggestion
```

columnflow/calibration/cms/egamma.py
```diff
@@ -300,52 +347,61 @@ def call_func(self, events: ak.Array, **kwargs) -> ak.Array:
             )
 
         # the correction tool only supports flat arrays, so convert inputs to flat np view first
-
         sceta = flat_np_view(events[self.source_field].superclusterEta, axis=1)
         r9 = flat_np_view(events[self.source_field].r9, axis=1)
         flat_seeds = flat_np_view(events[self.source_field].deterministic_seed, axis=1)
 
+        pt = flat_np_view(events[self.source_field].rawPt, axis=1)
```
riga, line 354:
```suggestion
        pt = flat_np_view(events[self.source_field].rawPt, axis=1)
```

riga (approved):

riga:
Ok, looks like all discussions are resolved, let's merge.

PR #675, opened 2025-04-11
implement own errorbar calculation

nprouvost:
Through this PR, the error bars calculated by mplhep are not used anymore, yerr is always set in the draw functions if not already defined before. The option to use an unweighted asymmetric poisson has been added on top of the weighted poisson already defined. The default for data has been set to the unweighted poisson. The norm for the plots is now being applied after the error calculation such that the correct Poisson error is calculated, and then scaled. 

Up to debate: I raised errors for the draw hist and draw errorbars functions in the case that no yerr has been set and the accumulator type of the histogram is not WeightedSum (should never happen for the default Histogram type in columnflow), since the errors would be calculated by mplhep without further indication in such a case. Would a warning be preferred?

### DIFF
--- a/columnflow/plotting/plot_all.py
+++ b/columnflow/plotting/plot_all.py
@@ -19,6 +19,7 @@ from columnflow.plotting.plot_util import (
     get_cms_label,
     remove_label_placeholders,
     apply_label_placeholders,
+    calculate_error,
 )
 
 hist = maybe_import("hist")
@@ -183,6 +184,7 @@ def draw_stack(
             # solution: transform norm -> [norm]*len(h)
             h = hist.Stack(*[i / norm for i in h])
 
+    # draw only the stack, no error bars/bands with stack = True
     defaults = {
         "ax": ax,
         "stack": True,
@@ -196,19 +198,37 @@ def draw_hist(
     ax: plt.Axes,
     h: hist.Hist,
     norm: float | Sequence | np.ndarray = 1.0,
+    error_type: str = "variance",
     **kwargs,
 ) -> None:
+    assert error_type in ("variance", "poisson_unweighted", "poisson_weighted")
+
     if kwargs.get("color", "") is None:
         # when color is set to None, remove it such that matplotlib automatically chooses a color
         kwargs.pop("color")
 
-    h = h / norm
     defaults = {
         "ax": ax,
         "stack": False,
         "histtype": "step",
     }
     defaults.update(kwargs)
+    if "yerr" not in defaults and not (h.storage_type.accumulator is hist.accumulators.WeightedSum):
+        raise TypeError(
+            "Error bars calculation only implemented for histograms with storage type WeightedSum "
+            "either change the Histogram storage_type or set yerr manually",
+        )
+    if "yerr" not in defaults and (h.storage_type.accumulator is hist.accumulators.WeightedSum):
+        yerr = calculate_error(h, error_type)
+        # normalize yerr to the histogram = error propagation on standard deviation
+        yerr = yerr / norm
+        # replace inf with nan for any bin where norm = 0 and calculate_error returns a non zero value
+        if np.any(np.isinf(yerr)):
+            yerr[np.isinf(yerr)] = np.nan
+        defaults["yerr"] = yerr
+
+    h = h / norm
+
     h.plot1d(**defaults)
 
 
@@ -216,11 +236,14 @@ def draw_profile(
     ax: plt.Axes,
     h: hist.Hist,
     norm: float | Sequence | np.ndarray = 1.0,
+    error_type: str = "variance",
     **kwargs,
 ) -> None:
     """
     Profiled histograms contains the storage type "Mean" and can therefore not be normalized
     """
+    assert error_type in ("variance", "poisson_unweighted", "poisson_weighted")
+
     if kwargs.get("color", "") is None:
         # when color is set to None, remove it such that matplotlib automatically chooses a color
         kwargs.pop("color")
@@ -231,6 +254,13 @@ def draw_profile(
         "histtype": "step",
     }
     defaults.update(kwargs)
+    if "yerr" not in defaults and not (h.storage_type.accumulator is hist.accumulators.WeightedSum):
+        raise TypeError(
+            "Error bars calculation only implemented for histograms with storage type WeightedSum "
+            "either change the Histogram storage_type or set yerr manually",
+        )
+    if "yerr" not in defaults and (h.storage_type.accumulator is hist.accumulators.WeightedSum):
+        defaults["yerr"] = calculate_error(h, error_type)
     h.plot1d(**defaults)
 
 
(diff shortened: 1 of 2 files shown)


columnflow/plotting/plot_util.py
```diff
@@ -1016,3 +1019,58 @@ def remove_label_placeholders(
...
+        - 'variance': the plotted error is the square root of the variance for each bin
+        - 'poisson_unweighted': the plotted error is the poisson error for each bin
+        - 'poisson_weighted': the plotted error is the poisson error for each bin, weighted by the variance
+    """
+    logger = law.logger.get_logger(__name__)
+
+    # determine the error type
+    if error_type == "variance":
+        yerr = hist.view().variance ** 0.5
+    elif error_type.startswith("poisson"):
```
riga, line 1040: Can you make the check explicit here?

```suggestion
    elif error_type in {"poisson_unweighted", "poisson_weighted"}:
```

columnflow/plotting/plot_util.py
```diff
@@ -1016,3 +1019,58 @@ def remove_label_placeholders(
...
+        # compute asymmetric poisson confidence interval
+        from hist.intervals import poisson_interval
+
+        if error_type == "poisson_unweighted":
+            variances = None
+        elif error_type == "poisson_weighted":
+            variances = hist.view().variance
+        else:
+            raise ValueError(f"unknown error type '{error_type}'")
```
riga, line 1049: With the above change, this can be simplified to

```suggestion
        variances = hist.view().variance if error_type == "poisson_weighted" else None
```

columnflow/plotting/plot_util.py
```diff
@@ -1016,3 +1019,58 @@ def remove_label_placeholders(
         sel = "[A-Z0-9]+"
 
     return re.sub(f"__{sel}__", "", label)
+
+
+def calculate_error(
```
riga, line 1024: We generally use this to compute statistical errors, right? If so, we might want to rename this accordingly, e.g. to `calculate_stat_error`. Just "error" sounds rather generic.

columnflow/plotting/plot_all.py
```diff
@@ -196,31 +198,52 @@ def draw_hist(
     ax: plt.Axes,
     h: hist.Hist,
     norm: float | Sequence | np.ndarray = 1.0,
+    error_type: str = "variance",
     **kwargs,
 ) -> None:
+    assert error_type in ("variance", "poisson_unweighted", "poisson_weighted")
```
riga, line 204:
```suggestion
    assert error_type in {"variance", "poisson_unweighted", "poisson_weighted"}
```

columnflow/plotting/plot_all.py
```diff
@@ -196,31 +198,52 @@ def draw_hist(
...
         "stack": False,
         "histtype": "step",
     }
     defaults.update(kwargs)
+    if "yerr" not in defaults and not (h.storage_type.accumulator is hist.accumulators.WeightedSum):
+        raise TypeError(
+            "Error bars calculation only implemented for histograms with storage type WeightedSum "
+            "either change the Histogram storage_type or set yerr manually",
+        )
+    if "yerr" not in defaults and (h.storage_type.accumulator is hist.accumulators.WeightedSum):
```
riga, line 221:
```suggestion
    if "yerr" not in defaults:
        if h.storage_type.accumulator is not hist.accumulators.WeightedSum:
            raise TypeError(
                "Error bars calculation only implemented for histograms with storage type WeightedSum "
                "either change the Histogram storage_type or set yerr manually",
            )
```

columnflow/plotting/plot_all.py
```diff
@@ -196,31 +198,52 @@ def draw_hist(
...
     ax: plt.Axes,
     h: hist.Hist,
     norm: float | Sequence | np.ndarray = 1.0,
+    error_type: str = "variance",
     **kwargs,
 ) -> None:
     """
     Profiled histograms contains the storage type "Mean" and can therefore not be normalized
     """
+    assert error_type in ("variance", "poisson_unweighted", "poisson_weighted")
```
riga, line 245:
```suggestion
    assert error_type in {"variance", "poisson_unweighted", "poisson_weighted"}
```

columnflow/plotting/plot_all.py
```diff
@@ -231,38 +254,51 @@ def draw_profile(
         "histtype": "step",
     }
     defaults.update(kwargs)
+    if "yerr" not in defaults and not (h.storage_type.accumulator is hist.accumulators.WeightedSum):
+        raise TypeError(
+            "Error bars calculation only implemented for histograms with storage type WeightedSum "
+            "either change the Histogram storage_type or set yerr manually",
+        )
+    if "yerr" not in defaults and (h.storage_type.accumulator is hist.accumulators.WeightedSum):
```
riga, line 262:
```suggestion
    if "yerr" not in defaults:
        if h.storage_type.accumulator is not hist.accumulators.WeightedSum:
            raise TypeError(
                "Error bars calculation only implemented for histograms with storage type WeightedSum "
                "either change the Histogram storage_type or set yerr manually",
            )
```

columnflow/plotting/plot_all.py
```diff
@@ -231,38 +254,51 @@ def draw_profile(
...
 
 
 def draw_errorbars(
     ax: plt.Axes,
     h: hist.Hist,
     norm: float | Sequence | np.ndarray = 1.0,
+    error_type: str = "variance",
     **kwargs,
 ) -> None:
+    assert error_type in ("variance", "poisson_unweighted", "poisson_weighted")
```
riga, line 274:
```suggestion
    assert error_type in {"variance", "poisson_unweighted", "poisson_weighted"}
```

columnflow/plotting/plot_all.py
```diff
@@ -231,38 +254,51 @@ def draw_profile(
...
         "elinewidth": 1,
     }
     defaults.update(kwargs)
+
+    if "yerr" not in defaults and not (h.storage_type.accumulator is hist.accumulators.WeightedSum):
+        raise TypeError(
+            "Error bars calculation only implemented for histograms with storage type WeightedSum "
+            "either change the Histogram storage_type or set yerr manually",
+        )
+    if "yerr" not in defaults and (h.storage_type.accumulator is hist.accumulators.WeightedSum):
```
riga, line 293:
```suggestion
    if "yerr" not in defaults:
        if h.storage_type.accumulator is not hist.accumulators.WeightedSum:
            raise TypeError(
                "Error bars calculation only implemented for histograms with storage type WeightedSum "
                "either change the Histogram storage_type or set yerr manually",
            )
```

columnflow/plotting/plot_util.py
```diff
@@ -1016,3 +1019,58 @@ def remove_label_placeholders(
...
+    error_type: str,
+) -> dict:
+    """
+    Calculate the error to be plotted for the given histogram *hist*.
+    Supported error types are:
+        - 'variance': the plotted error is the square root of the variance for each bin
+        - 'poisson_unweighted': the plotted error is the poisson error for each bin
+        - 'poisson_weighted': the plotted error is the poisson error for each bin, weighted by the variance
+    """
+    logger = law.logger.get_logger(__name__)
```
riga, line 1035: Can you make this logger a module citizen?

riga (changes requested):
Hey @nprouvost ,

thanks for solving this! I just have minor coding suggestions inline below.

riga (approved):
Looks good to me now!

@mafrahm Do you want to have another look?

columnflow/plotting/plot_all.py
```diff
@@ -231,38 +254,51 @@ def draw_profile(
...
+            )
+        defaults["yerr"] = calculate_stat_error(h, error_type)
     h.plot1d(**defaults)
 
 
 def draw_errorbars(
     ax: plt.Axes,
     h: hist.Hist,
     norm: float | Sequence | np.ndarray = 1.0,
+    error_type: str = "variance",
```
mafrahm, line 271: Since we mainly use this method to draw the data errorbars, using the corresponding `error_type` as default might be preferable

```suggestion
    error_type: str = "poisson_unweighted",
```

mafrahm (approved):
just one inline comment, LGTM otherwise

PR #676, opened 2025-04-22
parton shower weights

JulesVandenbroeck:
This pull request introduced a producer to produce parton shower weights. Initial State Radiation (ISR) and Final State Radiation (FSR) variations are produced as weights `isr_weight_up/down` and  `fsr_weight_up/down`  respectively. Their weights should be normalized in an analysis, similar to pdf variation weights.

All work in this pull request was done by @maadcoen


columnflow/production/cms/parton_shower.py
```diff
@@ -0,0 +1,42 @@
+from columnflow.production import producer, Producer
+from columnflow.util import maybe_import, DotDict
+from columnflow.production.cms.scale import set_ak_column_f32
```
riga, line 3: Just for consistency with other producer files.

```suggestion
# coding: utf-8

"""
Producers for storing parton shower weights.
"""

from __future__ import annotations

from columnflow.production import producer, Producer
from columnflow.columnar_util import set_ak_column
from columnflow.util import maybe_import, DotDict
```

columnflow/production/cms/parton_shower.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+
+    # setup nominal weights
+    ones = np.ones(len(events), dtype=np.float32)
+    events = set_ak_column_f32(events, "fsr_weight", ones)
+    events = set_ak_column_f32(events, "isr_weight", ones)
+
+    # now loop through the names and save the respective normalized PSWeights
+    for column, index in self.indices.items():
+        events = set_ak_column_f32(events, column, events.PSWeight[:, index])
```
riga, line 41: Just using the default `set_ak_column` here.

```suggestion
    # setup nominal weights
    ones = np.ones(len(events), dtype=np.float32)
    events = set_ak_column(events, "fsr_weight", ones)
    events = set_ak_column(events, "isr_weight", ones)

    # now loop through the names and save the respective normalized PSWeights
    for column, index in self.indices.items():
        events = set_ak_column(events, column, events.PSWeight[:, index])

```

riga (changes requested):
Hey @JulesVandenbroeck ,

thanks for that! Many analyses will profit from that I reckon.

I just have two minor inline code suggestions, and one additional request: could you add the "parton_shower" entry to the columnflow-cms specific `production_modules` in https://github.com/columnflow/columnflow/blob/384da2054dbb4e0fbd4eee43894974f0ffb613f0/analysis_templates/cms_minimal/law.cfg#L32?

With that, we can merge right away.

riga (approved):
Great, let's merge!

PR #679, opened 2025-04-30
Fixes for docs (pdf figures not displayed)

jolange:
Mainly a fix to display the plots in the figures. Currently, the docs on readthedocs do not show the plots, but only links to the PDFs (maybe the alt text):
image

Although I heard rumors that for some person/os/browser combinations the PDFs are shown, we should resort so something that works for the average viewer.

I converted all plot PDFs to additional SVGs. If readthedocs would use the Makefile, one could have made a rule for automatic conversion, but I think this is good enough, because these plots have not changed too often. 
For the figure directives, I changed the file endings to wildcards, so sphinx uses what is best for the export format (e.g. latex will still use pdf).
image

(+I added an lfs filter for svg files. The file `plots/user_guide/cf_plot.svg` was on lfs, already, but I did not find any filter for that -- probably somebody uploaded it to lfs, but did not commit the corresponding `.gitattributes`.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,4 @@
 *.jpeg filter=lfs diff=lfs merge=lfs -text
 *.root filter=lfs diff=lfs merge=lfs -text
 *.ico filter=lfs diff=lfs merge=lfs -text
+*.svg filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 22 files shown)


riga (approved):
Great, thanks @jolange !

PR #682, opened 2025-05-09
allow passing mask to apply JER smearing only to a subset of jets

mafrahm:
This mini PR allows passing a mask to the `jer` Calibrator to disable the stochastic smearing for a subset of jets.

I also added the `jer_horn_handling` that disables the stochastic smearing in `2.5 < |eta| < 3.0`, following the latest CMS recommendations: https://cms-jerc.web.cern.ch/Recommendations/#note-25eta30

Additionally, I also fixed an issue with the `msoftdrop` Producer similar to #603 (padding to 0 when `ak.max` returns None)

### DIFF
--- a/columnflow/calibration/cms/jets.py
+++ b/columnflow/calibration/cms/jets.py
@@ -958,7 +958,8 @@ def jer(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
 
 
 jer_horn_handling = jer.derive("jer_horn_handling", cls_dict={
-    "stochastic_smearing_mask": lambda self, jets: (abs(jets.eta) < 2.6) | (abs(jets.eta) > 3.1),
+    # source: https://cms-jerc.web.cern.ch/Recommendations/#note-25eta30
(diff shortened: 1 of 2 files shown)


columnflow/calibration/cms/jets.py
```diff
@@ -951,6 +957,11 @@ def jer(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
     return events
 
 
+jer_horn_handling = jer.derive("jer_horn_handling", cls_dict={
```
riga, line 960: Could you add a reference (Twiki or else) where you took these values from?

riga (changes requested):
Looks good to me!

I just a single comment regarding the reference for excluding the 2.6 - 3.1 region.

columnflow/calibration/cms/jets.py
mafrahm: apparently, the recommended region is 2.5-3.0
I added the source and updated the Calibrator

PR #685, opened 2025-05-26
set scale_factor to 1 instead of 0

mafrahm:
when using `scale="stack"`, but not passing any stacked histograms, histograms will be scaled to 0, resulting in empty plots. With this fix, the `scale="stack"` argument will basically be ignored in such cases, setting the scale factor to 1 instead.


riga (approved):
Thanks, looks good!

PR #686, opened 2025-05-26
allow skipping preparation_producer in MLEvaluation

mafrahm:
At the moment, the `preparation_producer` will always be invoked during `MLEvaluation`. This PR allows disabling the `preparation_producer` via the `ml_model.preparation_producer_in_ml_evaluation` attribute.


riga:
Thanks for opening this! Could you check why the CI seems to fail?

PR #692, opened 2025-05-28
use nominal for processes that don't have a requested shift

maadcoen:
The current code checks if the plot shifts are defined for a given process and only then the nominal is used if that shift is not found in the histogram. I'd suggest to do this always, whether the shift is defined for the process or not. Cause currently, the task `cf.PlotShiftedVariables1D` creates weird plots like the one below, which shows in the systematics band the renormalization and factorization scale variations. THese aren't defined for the QCD contribution, whcih means that the QCD constribution is not counted in the up and down variation. Both variations, therefore undershoot the nominal prediction.

I am wondering why this requirement of the shift being defined for the process is in there in the first place? I might be missing something?

*edit (Marcel)*: Removed the picture from public resource :)

### DIFF
--- a/columnflow/tasks/plotting.py
+++ b/columnflow/tasks/plotting.py
@@ -159,7 +159,7 @@ class PlotVariablesBase(_PlotVariablesBase):
         plot_shift_names = set(shift_inst.name for shift_inst in plot_shifts)
 
         # get assignment of processes to datasets and shifts
-        config_process_map, process_shift_map = self.get_config_process_map()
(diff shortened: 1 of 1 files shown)


riga:
@maadcoen Thanks for the PR, I'm going to have a look next week as well. Fyi: I removed the image, but feel free to add a link to the image on e.g. EOS/CernBox, where you can control how's allowed to access it.

riga:
This should be fixed by now, but feel free to reopen if this is still the case.

PR #697, opened 2025-06-16
Fix task pinning

riga:
This PR fixes the pinning of output locations, versions and resources of tasks through the law config file.

## Current state

So far, we are able to pin-point the exact task through a "__"-separated key in the respective config section, such as

```ini
[versions]
22pre_v14__cf.ProduceColumns__prod_default: prod9
22pre_v14__cf.ProduceColumns: prod10
```

which leads to all cf.ProduceColumns tasks that run with the `22pre_v14` config to use a default version `prod10` unless the producer is `default` in which case it uses `prod9`. The keys also allow for using patterns and regular expressions for more complex matching.

There is a slight inconsistency in the way the key is constructed (which is the crucial point later on): The producer is prefixed by `prod_` whereas config, task, and also the analysis name (if provided) are not prefixed by anything (**A**). This was usually okay, since we imposed an order for the matching of key fragments (**B**) such that config has to be placed after the analysis but before the task, which itself can be followed by dataset, shift, calibrator, selector, reducer, producer and histogram producer names. The former 5 fragments are **not** prefixed by anything, but the 5 latter ones are (e.g. `calib_`, `sel_`, etc.).

## Issue

However, it turns out that this inconsistency (**A**) **AND** the order of key fragments (**B**) becomes a huge problem when e.g. negating expressions. For example, when the versions for all tasks for **all configs except** for a specific one, say `22post_v14`, should be pinned, one would do

```ini
[versions]
^(?!22post_v14)$__cf.*: prod9
```

(using regex markers `^$` and a negative lookahead `(?!...)`). This first fragment actually **matches the analysis name first** rather than the config because it comes first in the matching order (**B**) and there is no prefix like `cfg_` to further help referring to the config name (**A**). For producers, this would be simpler, since the `prod_` prefix could be used for explicitly refer to the producer fragment like `^prod_(?!...)$`.

As a result, negation and likely other non-exclusive patterns are just not possible. However, this would be a powerful feature as it would avoid extreme verbosity in the configuration (think of pinning the version for tasks with datasets **other than** DY... you would have to list all of them, since `__^(?!dy_.+)$__` would not work; it would match the analysis name fragment already, same as above).

## Fix

This PR fixes both **A** and **B**:
- **A**: All key fragments have prefixes now. From the updated docs:
  1. analysis name, prefixed by `ana_`
  2. config name, prefixed by `cfg_`
  3. task family, prefixed by `task_`
  4. dataset name, prefixed by `dataset_`
  5. shift name, prefixed by `shift_`
  6. calibrator name, prefixed by `calib_`
  7. selector name, prefixed by `sel_`
  8. reducer name, prefixed by `red_`
  9. producer name, prefixed by `prod_`
  10. hist producer name, prefixed by `hist_`
- **B**: The matching order is no longer important, so defining matching keys no longer requires looking up and memorizing the order given in the docs. This was anyway not very intuitive and quite error-prone.
- All docs are updated accordingly.

With that, the example above would become

```ini
[versions]
^cfg_(?!22post_v14)$__task_cf.*: prod9
```

(note the `cfg_` and `task_` prefixes) and is perfectly working.

## Backwards compatibility

The most frequent usecase is likely something as simple as

```ini
[outputs]
cf.CalibrateEvents: wlcg
```

Now, this would need to be changed to

```ini
[outputs]
task_cf.CalibrateEvents: wlcg
```

However, to be backwards compatible, for a limited time the task fragment *can* but *does not have to* start with the `task_` prefix. There is a deprecation warning in case it is found to be missing though. With that, this change is hopefully not breaking user code.

### DIFF
--- a/analysis_templates/cms_minimal/law.cfg
+++ b/analysis_templates/cms_minimal/law.cfg
@@ -98,8 +98,8 @@ lfn_sources: wlcg_fs_infn_redirector, wlcg_fs_global_redirector
 # output locations per task family
 # the key can consist of multple underscore-separated parts, that can each be patterns or regexes
 # these parts are used for the lookup from within tasks and can contain (e.g.) the analysis name,
-# the config name, the task family, the dataset name, or the shift name
-# (see AnalysisTask.get_config_lookup_keys() - and subclasses - for the exact order)
+# the config name, the task family, the dataset name, or the shift name, for more info, see
+# https://columnflow.readthedocs.io/en/latest/user_guide/best_practices.html#selecting-output-locations
 # values can have the following format:
 # for local targets   : "local[, LOCAL_FS_NAME or STORE_PATH][, store_parts_modifier]"
 # for remote targets  : "wlcg[, WLCG_FS_NAME][, store_parts_modifier]"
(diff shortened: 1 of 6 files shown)


mafrahm (approved):
LGTM

PR #698, opened 2025-06-16
Fix requirements for MLEvaluation

pkausw:
In the current version, the `MLEvaluation` task simply forwards the `configs` argument when requesting `MLTraining`. While this is in principle not wrong, it can have unintended behavior when the MLModel instance is trained with a different set of configs than the current evaluation needs.

This PR fixes this by extracting the configs from the `used_datasets` dictionary provided by the `ml_model_inst`. This is the same logic as used in the `MLTraining` requirements and thus resolves the desired training always correctly.


pkausw:
Thanks @mafrahm  for the fast reply! Indeed, I overlooked the `training_configs` function and everything works as intended. I'll close the PR and delete the branch :+1:

PR #702, opened 2025-06-25
Improve categorizer calls in category_ids producer.

riga:
This PR targets the common category ids producer.

Categories allow defining multiple categorizers in a sequence whose produced masks are AND combined to form overall category assignments. We make heavy use of that feature in our analysis in a way where the same categorizers are used by many (many) categories but in different combinations.

In the current implementation, each categorizer is called multiple times, with their responses likely to be cached. However, even this cache lookup creates quite an overhead for us (some categorizers are called ~250 times).

With this PR, all unique categorizers are called first and their masks are stored. In a second step, categories are then assigned by looking up the previously evaluated masks, drastically reducing runtime in our case.

### DIFF
--- a/columnflow/production/categories.py
+++ b/columnflow/production/categories.py
@@ -6,6 +6,9 @@ Column production methods related defining categories.
 
 from __future__ import annotations
 
+import functools
+import operator
+
 import law
 
 from columnflow.categorization import Categorizer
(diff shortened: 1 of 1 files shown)


mafrahm (approved):
LGTM

PR #703, opened 2025-06-25
Add ml model task pinning.

riga:
This PR adds the choice of the `--ml-model` or `--ml-models` to the list of available keys for pinning task output locations, resources and versions.

The fragment should be prefixed by `ml_` as also added to the documentation. With that, a respective entry in the config file could look like

```ini
[versions]

task_cf.CreateHistograms__ml_RegNet: prod10
```


riga:
I think the gh diff might be misleading, the `MLModelsMixin` already have that :+1:

PR #707, opened 2025-07-02
Make merging chunk size configurable.

riga:
This PR extends the previously added possibility to control the chunk size of events when merging parquet files.

Before, this was only the case in `MergeReducedEvents` and only with a fixed row group size. This PR introduces the same concept to all tasks that perform a post-run merging (calibration, selection, ...) and makes the chunk size configurable through the law.cfg via `[analysis] merging_row_group_size`.


mafrahm (approved):
LGTM

PR #709, opened 2025-07-03
Feature/histogram user multiconfig

mafrahm:
This PR
- makes the HistogramsUserBase tasks compatible with MultiConfig
- updates the inspection tasks accordingly. I also moved this histogram slicing after the merging of histograms over all datasets to make things a bit more efficient, hopefully. The merging of histograms here is much less sophisticated compared to our default plotting task, e.g. the mapping of processes to datasets and the handling of missing shifts is not really considered.

### DIFF
--- a/columnflow/tasks/framework/histograms.py
+++ b/columnflow/tasks/framework/histograms.py
@@ -158,6 +158,7 @@ class HistogramsUserSingleShiftBase(
         return reqs
 
     def requires(self):
+        datasets = [self.datasets] if self.single_config else self.datasets
         return {
             config_inst.name: {
                 d: self.reqs.MergeHistograms.req_different_branching(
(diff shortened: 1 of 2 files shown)


riga (changes requested):
Looks good to me for most parts.

However, I have one comment on the input caching. `cached_property` can have some side effects in the context of tasks. Could you

- remove it (together with the import),
- accept `inputs` as the first argument in `load_histogram`, and
- add `inputs = self.input()` in `run()` and pass it to `load_histogram`?

This avoid caching, and in case the requirements are defined differently in inheriting classes (e.g. by also requiring a different task, which is typically structured by returning `{"default_reqs": super().requires(), "custom_reqs": ...}`), the user can pass `inputs["default_reqs"]`  in this case.

PR #710, opened 2025-07-15
Fix/asymmetric syst unc

mafrahm:
Before, the `shift_up_diff` and `shift_down_diff` were calculated w.r.t. the baseline, which might contain histograms of processes that are not present in the `shift_stacks` (e.g. when requesting top_pt shift source and non-top datasets as part of the stack).

With this change, we calculate the difference only from histograms that are present in the `syst_hists` sequence.


riga (approved):
Looks good, nice catch!

PR #712, opened 2025-07-16
check and remove overlaping processes

aalvesan:
Add a check for overlaping between the processes added to the config. If it exists, remove the chid processes in favor of parent processes.


riga (approved):
Awesome, looks good to me!

PR #714, opened 2025-07-18
inference model caching

mafrahm:
cache inference model


mafrahm (commented):
Looks good to me, thanks for generalizing!

riga (approved):
Ok, let's go for it then.

PR #718, opened 2025-07-30
Generalize normalization weight producer.

riga:
This PR updates the normalization weight producer, in particular its stitching algorithm to compute branching ratios from the parent/child structure of underlying processes and available, overlapping datasets.

**Before**, the algorithm made assumptions about the process structure in the sense that the sub processes for which branching ratios need to be calculated are in fact *leaf* processes. In general, this must not always be the case. In addition, the algorithm assumed that there are so-called "lowest level datasets", i.e., datasets that contain a specific leaf process and nothing else. In the latest addition to our analysis, our stitching requirements became far more complex and none of these assumptions actually hold.

**Now**, with the updates in this PR, none of these assumptions are necessary. The algorithm to identify the statistically most precise branching ratio is entirely rewritten, thus the many code changes. However, it is able to handle even complex cases where there are multiple possible ways in the dataset/process dependency tree (a DAG) to compute a specific BR, and it builds all of them and identifies the most precise solution. The actual producer part (identifying processes and assigning weights from a lookup table) is not really affected, but just the filling of the LUT.

@haddadanas

Reasons for draft mode:

- [x] I'm testing some extreme edge cases, but they should be finalized by the end of the day.
- [x] I want to include a per-dataset weight sum normalization, to avoid edge cases where the numeric domain of weights varies between datasets that should be stitched.
- [x] There should be a warning raised in case multiple possible BR solutions exist, but they actually seem incompatible (e.g. by checking if their stat. pull is > 2).

### DIFF
--- a/columnflow/production/normalization.py
+++ b/columnflow/production/normalization.py
@@ -227,7 +227,7 @@ def get_br_from_inclusive_datasets(
         if (rel_unc := best_br(sn.UP, unc=True, factor=True)) > 0.1:
             logger.warning(
                 f"large error on the branching ratio of {rel_unc * 100:.2f}% for process '{process_inst.name}' "
-                f"({process_inst.id}), calculated along\n   {path_repr(best_br_path, best_dag_path)}",
+                f"({process_inst.id}), calculated along\n  {path_repr(best_br_path, best_dag_path)}",
             )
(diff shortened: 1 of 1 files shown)


haddadanas (approved):
LGTM 👍

PR #719, opened 2025-07-30
update met_phi Calibrator to new format

mafrahm:
This PR updates our met_phi Calibrator to the newest corrections available.

The structure of corrections changes such that there is now one central correction set that contains identifier whether to correct met pt or phi and whether one corrects MC or data.
To keep backwards compatibility with older versions, I am still loading two corrections (pt, phi), but we could remove this if it is not necessary anymore.

I also added a `MetPhiConfig` that contains all the configuration entries related to this Calibrator. In this way, users do not need to derive new Calibrators just to change e.g. the `met_name` etc.

I also made it possible to configure, which met variations (e.g. `pt_stat_xup`) with the `variables_config` dictionary.

I also tried to keep backwards compatibility to the older format where only the `name_tmpl` was passed, but I did not test how well this works with older versions of the met phi corrections.

corrections can be found e.g. [here](https://gitlab.cern.ch/cms-nanoAOD/jsonpog-integration/-/blob/master/POG/JME/2022_Summer22EE/met_xyCorrections_2022_2022EE.json.gz?ref_type=heads)

### DIFF
--- a/columnflow/calibration/cms/met.py
+++ b/columnflow/calibration/cms/met.py
@@ -4,7 +4,11 @@
 MET corrections.
 """
 
+from __future__ import annotations
+
 import law
+
+from dataclasses import dataclass
 
 from columnflow.calibration import Calibrator, calibrator
 from columnflow.util import maybe_import, load_correction_set, DotDict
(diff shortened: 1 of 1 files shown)


riga:
I made the following minor adjustments:

- Renamed `MetPhiConfig` -> `METPhiConfig` (for consistency)
- Removed commented-out lines
- Applied the mask to all inputs as I was getting shape conflicts.
- Changed the `_uncorrected` suffix to `_xy_uncorrected` as there is already a similar suffix being used by JEC/JER.

PR #726, opened 2025-09-17
Remove year from intrinsic btag weight names.

riga:
This PR fixes the column names of stored CMS btag weights (one of our earliest common producers).

In particular, the old implementation added a `_{year}` suffix to the `{hf,lf}stats{1,2}` columns, as a per-era decorrelation is recommended by BTV. However, we should never encoded potential volatile recommendations on systematic shifts into column names as there are better means to do (shift-dependent aliases, name changes on inference model level, ...). Also, `_{year}` is ambiguous for the two eras in both 2022 and 2023, so it was not even correct in the first place.

This PR removes the suffix.

### DIFF
--- a/columnflow/production/cms/btag.py
+++ b/columnflow/production/cms/btag.py
@@ -248,18 +248,17 @@ def btag_weights(
         events = add_weight("central", None, self.weight_name)
         for syst_name, col_name in self.btag_uncs.items():
             for direction in ["up", "down"]:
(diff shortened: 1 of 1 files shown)


mafrahm (approved):
LGTM.

Only possible downside might be the transition phase from one column name to another, but I suppose one could get around that by having two column alises for each of the `{lf,hf}stats` uncertatinty

PR #728, opened 2025-09-24
Sequentialize and optimize datacard writing.

riga:
This is a crucial PR that optimizes the way that cms datacards are written.

### Our workflow / situation

When writing datacards with all systematic shifts, a lot of `{Create,Merge}Histograms` tasks are processed, and the resulting n-D histogram of `MergeShiftedHistograms` is consequently large. In our workflow we perform a late binning optimization which requires us to use 2-5k bins on the variable axis.

However, with about 100 variations, all the categories, sub-processes due to stitching, etc., the required memory to load all outputs of `MergeShiftedHistograms` inside the `CreateDatacards` task is in the order of 20-30GB.

### The current issue

`CreateDatacards` is a workflow that produces each datacard category in a new branch, i.e., a separate task. In case of - say - 10 categories, the memory mentioned above is exactly 10 times as large as roughly the same in-memory data is needed. This is prohibitively large.

### The fix

This PR sequentializes the datacard creation again and also cleans up some smaller code issues that I encountered along the way. Now, in our typical ("worst") case, data loading takes about 3 minutes (overhead) with all possible information loaded into memory **once**, and each datacard takes another 15 seconds per category (3min + 10 * 15s = 5.5min). This is an acceptable ratio compared to the situation before which effectively forced us to run the workflow in sequence with a single worker anyways due to the overly large memory requirement (10 * (3min + 15s) = 32.5min).

### DIFF
--- a/columnflow/tasks/framework/inference.py
+++ b/columnflow/tasks/framework/inference.py
@@ -6,6 +6,8 @@ Base tasks for writing serialized statistical inference models.
 
 from __future__ import annotations
 
+import pickle
+
 import law
 import order as od
 
@@ -15,7 +17,7 @@ from columnflow.tasks.framework.mixins import (
     InferenceModelMixin, HistHookMixin, MLModelsMixin,
 )
 from columnflow.tasks.framework.remote import RemoteWorkflow
-from columnflow.tasks.histograms import MergeHistograms, MergeShiftedHistograms
+from columnflow.tasks.histograms import MergeShiftedHistograms
 from columnflow.util import dev_sandbox, DotDict, maybe_import
 from columnflow.config_util import get_datasets_from_process
 
@@ -42,7 +44,6 @@ class SerializeInferenceModelBase(
     # upstream requirements
     reqs = Requirements(
         RemoteWorkflow.reqs,
-        MergeHistograms=MergeHistograms,
         MergeShiftedHistograms=MergeShiftedHistograms,
     )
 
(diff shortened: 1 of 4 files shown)


mafrahm (changes requested):
Loading everything before processing everything in sequence works well when requesting a single histogram for all cards, but when multiple histograms are requested, this mostly increases the initial overhead before the actual cards are being created.

I'll approve this PR if we are able to load just one variable at a time.

Additionally, it might be nice to keep some functionality for parallelization for cases where the runtime is not dominated by overhead, e.g. allowing to parallelize over all the different requested histograms

columnflow/tasks/framework/inference.py
```diff
@@ -106,153 +107,184 @@ def get_data_datasets(cls, config_inst: od.Config, cat_obj: DotDict) -> list[str
...
 
-        :param cat_obj: category object from an InferenceModel
-        :param merge_variables: whether to merge the variables from all requested category objects
-        :return: requirements for the category object
-        """
+        return config_data
+
+    def create_branch_map(self):
+        # dummy branch map
+        return {0: None}
```
mafrahm, line 192: As discussed offline, just to keep it documented:
It might be nice (in a separate PR) to add the inference model categories to the branch data (branch map with length 1) & loop over the categories from the branch data. That way, one could overwrite the branch map from analysis side to reimplement the parallelization

mafrahm (approved):

PR #730, opened 2025-09-29
allow removing negative contributions per process

mafrahm:
Currently, when plotting a stacked histogram, where one process has a bin value < 0, this process can reduce the overall yield. While this might be valid some processes with negative interference between each other, this is usually not intended (e.g. reducing ttbar yield due to negative DY event weights doesn't make much sense).

This PR adds another parameter, `"remove_negative"`, which simply sets all negative bin entries to zero.

Open questions:
- should we also set the variance to 0 when `value < 0`?
- should we set this parameter to `True` by default?

### DIFF
--- a/columnflow/tasks/framework/base.py
+++ b/columnflow/tasks/framework/base.py
@@ -191,8 +191,7 @@ class AnalysisTask(BaseTask, law.SandboxTask):
         if (
             isinstance(getattr(cls, "version", None), luigi.Parameter) and
             "version" not in kwargs and
-            not law.parser.global_cmdline_values().get(f"{cls.task_family}_version") and
-            cls.task_family != law.parser.root_task_cls().task_family
+            not law.parser.global_cmdline_values().get(f"{cls.task_family}_version")
         ):
             default_version = cls.get_default_version(inst, params)
             if default_version and default_version != law.NO_STR:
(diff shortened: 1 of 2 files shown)


columnflow/plotting/plot_util.py
```diff
@@ -314,6 +314,15 @@ def apply_variable_settings(
     return hists, variable_style_config
 
 
+def remove_negative_contributions(hists: dict[Hashable, hist.Hist]) -> dict[Hashable, hist.Hist]:
+    for proc_inst, h in list(hists.items()):
+        h = h.copy()
+        h.view().value[h.view().value < 0] = 0
+        h.view().variance[h.view().variance < 0] = 0
+        hists[proc_inst] = h
+    return hists
```
riga, line 323: To be consistent with the other methods, we shouldn't adjust the input arguments in-place. The copy() below is shallow.


```suggestion
    _hists = hists.copy()
    for proc_inst, h in hists.items():
        h = h.copy()
        h.view().value[h.view().value < 0] = 0
        h.view().variance[h.view().variance < 0] = 0
        _hists[proc_inst] = h
    return _hists
```

riga (commented):
Looks good! Just one inline comment regarding in-place changes of arguments.

PR #731, opened 2025-09-29
add ParameterTransformation for ratifying + envelope if one-sided

mafrahm:
The PR adds two new ParameterTransformations in the DatacardWriter:

- `ratify`: transform shape into constant rate effect (but keeping it as shape type)
- `envelope_if_one_sided`: If both shape variations point in the same direction overall (using the integrals), use the  symmetrized effect of the larger variation

### DIFF
--- a/columnflow/inference/__init__.py
+++ b/columnflow/inference/__init__.py
@@ -95,7 +95,6 @@ class ParameterTransformation(enum.Enum):
     normalize = "normalize"
     effect_from_shape = "effect_from_shape"
     effect_from_rate = "effect_from_rate"
-    ratify = "ratify"
     envelope_if_one_sided = "envelope_if_one_sided"
 
     def __str__(self: ParameterTransformation) -> str:


columnflow/inference/__init__.py
```diff
@@ -95,6 +95,8 @@ class ParameterTransformation(enum.Enum):
     normalize = "normalize"
     effect_from_shape = "effect_from_shape"
     effect_from_rate = "effect_from_rate"
+    ratify = "ratify"
```
riga, line 98:
```suggestion
```

columnflow/inference/cms/datacard.py
```diff
@@ -526,7 +526,28 @@ def get(hd: dict[Hashable, hist.Hist]) -> hist.Hist:
...
                     for trafo in param_obj.transformations:
-                        if trafo == ParameterTransformation.centralize:
+                        if trafo == ParameterTransformation.ratify:
+                            n, d, u = integral(h_nom), integral(h_down), integral(h_up)
+                            ratio_up = safe_div(u, n)
+                            ratio_down = safe_div(d, n)
+                            h_down = h_nom.copy() * ratio_down
+                            h_up = h_nom.copy() * ratio_up
+
+                        elif trafo == ParameterTransformation.envelope_if_one_sided:
```
riga, line 536:
```suggestion
                        if trafo == ParameterTransformation.envelope_if_one_sided:
```

riga (approved):
As discussed offline, dropping "ratify" here in favor for the recommended "effect_from_shape" transformation.

PR #735, opened 2025-10-02
Update sandboxes, avoid dask_awkward in IO

riga:
This PR addresses our IO in light of the new upstream features in awkward and coffea, and has three main components.

1. All sandboxes are updated to ship the latest versions of our dependencies.
2. **Main change**: The `awkward_parquet` mode of our chunked IO reader used to rely on dask awkward. I renamed this to `dask_awkward_parquet` and added a new set of methods that are triggered by using the old `awkward_parquet` name. Hence, no change was required by any task to use the new IO. This is now based on a new `ChunkedParquetReader` class that very much resembles the `DaskArrayReader`, but uses ak metadata to identify row groups, and the new `row_groups` feature of `ak.from_parquet`. Note that the threaded chunking algorithm is an adjusted copy of our dask-based variant. The implementation is slightly faster, and can profit from further optimizations upstream in the future. Ultimately, we might be able to sunset the dask functionality altogether and streamline our IO again.
3. By setting `CF_INTERACTIVE_VENV_FILE`, it is now possible to tell columnflow to use a different venv other than `cf_dev` for the basis of the interactive shell after sourcing `setup.sh`. We use this to always start from our default sandbox that contains our columnar software stack, avoiding the need to use `claw` for most of the time. Sandbox switching - in case a different venv is required - is not affected by this.

### DIFF
--- a/columnflow/columnar_util.py
+++ b/columnflow/columnar_util.py
@@ -3573,7 +3573,7 @@ class ChunkedIOHandler(object):
                     # priotize coffea nano events
                     source_type = "coffea_root"
                 elif source.endswith(".parquet"):
-                    # priotize non-dask awkward reader
+                    # prioritize non-dask awkward reader
                     source_type = "awkward_parquet"
 
             if not source_type:


nprouvost (changes requested):
just one comment and a typo

nprouvost (approved):
LGTM

PR #750, opened 2025-11-03
mutliple fixes regarding empty files or (almost) empty chunks

mafrahm:
This PR includes 4 hotfixes that happen if files or chunks are empty (or almost empty)

- EGamma: when no electrons are present, the SF evaluation fails. Solution: return directly when there are no electrons in a chunk
- ReduceEvents failed when all chunks were empty after reduction (due to skipping a chunk if it is empty). Solution: never skip the 1st chunk
- histograming fails when a chunk is empty (can happen e.g. if a user applies an event selection as part of a HistProducer). Solution: skip empty chunks.
- applying the process weights to our `process_id` column fails when there is exactly one event in a chunk. Solution shown in the following code snipped (I checked for two separate datasets that the outputs of `process_weight_previous` and `process_weight` are consistent)
```python
    process_weight_previous = np.squeeze(np.asarray(self.process_weight_table[process_id, 0].todense()))
    process_weight = np.squeeze(np.asarray(self.process_weight_table[process_id].todense()), axis=-1)
```

### DIFF
--- a/columnflow/calibration/cms/egamma.py
+++ b/columnflow/calibration/cms/egamma.py
@@ -71,13 +71,8 @@ class EGammaCorrectionConfig(TAFConfig):
 def _egamma_scale_smear(self: Calibrator, events: ak.Array, **kwargs) -> ak.Array:
     # gather inputs
     coll = events[self.collection_name]
-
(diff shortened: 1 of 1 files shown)


riga (approved):
Looks good!

PR #754, opened 2025-11-10
Add muon scale and resolution calibrator.

riga:
This PR provides a calibrator that calibrators the scale and resolution of muons.

The procedure is straight forward and described at

- https://gitlab.cern.ch/cms-muonPOG/muonscarekit and
- https://cms-analysis-corrections.docs.cern.ch/corrections_era/Run3-22CDSep23-Summer22-NanoAODv12/MUO/latest/#muon_scalesmearingjsongz

What is different here compared to other calibrators evaluating central correction files is that external code is required. In this instance, it is a single python file that contains helpers provided by the CMS MUO POG.

It is meant to be provided as an external file and then simply loaded with out `import_file` util.


mafrahm (approved):
LGTM

PR #773, opened 2026-01-12
fix data uncertainties for density

mafrahm:
When applying the `--density` parameter, the unweighted poisson error calculation does not take into account the rescaling, leading to wrong data stat. uncertainties.

This PR fixes this issue by "reverting" the density before error calculation and the reapplying the density afterwards.
To achieve this, one needs to pass the "density" argument through the plot functions. I only added this argument for the data error calculation in the `plot_variables_1d` function because the MC uncertainty bands are considering the variance of the histogram and therefore working out of the box.

I also updated the ylabel when using density to display e.g. "Entries / GeV" instead of "Entries"


riga (approved):
Looks good!

PR #775, opened 2026-02-04
Add btag count selector.

riga:
This PR adds a selector that is intended only to fill selection histograms with the counts of jets passing specific b-tag working points.

Histograms of this type are necessary to measure b-tagging efficiencies (in MC) required for WP-based scale factor methods like [here](https://btv-wiki.docs.cern.ch/PerformanceCalibration/fixedWPSFRecommendations/#b-tagging-efficiencies-in-simulation).

As usual, a dataclass can be created and stored in the main config's auxiliary fields to configure everything about the counting process.

**Also note** that the selector has no return value and is therefore not exposed to the command line (it cannot be invoked with `--selector btag`), but it adds the histogram in-place to the overall selection histograms.

@Lara813 @jolange @mafrahm

### DIFF
--- a/columnflow/plotting/plot_all.py
+++ b/columnflow/plotting/plot_all.py
@@ -277,6 +277,7 @@ def draw_errorbars(
     h: hist.Hist,
     norm: float | Sequence | np.ndarray = 1.0,
     error_type: str = "poisson_unweighted",
+    density: bool = False,
     **kwargs,
 ) -> None:
     import hist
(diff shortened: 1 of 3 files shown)


columnflow/selection/cms/btag.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # add used columns
+    self.uses.add(f"{self.cfg.jet_name}.{{pt,eta,phi,mass,hadronFlavour,{self.cfg.btag_column}}}")
+
+    # compute working point ranges, starting with the lowest bin ranging from "veto" to the first wp entry
+    self.wp_ranges = {}
+    for lower_key, upper_key in zip([None] + list(self.cfg.btag_wps.keys()), list(self.cfg.btag_wps.keys()) + [None]):
+        self.wp_ranges[lower_key or "veto"] = (
+            self.cfg.btag_wps.get(lower_key, float("-inf")),
+            self.cfg.btag_wps.get(upper_key, float("inf")),
```
aalvesan: we shouldn't get values outside the discriminant range ?

```suggestion
            self.cfg.btag_wps.get(lower_key, 0.0),
            self.cfg.btag_wps.get(upper_key, 1.0),
```

PR #790, opened 2026-03-09
Keep object level association when filling histograms

BalduinLetzer:
When filling histograms, currently ak.cartesian is used to build all combinations between all entries of an event. This is not an issue when plotting individual or event level variables. When plotting something like Jets pt vs Jets eta for all Jets, however, this counts all combinations of Jets pt and eta values, loosing the association between the two properties and effectively double counting of the events.
For two test variables with test_var1 = [0, 1] and test_var2 = [1, 3] for each event this results in [this plot](https://github.com/user-attachments/files/25844481/plot__proc_qcd__cat_incl__var_test_var1-test_var2.pdf), where all combinations are plotted.

This PR allows the addition of an entry called association in the variable auxillaries:
```
config.add_variable(
        name="test_var1",
        expression="test_var1",
        binning=(10, -5, 5),
        x_title="Test variable 1",
        aux={
            "associated_with": "test_vars",
        },
    )
```
When filling the histograms, if two or more variables with the same association exist, the only the pairwise combination of the associated entries are build, and then later combined with the usual axes (category, shift, process, etc...).
For the two test variables this now yields this [this new plot](https://github.com/user-attachments/files/25844596/plot__proc_qcd__cat_incl__var_test_var1-test_var2_diag.pdf), with only the diagonal entries filled.

### DIFF
--- a/columnflow/hist_util.py
+++ b/columnflow/hist_util.py
@@ -33,6 +33,7 @@ def fill_hist(
     *,
     last_edge_inclusive: bool | None = None,
     fill_kwargs: dict[str, Any] | None = None,
+    **kwargs,
 ) -> None:
     """
     Fills a histogram *h* with data from an awkward array, numpy array or nested dictionary *data*. The data is assumed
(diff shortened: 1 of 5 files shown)


riga:
Fixed by #796.

PR #798, opened 2026-04-21
remove discrete_x of variables that should not be constructed as Integer axis

mafrahm:
All variables with the `discrete_x` tag are automatically constructed as histograms with an Integer axis, overwriting the number of bins. Instead, histograms with millions of bins are constructed, leading to memory issues in CreateHistograms.

@LukasEbeling

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -184,6 +184,15 @@
       "contributions": [
         "code"
       ]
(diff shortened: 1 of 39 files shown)


riga:
Seems to have been fixed in an earlier PR, there are zero files changes :)

PR #807, opened 2026-07-08
Enable new JER format in jer calibrator.

riga:
This PR extends the jer calibrator to accept the new JME format with a dedicated `*_SFUncertainty_*` correction in the clib files.


LennertGriesing (approved):
Looks good!

